PR #283, opened 2015-05-15
Market Tooltips Patch

Bombad:
Added the trade route bonus in the mouseover tooltip for the Market and the Bank (it was missing).

Changed some tooltips to correct them and/or add consistency (consistent lineskips, order of the various bonus (for bazaar/market), and the word "gold" was missing after the gold icon in some cases).


LoneGazebo:
Awesome, thanks!I I've manually merged.

PR #1048, opened 2016-01-04
Turn time improvements

alberts2:
* Added the COMMUNITY_MINIMIZEAITURNSLICES option.
When enabled, it runs the AI turns in the minimal number of turn slices.
It will speed the turn processing up at the expense of animation and general graphical responsiveness while the AI is processing its turns.
This was done by Koshling for Civ4 a few years ago but it works with a few modifications in Civ5
*  Made a small efficiency improvement in CustomMods::getOption

The COMMUNITY_MINIMIZEAITURNSLICES option can be disabled in the Community Patch/CommunityOptions.sql  if it should cause problems.


alberts2:
The COMMUNITY_MINIMIZEAITURNSLICES option is Singleplayer only.

LoneGazebo:
Awesome, thanks!

PR #1054, opened 2016-01-06
Pedia Entry for Hidden antiguity sites

GrantSP:
The vanilla text still describes Exploration as being the name of the
policy that reveals the hidden sites.


LoneGazebo:
I saw this too! Good catch.

GrantSP:
I was a little unsure about where to make that edit, still getting my head around the various places for changes. Please let me know at any time if my changes are in the wrong place.

LoneGazebo:
Well it needs to be in the CBP, not the CP. I'll fix.

PR #1446, opened 2016-03-13
C4DF Final Changes

mwhiter:
I don't think much more can be done for this version at the time.

```
[b]General[/b]
[*]A team with map trading cannot become the target of a City State quest to find their lands (too trivial for a player to just buy their land for easy influence)
[*]Changed mod descriptions

[b]Bugfixes[/b]
[*]Minor DLL fixes for C4DF (removal of unnecessary stuff, no effect on gameplay)
[*]Fix where you can't put "Vassal State / Capitulation" on both sides of the trade table at any one time. Also applies to "Liberation".
[*]Fixed minor tooltip bug: Master Pop %
[*]Added missing top panel texts for vassals
[*]Adjusted background for Autocracy in Vassal Overview - was incorrectly set to "loadingbasegame_13.dds", now "loading_13.dds"
[*]Crash fix protection for getting opinion of vassalage - protect and failed protect
```


LoneGazebo:
Merged up yesterday evening.

PR #1512, opened 2016-03-26
Workaround to fix a possible crash issue on (1a)

mwhiter:
See [this thread](http://forums.civfanatics.com/showthread.php?p=14213756#post14213756) here.

Two things:
1) I was able to reproduce on C4DF-standalone, but not (1a), but it was also reported on (1a) so I figure I'll push something I know works (for now).
2) I had to delete my local branch C4DF-dev and the remote one. Kept telling me diverging, and didn't feel like reconciling it.


mwhiter:
Want to spend some time looking at a proper fix to this, which I think is simply not being an idiot while programming.

PR #1513, opened 2016-03-26
Real fix to the C4DF crash issue.

mwhiter:
Pushing the fix to the crash issue on "Restart Game". Also moving around some text to host C4DF code in C4DF only.

Experiencing a lot of issues when trying to merge and reconcile with the source - I'm probably doing something wrong. Let's discuss ways to keep me in sync with master in depth so that it's not so frustrating to develop.


mwhiter:
closing this, want to re-do it from latest master similar to corp-helper

PR #1551, opened 2016-04-05
Corporations Overview

mwhiter:
Corporations Overview. Tested and stable enough for a run for UAT.

Disabled in-game Economic Overview, replaced with Corporations.


LoneGazebo:
Awesome. I'll bundle into the next beta. Anything I need to know aside from the basics? Fantastic work here, by the way.

mwhiter:
It should work fine on its own. Thank you! 
My next work involves doing an architecture revamp of Corporations.

LoneGazebo:
Awesome. Keep in mind that the helper functions (hasoffice, isfranchised, etc.) depend on values that are set when buildings are constructed. If you're talking about the struct for corps from a base level (i.e. the struct we discussed earlier) that should be fine.

LoneGazebo:
Quick question: should resources in the monopoly tab populate as the player 'sees' them? It's not terribly 'spoilerish' for them to all be there at the start of a game, but it is a little odd that the player is omniscient about all resources in the game! :)

mwhiter:
Theyway it works now, the resoutve shows up if it's in the game or the player owns an amount of it. I don't know the answer to what it should be.

PR #2442, opened 2016-08-31
Stuff I forgot to commit yesterday

Iamblichos:
Something to note: if a modder wishes to use the CP only and it is using entries from the DLL that involve new UI text from the EnemyUnitPanel, UI texts will be missing unless we move those texts over into the Community Patch...


Iamblichos:
I would move all Texts that are essential to DLL entries over to the Community Patch.

LoneGazebo:
What texts are essential? Also, your cargoships stuff - that's simply the support framework for your modmod, right? Not base-CP included?

Iamblichos:
It's base CP/DLL, the CARGO_SHIPS in CustomModOptions is turned off in the XML by default, if that switch is turned on, it changes all the units based on what I did in CargoShips.sql.

Once you make a build of DLL, don't release let me test it real quick to make sure everything isn't F'ed up before release....I've been playing with it On for a while now, but I need to make sure DLL is behaving properly with it Off.

Iamblichos:
The Promotions for the CARGO_SHIPS mod is hardcoded in the DLL...not much just some basic rules for land units on Melee ships....and a couple of neado tweaks for ships with Recon units, Prize Ships promotion (Privateer), and Polynesia boats.

Iamblichos:
In regards to 'what is essential' compare the basic Expansion2 EnemyUnitPanel.lua to the Community Balance Patch EnemyUnitPanel.lua.........those are functions that could be used in base CP by a modmod, but Expansion2 EnemyUnitPanel.lua would not show those changes unless that Panel were included in the CP and also those UI texts associated with those functions.

Iamblichos:
This could also apply to other functions and entries, not just EnemyUnitPanel.

Iamblichos:
This is isn't a pressing issue, but it's probably best to include all essential functions and entries that are a part of the CP DLL Lua calls into the base CP LUA folder, that way a modmod doesn't have to recreate the LUA calls that are already a part of the CP DLL LUA hook. Just SQL and XML changes.

LoneGazebo:
Just for my own sanity, mind moving over the requisite text files to the CP and making a pull? That way I don't miss anything.

Which lua functions are you referring to?

Iamblichos:
I don't mind at all.

The only one I know of at the moment is the EnemyUnitPanel.lua

Iamblichos:
For my pull requests I did today and yesterday, I've included all those texts into CP. I don't want to let text finding and updating .lua's hold up release. But I will take point on this.

Iamblichos:
@LoneGazebo , when you get done on your end updating DLL, put a clean build of the CP in your depository Build folder that way I can do a quick test of the Cargo Mod, just be sure to include the .sql of it in the build as a database update probably at the very end.

LoneGazebo:
Sounds good.

PR #2450, opened 2016-09-01
Some bug fixes and consistency

Iamblichos:


LoneGazebo:
cargospaceavailable has a 'not all paths return value' warning - fixed by adding a return 0 after the if(DOMAIN_LAND) check.

Iamblichos:
he he...yeah I did a pull request for this ...

Iamblichos:
but use mine, tested in game, all good. Plus we need to return the correct value if it's not DOMAIN_LAND

PR #2451, opened 2016-09-01
Make sure all control paths return a value.

Iamblichos:
Tested. with and without option enabled.


Iamblichos:
Okay I'm happy now....all the custom mods are working off or on as intended.

Iamblichos:
How far away are you from release, I'm anxious to get this thing fired and rolling again!

LoneGazebo:
Not too far.

PR #2496, opened 2016-09-09
Fix a CTD.

Iamblichos:


LoneGazebo:
Good catch. I fixed it already (same thing) so I'll close. :)

Iamblichos:
Just trying to be on top of these :)

PR #2544, opened 2016-09-20
Some changes...

Iamblichos:
I added into the DLL for mountain plot cities to be IMPROVEMENT_JFD_MACHU_PICHU. Not sure if you are still using that LUA hook or not.


LoneGazebo:
Sounds good!

PR #2551, opened 2016-09-21
fix Shoshone MP, revert a bad change.

Iamblichos:


Iamblichos:
@LoneGazebo 
I changed that one part in the goody hut picker isReallyNetworkMultiplayer from what I posted in the Issues. Just FYI...

PR #2600, opened 2016-10-06
Lots of changs and bugfixes

Iamblichos:


Iamblichos:
Let me know if you have any questions about some of the things I did here. I did some modifications to some pieces of your code which I think needed some consideration or could give an improvement.

LoneGazebo:
Looking at it now - looks fine so far. Re: changing existing code, just some speed optimizations, looks like?

LoneGazebo:
You tested your changes, ja? I'm kinda hoping to make a release sometime soon.

Iamblichos:
Yes I have. Been testing all week. Yes some optimizations and some more flexibility for GG/GA promotions. For example, what if you want to consolidate a special GG promotion with the standard great general promotion + something else? The old code would always grab the default great general promotion because it breaks when it finds and then spawns that unit. If you don't have that unit set with that free promotion, then no spawn. Modified some things with the WLTKD UA, thought it should extend to GG's and GA's, along with free one's.

LoneGazebo:
For China? Theoretically it should only be for ones born in the city naturally, but China isn't a powerhouse like it was (b/c growth is weaker now), so yeah, I think its fine. I'll edit China's UA text.

Iamblichos:
Yep...but that sorta newters the GG and GA production since they are GP (not tied to a city, except the capital), so since they are tied to the capital (or a unit) the capital gets the reward.

Iamblichos:
I also edited the retro promotion stuff....Starting units don't have a city of origin, since there is no capital on start (and they aren't produced there anyways), so if that origin city is NULL, they get the promotion when the capital builds the building or buys it with faith.

Iamblichos:
and then civilian units....I modified now that civilians can get a free promotion from buildings

LoneGazebo:
Awesome. I adjusted the GG/GA stuff so that it gives the WLTKD to the city they spawn in, just for flavor's sake.

PR #2793, opened 2016-11-09
Merill civ6 addition

supermerill:
Hi

in this pull request:

debug the auto roads building by trade route
add lua hooks for ui for builders
rename previously bad/ambiguously named methods (from my previous pull request).
system & lua hooks for eureka system (in creation) (with a new define)
note: in CvTechClasses.cpp line 2374, i didn't put the eureka code into a "if (MOD_CIV6_EUREKA)" because it didn't change anything with the default value, but you can add it if you feel the need.

Question: do you prefer small pull requests or a fewer amount of bigger pull request?

Thank you very much for your dedication to the community!


LoneGazebo:
A pull is a pull - works for me!

PR #2869, opened 2016-12-03
City states evolved fix

supermerill:
Hi
It's a fix for some regressions since vanilla (request from tpangolin):
Some civ.isMinor() has been added at some places to optimize the compute time.
I replaced them by checking if the IsNoMinorCivs of the adequate strategy is checked.

relevant thread:
https://forums.civfanatics.com/threads/city-states-evolved-for-the-community-patch.606062/

note: i added all fix inside "#define" but some times, it's a revert from current code to a vanilla one. I don't know how to handle this in terms of "we must put all our changes inside #defines to track them" 
  
note2: I may be working a bit much to optimize CS AI. Can I?


LoneGazebo:
Thanks! Feel free to optimize the CS as you see fit.

PR #3220, opened 2017-05-06
fix event overview panel lock UI next turn issue

neowu:
Hi, I am new to Civ5 mod, 
When I play, I found occupationally if i open/close event overview panel and switch sub panel inside (with EUI), the next turn button can be locked, (no effect when click, but still clickable) , i can reproduce this issue very easily by repeating those operations.

after taking look into the lua file, I think I may found the cause, in those 4 sub panel lua files, 
PlayerEventsInfo.lua, RecentEventsInfo.lua, CityEventsInfo.lua,  CityRecentEventsInfo.lua
they all have ShowHideHandler, and do UI.incTurnTimerSemaphore(); on show and UI.decTurnTimerSemaphore(); on hide

and in EventOverview.lua, it saves m_CurrentPanel as state and call setHide() method for all panels, on close it looks like not calling subpanel SetHide(), and i am not sure how setHide() works internally, whether repeat clicking same sub panel link can result in execute showHideHandler multiple times which can result in unpaired incTurnTimerSemaphore/decTurnTimerSemaphore 

anyway i don't think it's necessary to use incTurnTimerSemaphore on sub panel, since the pair in EventOverview panel will do the job. 

so I removed those code, also cleaned dead code within those 4 files, then it fixed my issue.

please review the change, and let me know if I missed anything.

Thanks,

### DIFF
--- a/Community Patch/Core Files/CoreLua/CityEventsInfo.lua
+++ b/Community Patch/Core Files/CoreLua/CityEventsInfo.lua
@@ -9,42 +9,22 @@ local g_ItemManagers = {
 	InstanceManager:new( "ItemInstance", "Button", Controls.CityEventsStack ),
 }
 
-local bHidden = true;
-
-local screenSizeX, screenSizeY = UIManager:GetScreenSizeVal()
-local spWidth, spHeight = Controls.ItemScrollPanel:GetSizeVal();
-
--- Original UI designed at 1050px 
-local heightOffset = screenSizeY - 1020;
-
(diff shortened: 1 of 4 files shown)


TechpriestEnginseer:
Since @JFDaly handled part of the Event on the Lua side. He should probably be notified of this fix!

LoneGazebo:
Looks good, thanks! This is one of those issues I'd noticed but never taken the time to debug. Strong work!

PR #3379, opened 2017-07-26
Event "Balance"

TechpriestEnginseer:
Old Man and the Sea Choices modified so you actually chose instead of saying free harbor everytime this  pops up.
Pay 50 Gold for 50% Chance for +200 Food in the City.
Pay 150 Gold for 50% Chance for +5 Food on Fishing Boats for a # of turns in the City.
2.5x the Gold required to purchase the Harbor.

A Local Stadium Event is great, but the 4th Event is extremely lacking in being chosen, so...
1st, 2nd, 3rd Choice had its effect 0.5x

Bank Decision Event
The 1st Event Choice grants a free Social Policy which is ridiculously going to be picked no matter what. It has been changed to incorporate Laborers and Merchants earning +1 Gold instead for decentralization.


LoneGazebo:
Thanks for this, @TechpriestEnginseer.

PR #3422, opened 2017-08-09
Update UnitPanel.lua

TechpriestEnginseer:
YieldTypes don't work, but GameInfoTypes work. This'll have to do for a temporary solution(or permanent).


TechpriestEnginseer:
Can you replicate this change for YieldTypes.YIELD_CULTURE_LOCAL since I think the 3UC and 4UC project are already using the yield.

Issue #3406, opened 2017-08-06
Featurequest-Diplo screen shows which city states already have an embassy.


preyator:

TechpriestEnginseer:
If the UI can support it, you could just add it as a pseudo-quest where they can hover over the embassy icon and see who owns it.

TechpriestEnginseer:
I think I got this solved already. Pulling a request now.

TechpriestEnginseer:
I actually got a simple proposal for this. It may not be as what people expected, but it'll do considering optimal performance. Requesting a pull from it right now.

LoneGazebo:
Happy to look at it.

thecrafybee:
If you are able to make an icon for showing if there is an Embassy can you also add one for Feitorias too?  Just played a game with Portugal and sometimes found that it was difficult to see which CS had one.

LoneGazebo:
It's a lot of work for limited gain, tbh - you can see where they are from your trade screen. Besides, scanning the map is half the fun! :)

PR #3605, opened 2017-10-23
Update CityStateStatusHelper.lua

TechpriestEnginseer:
image
image

This code I tested out for EUI. It basically allows you to view whether or not an Embassy has been established. It won't tell you who owns it because that's too much codework for a simple info that you should just hover to the city-state itself and check it out for yourself.

Ignore the linebreak though. My modmod makes a linebreak to distinguish its CS trait from the rest of the info.

BUT FOR NON-EUI, I DO NOT KNOW HOW THE FORMAT WORK SO THIS IS JUST BASED ON MEMORY. TEST OUT THE NON-EUI UI IF APPROVED

https://github.com/LoneGazebo/Community-Patch-DLL/issues/3406


LoneGazebo:
Danke!

PR #3612, opened 2017-10-26
Fixed typos in Denouncement dialogues

drnooooo:
I noticed that all the Denouncement dialogue responses were missing a closing parentheses, which is a little embarrassing to me, since I'm the one who wrote them a few years ago. I fixed them so they no longer have that typo.


LoneGazebo:
I added this manually after you pointed it out, thanks!

Issue #3819, opened 2018-01-07
Yield valuation bugs in CityStrategyAIHelpers::GetBuildingYieldValue()


Infixo:
_1. Mod version (i.e Date - 4/23): 1-6
_2. Mod list (if using Vox Populi only, leave blank): none
_3. Error description:

While calculating iFlatYield gains, there is:
```
int iValue = (pCity->getPopulation() * 100 / pkBuildingInfo->GetYieldChangePerPop(eYield));
```
The iValue is bigger for worse buildings, and gets smaller for better buildings. E.g. 1 yield per 2 pop is coded as 50 in YieldsChangePerPop, so at 10 pop it evaluates to 20. 1 yield per 6 pop coded as 17 evaluates to 59. Should be the other way around?

Similar issue is with YieldPerXTerrain yields. Basically, the more Mountains or that specific Terrain is there within our borders, the less valued it is.
```
iFlatYield += (pkBuildingInfo->GetYieldPerXTerrain(eTerrain, eYield) / pCity->GetNearbyMountains());
iFlatYield += (pkBuildingInfo->GetYieldPerXTerrain(eTerrain, eYield) / iNumTerrain);
```

Plus a problem with Oil not properly evaluated.

Fixed all of the above.

Also, not bug but an observation. The main yield is given very small valuation compared to all other tests in that function. Basically it's value is yield value (so in range 1-5 usually).
```
iFlatYield += pkBuildingInfo->GetYieldChange(eYield); }
```
Other flat yield gains are multiplied by 5 very often.

LoneGazebo:
Ugh, duh, that's what I get for bulk-writing code without thinking.

PR #3821, opened 2018-01-07
Fix for #3819 #3821 #3828, Civilopedia

Infixo:
Fixed bugs in yield valuation


LoneGazebo:
Does your most recent set of commits include LUA for the CBO and the 43-civ LUA?

PR #3835, opened 2018-01-10
PolicyGEorGM string update

HungryForFood:
- Notification for triggering PolicyGEorGM no longer repeated for every city.
- Notification now able to show the yields given to each city.
- On-map popup now uses the specific great person icon, and also shows the yields given to each city.
- Notification now uses the golden age begun notification type, rather than the generic notification type.


LoneGazebo:
I think this is already merged.

HungryForFood:
Hmm,what do you mean? Are you referring to my pull request one month ago? If so, I did some changes to how the UI handles it.

PR #3844, opened 2018-01-12
TR Turns, Invested Flag, Gold Button, Revert #3840

### LINKED ISSUE #3840: Monopoly percentage change
_1. Mod version (i.e Date - 4/23): 1-8
_2. Mod list (if using Vox Populi only, leave blank): none
Something that boggled my mind from the first day of using VP: having >100% of resources. 
monopolyissue

Ofc I know why it is like that in terms of mechanics, but imho they should be corrected. Unless there is some deep logic and meaning that I don't see - I have changed it to work as it should imho: percentage is calculated against all "activated" resources in the world. Example. We have 8 on the map of which 6 are connected and 5 from other means (GA, CS, bonuses, etc.). The percentage will be calculated not against 8 but against 11. It wil never be >100%.

Infixo:
You pulled my changes into master but this revert didn't make it. This change is not optional yet, so it needs to be reverted.


LoneGazebo:
We can leave it in master, my working copy on my machine doesn’t have it, I edited it out for now. ;)

PR #3856, opened 2018-01-15
Added column AoEDamageOnMove to UnitPromotions

HungryForFood:
- Does AoE damage to adjacent units equal to the integer defined in AoEDamageOnMove when a unit moves (including melee units after killing another unit).
- Changed function CvUnit::DoAoEDamage so that the string tag to be displayed is an argument. Changed code relating to AoEWhileFortified to account for this.

I do not see any functions in the whole DLL which takes char 256 as an argument. Is there a particular reason for this?


LoneGazebo:
Not sure, but be careful with some types, they may not pass through the netcode data size filter.

HungryForFood:
What's the netcode data size filter?

LoneGazebo:
Max size of data type and throughput for data. Shouldn’t be an issue if you’re passing the data internally within a function, so long as you’re not trying to return the type.

PR #3864, opened 2018-01-16
Prussian Discipline

HungryForFood:
- Added table UnitPromotions_CombatModPerAdjacentUnitCombat. The unit with this promotion gets combat bonus based on the number of friendly units nearby of the specified unit combat types.
- Also found a possible bug in line 349 and 350 of CvUnit.cpp.

Haven't thought about this before, but is there a practical limit to the amount of new code / tables you can add to the DLL?


LoneGazebo:
Yes there is a limit. We need to be judicious with our additions.

HungryForFood:
Ah, how close are we, I'll think about what's really necessary to add.

PR #3872, opened 2018-01-20
Important: fixes CTD on Turn 0 and DB errors due to wrong character

Infixo:
Missing NULL in recently added code, plus some wrong characters in SQL that caused DB errors

### DIFF
--- a/43 Civs CP/43 Civs CP/EUI/CityView.lua
+++ b/43 Civs CP/43 Civs CP/EUI/CityView.lua
@@ -888,11 +888,13 @@ local function SetupBuildingList( city, buildings, buildingIM )
 				-- Vox Populi Comparable Yields
 				--buildingYieldRate = buildingYieldRate + (foodConsumed < foodPerPop * population and foodPerPop * numSpecialistsInBuilding / 2 or 0)
 				--buildingYieldModifier = buildingYieldModifier + (tonumber(building.FoodKept) or 0) -- FoodKept has a different meaning
+				--[[ Infixo turned off due to confusion
 				if foodConsumed < foodPerPop * population then 
 					-- this only happens when specialists in the city consume less food that normal population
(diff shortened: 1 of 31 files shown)


HungryForFood:
For the missing NULL, there's similar code in lines 15484 and 16090. Why don't those CTD?

When I tested the changes in game, I don't remember crashing. Any idea why is that? Maybe its because I had a unit which has a promotion defined in UnitPromotions_CombatModPerAdjacentUnitCombat from turn 0?

How did you manage to find the error? Might be helpful for me in the future.

Sorry for the multiple questions, I only do programming as a hobby, and have zero formal training.

LoneGazebo:
@HungryForFood because units that are init'd at game start don't always have a fromplot defined on movement.

Error happened for me too just by doing 1 turn. As a rule, always make sure you're passing non-NULL values through a function if the function doesn't protect against NULL being an option.

Infixo:
I have checked a minixump. The stack showed direct call like xxx(null,null,null). So, it was obvious.
@HungryForFood your text editor adds some invisible unrecognized char at the begining of the file. Can you check that out? It causes both db and lua errors.

HungryForFood:
@Infixo I'm using Notepad++, known issues with that?

Perhaps something to do with encoding? Right now, under the encoding menu, the option "Encode in UTF-8-BOM" is highlighted.

LoneGazebo:
You're using Notepad++ to write your c++? Use VS 2013.

LoneGazebo:
Turn off BOM, use just UTF8.

Infixo:
Exactly, use only UTF-8.

HungryForFood:
Ok, I'll switch over to VS.

How about lua, is it fine as long as I check that it is on UTF-8?

The curious bit is that it worked on my side, like for EnemyUnitPanel.lua which was fine when I tested it. Something to do with git? Does it work in UTF-8, so it messes things up when I push?

Issue #2576, opened 2016-09-29
Autosaves and AI 'double turns' in MP


LoneGazebo:
@ilteroi and @Iamblichos,

Let's solve this. I see this line in CvGame::Read:

```
//when loading from file, we need to reset m_lastTurnAICivsProcessed 
//so that updateMoves() can turn active players after loading an autosave in simultaneous turns multiplayer.
m_lastTurnAICivsProcessed = -1;
```

Seems, however, that this isn't happening. If you load ANY save, the AI gets to take a turn before the human player does, effectively skipping a human turn in the process.

Thoughts?

ilteroi:
this is the signature for the function creating the save:

AutoSave(bool bInitial = false, bool bPostTurn = false)

notice the second flag. it's called in several places.

for manual play, we have

	case CONTROL_ENDTURN:
	case CONTROL_ENDTURN_ALT:

			if (!isNetworkMultiPlayer() && kActivePlayer.isHuman() && GC.GetPostTurnAutosaves())
			{
				gDLL->AutoSave(false, true);
			}

in autoplay mode, it's handled differently (before turn processing)

	if(getAIAutoPlay())
	{
		gDLL->AutoSave(false);
	}

and then in networked games again differently (after turn processing)

	if(isNetworkMultiPlayer())
	{//autosave after doing a turn
		gDLL->AutoSave(false);
	}

so there are different kinds of autosaves and i don't know if it's possible to tell them apart when loading

LoneGazebo:
Think it would hurt to set that flag in the post turn call?

Iamblichos:
This is just an issue with Simult/Hybrid right? Could we just differentiate that out above for that type of game?

LoneGazebo:
I think so, yep. So perhaps?

Iamblichos:
We'll have to experiment (I think), but let's start just adding in the post turn flag to true.

LoneGazebo:
And so I have done it.

Adonis-VII:
This issue is still around in the May 26th 2017 version. My friend and I play on epic, so our games usually last through a couple updates. We played a couple separate sessions, making saves as we go. We normally just load the last save from previous session and go from there. We noticed on our last session that all saves from the previous session would do a double turn on load, so we tested a number of them.

>302 - Double
302 - Works
298 - double
295 - double
292 - double
285 - double
283 - works
279 - double

Here is dropbox link to modpack and saves.
[...]

PR #4002, opened 2018-02-28
Fix for skipped turn on load in MP issue

davenye:
After loading an MP game the first turns of human players can often be skipped, meaning the AI gets two turns in a row. I traced the problem down to being a result of the players erroneously being flagged as
having already submitted their ready message for the turn. The fix isn't really a fix as I believe the problem is in the host. Instead I just check for the problem upon the first activation of the local player and
send the unready message if need be. 

In short, HasReceivedTurnComplete was sometimes already returning true after a load so I call sendTurnUnready when it does so their turn doesn't end immediately.

I have played a few games (2 human/6 AI) with the change and have not noticed any negative effects but maybe you can. 

I can probably dig up some save games which exhibit the issue but they will have been using a modded version of the 2/7 release. I didn't figure out why the problem occurs only sometimes.

The issue has been spoken about in:
#2576 

EDIT:
Naturally, I just remembered that I, possibly wrongly, assumed that the problem only occurs when player with dynamic/simultaneous turns. My code will only execute in those modes and changes will still need to be made if sequential turns modes also have the same issue. Should be straight forward though.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CustomMods.h
+++ b/CvGameCoreDLL_Expansion2/CustomMods.h
@@ -955,6 +955,8 @@
 #define MOD_BUGFIX_NO_HOVERING_REBELS               gCustomMods.isBUGFIX_NO_HOVERING_REBELS()
 // Fixes some bugs/regressions that disable the effect of IsNoMinorCivs of some strategies
 #define MOD_BUGFIX_MINOR_CIV_STRATEGIES				gCustomMods.isBUGFIX_MINOR_CIV_STRATEGIES()
+// Workaround for the first human turn after loading a MP game being skipped sometimes
+#define MOD_BUGFIX_SKIPPED_HUMAN_TURN_ON_MP_LOAD	(true)
 
 #endif // ACHIEVEMENT_HACKS
 
--- a/CvGameCoreDLL_Expansion2/CvGame.cpp
+++ b/CvGameCoreDLL_Expansion2/CvGame.cpp
@@ -9087,6 +9087,10 @@ void CvGame::updateMoves()
 
 	int currentTurn = getGameTurn();
 	bool activatePlayers = playersToProcess.empty() && m_lastTurnAICivsProcessed != currentTurn;
+
+#if defined(MOD_BUGFIX_SKIPPED_HUMAN_TURN_ON_MP_LOAD)
+	bool firstActivationOfPlayersAfterLoad = activatePlayers && m_lastTurnAICivsProcessed == -1;
+#endif
 	// If no AI with an active turn, check humans.
 	if(playersToProcess.empty())
 	{
@@ -9364,6 +9368,22 @@ void CvGame::updateMoves()
 				if(!player.isTurnActive() && player.isHuman() && player.isAlive() && player.isSimultaneousTurns())
 				{
 					player.setTurnActive(true);
+#if defined(MOD_BUGFIX_SKIPPED_HUMAN_TURN_ON_MP_LOAD)
+					if (firstActivationOfPlayersAfterLoad && player.isLocalPlayer())
+					{
+						// DN: There is a strange issue with players missing their turns after loading a game, with the AI getting two turns in a row.
+						// It seems *to me* that Civ is incorrectly thinking telling us that the players have already indicated they have finished their turns
+						// A hacky solution to this is to tell Civ to cancel the player turn complete state.
+						// Otherwise they get their turn ended in the next call to updateMoves after the condition (!player.isEndTurn() && gDLL->HasReceivedTurnComplete(player.GetID()) && player.isHuman())
+						if (gDLL->HasReceivedTurnComplete(player.GetID()))
+						{
+							bool unreadied = gDLL->sendTurnUnready();
+							bool turnComplete = gDLL->HasReceivedTurnComplete(player.GetID());
+							NET_MESSAGE_DEBUG_OSTR_ALWAYS("UpdateMoves() : Attempting to fix skipped first turn issue - HasReceivedTurnComplete(" << player.GetID() << ") returned 1, sendTurnUnready() returned "
+								<< unreadied << " and now HasReceivedTurnComplete(" << player.GetID() << ") returned " << turnComplete);
+						}
+					}
+#endif
 				}
 			}
 		}


LoneGazebo:
Interesting! We spent some time way back in the day messing with this and had a hard time tracking the bug down. Good work!

davenye:
Do we know if the issue also occurs with sequential MP? If so, the implementation of the fix will probably be a bit different to what I have currently got. Assuming it even fixes the problem in that mode.

davenye:
I couldn't get a sequential MP game to exhibit the problem so I reverted to my original, simpler implementation. Re-reading the bug report I saw that it was already assumed to only be a simultaneous/hybrid issue anyway, so all should be well.

LoneGazebo:
Oops, didn't mean to merge it. Let me know how your tests go.

raphaell666:
So, how is this fix going? I am having this issue 100% of the time when playing multiplayer, a fix would be simply marvelous. I am very happy to have stumbled upon this here, I had no idea a fix even existed!

LoneGazebo:
@davenye never reported back, we'll see what he says. Honestly, though, the easiest solution in the meantime is just to save your game twice before exiting. That removes the issue.

raphaell666:
>@davenye never reported back, we'll see what he says. Honestly, though, the easiest solution in the meantime is just to save your game twice before exiting. That removes the issue.

I will have to try that, thanks a lot for the advice! I will also see if I can figure out how to compile the DLL and will test this fix to help you guys if I am able.

davenye:
I have been using my fix as per the submitted pull request during a lot of other development work and testing (prepare for some pull requests!).

I did not investigate any further but I did not notice any issues. I see raphaell666 has noticed some deficiencies though.

LoneGazebo:
The two of you should coordinate, make sure you are happy with the changes, as they are big under the hood changes.

davenye:
I will look at incorporating their additions to my dev branches.

I did have a fairly in depth look at how saving works. I can tell you all the following:
* The various incantations of ICvEngineUtility1::AutoSave do not appear to do anything but change the filename.
* The Save function in MP games simply copies the "most recent" (determined somehow) autosave and gives it your specified name. As I understand it, Civ 5 originally shipped with only MP autosaves. Their manual save implementation was obviously quickly built on top. Presumably they do not allow the writing of new save files during the turn because it must cause some bad state when loading...
* The "double save" trick only works because of a bug that bypasses the special check they have in SaveMenu.lua that looks if it is MP and performs the copying of the autosave and instead it performs the normal SP code which gets the game to write a actual new file. So given Firaxis when out of their way to prevent mid-turn saves, it stands to reason that it is a bad idea to use the "double save" as a fix. Of course, it may be safe to do under certain circumstances. I have not experienced any such problems except for Active Diplomacy requests are not restored after load (as they are not written out in the first place).

I have some pull requests incoming regarding these.

davenye:
Also, if this is considered a big change, can we maybe get a beta for MP like you have done for other big changes in the past?

I have been making changes for the past two months and are much wider reaching than this change. I have been testing but as you know testing (esp. MP testing) takes a long time and I can only do so much. I believe my changes will greatly enhance MP experience and will be well worth the effort incorporating.

I am in the process of making the pull requests for you but it is taking some time. I have them all lined up in branches under "pull/*" on my github if you want to see how much there is to go.

LoneGazebo:
Make a new branch for yourself on the main repo. I’ll do a merge once your stuff is up and then do an Mp dll.

raphaell666:
Hey @davenye, your findings regarding saving and autosaving seem to corroborate with mine, as I also mentioned a little bit here: https://github.com/LoneGazebo/Community-Patch-DLL/issues/4166. In addition, I tried documenting my findings there if you could take a look. I have gotten slightly closer to the cause of the issue as I mentioned in the topic, but I could not think of a better fix yet than `gDLL->sendTurnUnready()` unfortunately.

However, I recently had an idea: I am testing treating the game as sequential for only the first turn after load, if it was previously simultaneous. This solves the only issue that I could find with your fix, that is with offline players during the load of the save in pitboss mode still lose their turns. Nonetheless this is considerably more challenging and requires a few more changes in the code, but it avoids the dirty `gDLL->sendTurnUnready()`.

@LoneGazebo I played a couple more games this weekend, using the latest version of the fix in my repo, and everything seems to be working perfectly (except for desyncs, which have nothing to do with this). I wish we had access to the netcode of civ5 so we could try fixing desyncs too. In addition, I am unsure a separate MP DLL would really be required, considering these fixes we are discussing would only activate when loading a MP Simultaneous game (unless @davenye is working on other MP things that I am unaware of).

LoneGazebo:
That's fine, but just make sure you're wrapping your code in MP-only specifics. I saw a bunch of pull requests earlier but I'm going to hold them until you're comfortable with the code.

LoneGazebo:
re: desyncs, at this point there are no RNG elements left in the DLL - any desyncs going on now aren't related to RNG, so there's either something else afoot that we're not seeing or there's just no hope for desync-free MP.

raphaell666:
>Even though HasReceivedTurnComplete is returning true for the local player on the first turn after load, HasSentTurnComplete is not

This is the same that I have experienced. Both `Player.HasSentTurnComplete()` and `Player.isEndTurn()` return false, while `gDLL->HasReceivedTurnComplete(player.GetID())` returns true. This is why I think the issue is in someway related to the graphical interface. I might very well be wrong though.

>I don't understand why PitBoss has the problem if the issue here is successfully being resolved by fixing the local player only

I believe that all players are marked as having sent (through the interface) that their turn has ended, when loading MP games. If they are not active at the time of the loading, they do not send the `gDLL->SendTurnUnready` of your fix as that only works for active players. Does this makes sense considering what you have experienced?

>that would explain why I think it is the humans that miss the turn, while you think the AI have a double turn

What I know for a fact is that the turn count does not increase when loading MP games and the AI playing twice before the human players. The turn count only increases after the human players play (after all the AI playing twice). This is why I say that what actually happens is that the AI plays twice in one turn, rather than the players losing a turn. The players do not lose a turn as the turn count does not increase until after they play. Makes sense?

>@raphaell666 , this version may appeal more to your sensibilities - davenye@85a5c32

I will test it out once I have time and report back.

raphaell666:
I think I successfully managed to merge those changes by merging the branch into which you commited them, to the master at my repository (still getting used to git and github):
https://github.com/raphaell666/Community-Patch-DLL/commit/087747065f7545b46a6bb1ba2a091d5683e47fc7

I will test these soon.

raphaell666:
@davenye With both fixes applied (SendTurnUnready and this newest one), I tried loading the same autosave I uploaded  in the other reply (this is a pitboss game) and the result was the same as before, inactive players at load still lose their turns (I am Siam, the other player is Portugal):
load_mp_save

These are the logs from loading this autosave, if they help:
[Logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/1940141/Logs.zip)

davenye:
Cheers, I thought as much but I figured you would still be interested since you were spending time making the original fix more elegant (this is an alternative to the original fix, not an addition).

raphaell666:
Fair enough. I think it's up to @LoneGazebo to decide which fix he thinks is best. I will take a closer look at this new fix (see if I have any new ideas) and test next time with only this newest fix enabled, to be sure it works.

LoneGazebo:
Not up to me here, I’m not the tester or the testee. :)

davenye:
Yeah, I found that out too when I ran a 2 player MP game. I was a bit hasty with letting you know about that possible fix. It does not work for any more than 1 player.

In other news, I finally ran a PitBoss game today for the first time ever. Unfortunately, I only have enough machines to add a solitary human.

I was able to gather enough information to implement what I believe to fix the double turn issue for PitBoss also. It works for one player (heh) but I am pretty sure it works for any number. Once I have cleaned up my code a little I will send you a link to try it out. I will warn you that it is even less elegant than before! It is an addition to the old fix, not a replacement.

@raphaell666, I incorporated your double turn fix (the one in setTurnActive, before the turn publish) and noted that it is indeed an improvement over my original fix. I am not sure if I messed up my testing but I had not seen the difference in the flashing player icon previously. As a result, I built the PitBoss fix upon your version. I think you should submit your fix as a pull request for the mp_bugfixes branch.

raphaell666:
>Once I have cleaned up my code a little I will send you a link to try it out.

Thank you, I will be waiting. It being elegant or not at this point is irrelevant, let's first get it working. At least with C++ I can help more than with LUA (still have no idea what wizardry you used to understand LUA+XML and make AdvancedSaves, props for that).

>I think you should submit your fix as a pull request for the mp_bugfixes branch.

I think I know how to do that without breaking things up, I will try to do that later today. It is still your fix though, I just improved it a tiny bit. Glad to be of help.

davenye:
After going though all this we still have one **major** problem to resolve. I still think the humans are having their turn *skipped*! 🤣 

What is worse is that I am using the same logic you are to come to that conclusion...It is most vexing.

When I load, let's say the turn is 10.
Without the fix, the AI has turn 10, then 11 and then humans get turn 11 (and things progress as expected from there), i.e. humans had turn 10 skipped.
With the fix, the AI has turn 10, then the humans have turn 10 (and things progress as expected from there).

This is all in keeping with how the fix prevents the first human turn from finishing before they have a chance to do anything.

Are you seeing this in both PitBoss and Standard MP? I only ever play Standard and usually the host.

raphaell666:
>Without the fix, the AI has turn 10, then 11 and then humans get turn 11 (and things progress as expected from there), i.e. humans had turn 10 skipped.

What I noticed without the fix is that AI literally plays twice in turn 10, and then the humans get their turn still in turn 10. Only after the humans play it advances to turn 11. I may be mis-remembering. I guess it also depends on which version of these recent changes you are using. I will try deactivating all mods later and verify this again.

>Are you seeing this in both PitBoss and Standard MP? I only ever play Standard and usually the host.

One thing we need to get out of the way is what is actually Pitboss in Civ5. What I usually play is either with an actual dedicated server (which is the actual Pitboss mode imo), or in "Standard MP/Pitboss Internet" or whatever the interface calls it, which is simply standard MP but with the Pitboss option active (this option does not let AI jump into the players place when they are not active/disconnected). 

Choosing "Pitboss" in the game interface menu when hosting the game using the client you will actually play with, is the same as choosing normal "Multiplayer" with the Pitboss option active from my experience.

Recently I noticed that using an dedicated server did not help at all (makes the turn slower even, good job to Firaxis), so I have been playing in standard multiplayer with the Pitboss option active.

PR #4056, opened 2018-03-14
fix a recently introduced regression

ilteroi:
stupid bug but quite severe ...


LoneGazebo:
Savegame compatible, looks like?

PR #4067, opened 2018-03-16
Shiny

LoneGazebo:


LoneGazebo:
@ilteroi since this breaks savegame I'm going to sit on it for a bit. If there's any more kinks to work out we can do that and then I'll push out a beta branch.

Issue #4117, opened 2018-03-29
Puppets don't get reduced Tourism


HungryForFood:
Similar issue to #4112. In this case, it is shown in the UI but the actual yield is not reduced correctly.

I looked in the codes, and the problem is that the unified yield project is only half complete for tourism; the tourism calculations for cities (in CvCultureClasses.cpp) does not actually take into account getBaseYieldRateModifier (from CvCity.cpp), which is also where the puppet penalty is calculated.

I've fixed the problem on my end as I needed to use some unified yield stuff for tourism. Will post a pull request (along with other stuff for the civ I'm working on) in the next few days if nothing goes wrong with the game.

Posting this here so we don't go about fixing the same thing.

LoneGazebo:
Let me know how this works out!

PR #4129, opened 2018-04-01
Piux IX

HungryForFood:
Looks like I just missed the release. Anyway, a brief summary:

### Unified yields for Tourism
- Linked the unified yield functions to Tourism, including puppet penalties (see #4117).
- UI changes for Tourism tooltip as well.

### Extra missionary spreads on prophets
- Fixed the vanilla BNW bug where naturally born prophets did not get the 1 additional spread, while faith bought ones did.
- Can be toggled off using BUGFIX_EXTRA_MISSIONARY_SPREADS.

### Local faith purchase cooldown
- Similar to purchase cooldown, except for faith. Affects the city only.

### UI fix for Trait_YieldChangesWorldWonder
- Previously the additional yields did not show up anywhere in the UI (yields themselves added correctly).

### Trait version of shared religion tourism modifier
- Self explanatory.


LoneGazebo:
Cool. I did my own temp fix for puppet yields on tourism, so I'll merge these later (esp. since your new trait table will break savegames).

PR #4189, opened 2018-04-15
Fixed blank names in AD deal withdrawn notes

davenye:
Now using getName() since nicknames are not populated for AI players, it
seems. getName() is used with much success all over the codebase.

There are still many instances of getNickName being used elsewhere. They obviously must work sometimes but it is definitely not always the case. I am of the opinion that getName is the better choice since it looks to be coded to handle all situations. Your thoughts?


LoneGazebo:
I'm honestly not sure, I haven't looked into it. I think the only consequence would be if a player changes their civ's name in the custom setup screen, but that's rare.

PR #4217, opened 2018-04-20
PolicyGEorGM fixes

HungryForFood:
- Fixed missing null check (similar to b490ec2).
- Removed from two places where it doesn't belong.
- Tidied up comments.


HungryForFood:
Question: Should I be merging my own pull requests?

Iamblichos:
Why did you remove this from the adopt Ideology?

HungryForFood:
I'm just following whenever there is a doInstantYield with INSTANT_YIELD_TYPE_POLICY_UNLOCK, as I'm assuming they cover all instances of what one would consider policy unlock.

Are you aware if anyone is using policy GE or GM by the way? When I first made changes to it, I assumed no one was using it since it wasn't used in VP. Realised that might not be such a good assumption now.

Iamblichos:
Yeah I use it in my own mod. I created it.

Iamblichos:
It's main intent is to give GE or GM points and change their progress. INSTANT_YIELD_TYPE_POLICY_UNLOCK everybody gets IIRC.

HungryForFood:
Oops. I'll do it your way then.

So right now gameplay wise the main change I did was that gold yield is 4x and production is 2x. You want that, or you want me to change it back? I've been thinking of only having the GP points for Roosevelt, the yields seemed quite snowball-y, so I'll leave it the way you want it.

Iamblichos:
We can nerf it a little if you want. America is a little overpowered.

HungryForFood:
It's up to you what you want, I'm thinking of moving away from the policy GE or GM column.

You sure policy GE or GM code should go into the ideology stuff? It should trigger at the exact same conditions as the policy unlock instant yields, right?

I think I'll also put all the GE or GM code in one function somewhere, and reference it when we need it, instead of having the exact same stuff up to 4 times.

Iamblichos:
It’s not really related to instant yields per se. Yeah my intention was when America adopts a policy, any policy, ideology or regular. Most of late game policy adopting is ideology.

HungryForFood:
I tested, with how it is right now, adopting ideological tenets will trigger the bonus, so I'm assuming DoChooseIdeology is for when you make the choice between the three ideologies, and not for selecting the tenets.

If you still want that, then sure, I'll add it to DoChooseIdeology. Give me a day or two, I should be able to do it by then.

Issue #4222, opened 2018-04-20
Acropolis yields does not match description


stage77:
_1. Mod version (i.e Date - 4/23): 4/20

_2. Mod list (if using Vox Populi only, leave blank):_

_3. Error description:_ Yields for Dye and Silk does not match the description (+1Gold,+1 Culture). Instead it provides the double amout of it. See screenshot.

![acropolis](https://user-images.githubusercontent.com/37305571/39057075-b11a8518-44b8-11e8-9fe4-e16570e3e81c.jpg)

_4. Steps to reproduce (optional):_

---------------------------
Supporting information:
Please note that you can attach .zip files by dragging-and-dropping them. If possible, zip up all supporting data and post that way.

1. Log files (always attach your Logs folder, located at My Documents/My Games/Sid Meier's Civilization 5. Make sure you have enabled logging before experiencing an error! Go here to find out how: http://forums.civfanatics.com/showthread.php?t=487482):

2. Save game (always attach a save that was made a turn before the error; located at My Documents/My Games/Sid Meier's Civilization 5/ModdedSaves):

3. CvMiniDump.dmp file (attach if experiencing a game crash. Located at Program Files/Steam/steamapps/common/Sid Meier's Civilization V):

4. Screenshots (optional):

Infixo:
Seems like a double entry in a table.

Issue #4220, opened 2018-04-20
Put a semicolon.


Iamblichos:
UPDATE UnitPromotions
SET Help = 'TXT_KEY_PROMOTION_FASTER_HEAL_HELP'
WHERE Type = 'PROMOTION_FASTER_HEAL'

Forgot the ;

Infixo:
Fortunately it's the last statement in the file :)

raphaell666:
Oops that was actually my mistake here: https://github.com/LoneGazebo/Community-Patch-DLL/issues/4178

Sorry. :smiley:

PR #4224, opened 2018-04-20
Fixes for #4220 and #4222

Infixo:


LoneGazebo:
Danke. Did you merge @iamblichos fix? If so I’ll hotfix today.

Infixo:
If you mean that Tradition updates then no.

Iamblichos:
I've got a number of little issues I've brought up today. As far as the Policy stuff goes for Tradition, Autocracy, Order, and Freedom, I'll change the DLL and Infixo can do the SQL code.

Iamblichos:
@LoneGazebo looks like you already have a check for the Tradition buildings in CvBuildingClasses.cpp CapitalOnly, so we won't need a Policy check.

PR #4236, opened 2018-04-21
refreshing some cached data on make peace

davenye:
This data depends on the warring state but is only updating in the event of war being declared. It seems to me it should be updated on peace too, keeping in mind that the parties could still be at war with other players.

Also renamed arg since it is being called for peace now too.

Code was simply copied from declareWar but am not fussed at the moment. In an upcoming pull request I remove some of this code since I don't think it is particularly as effective as it could be. I plan on cleaning up once proven effective and agreed as the right course of action.


LoneGazebo:
Looks good to me.

PR #4246, opened 2018-04-21
fix runtime error in mplist.lua with observer slots

davenye:
The fix is a bare minimum. Each observer will have their own "Observers" label, which seems odd but I guess that is the way it is - better than the whole list not being visible at all.

Is the file in the correct location though?

To repo, just create a MP game with observers. The observer slot may need to be empty. I had 1 human and 2 AI. The player list in the UI will not be present and there will be a error in the lua log.


davenye:
@LoneGazebo, I believe I have done this now (and the ModBuddy stuff for 3 other PRs)

PR #4248, opened 2018-04-21
fix bypass MP check when overwriting savegame 

davenye:
Fixed bug where overwritten saves save differently

There is deliberate Firaxis code which means manual multiplayer savegames are only ever copies of the most recent autosave. Overwriting save games was going through a code path that avoided the checks that enforced this. This has been fixed and overwritten behavior is consistent with the normal behavior.

This bug is the reason why people have been able to avoid the "double ai turn" problem. However, it is quite possible that this was having other unwanted effects on the game when loaded again. I have not witnessed any problems but I suspect they are possible and Firaxis certainly thought so themselves, hence the checks! I believe Firaxis originally only had autosaves in MP and this is was their quick fix. They could have easily not put in the checks so there is almost certainly some perceived issues with creating fresh saves mid-turn.

Unfortunately, some people may be bothered by not having this gleature, so...

Added code to emulate old-overwrite save bug that allow saving of mid-turn state. Looks nasty. Is nasty.

Hopefully this can be removed. Only provided in case the bug was being relied upon for reasons in addition to working around the "double ai turn" issue.

I am not sure about your philosophy in regards to these kinds of matters.
I am also not sure if I have the files in the correct location.

I have a upcoming pull request that aims to give people more control over their saving their games.

### DIFF
--- a/Community Patch/Core Files/CoreLua/SaveMenu.lua
+++ b/Community Patch/Core Files/CoreLua/SaveMenu.lua
@@ -1,0 +1,662 @@
+include( "InstanceManager" );
+include( "IconSupport" );
+include( "SupportFunctions" );
+include( "UniqueBonuses" );
+include( "MapUtilities" );
+
+local g_IsSingle = true;
+local g_IsAuto   = false;
+local g_IsDeletingFile = true;
+
+-- Global Constants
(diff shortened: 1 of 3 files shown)


davenye:
Indeed and it would have been easier to not put in the special code copy autosaves, so they must have at least thought it was required.

PR #4255, opened 2018-04-22
Save Diplo Reqs and Proposed Deals

davenye:
ACTIVE_DIPLOMACY specifc code was removed from writing out proposed deals.Comments stated that it was not writing out requests until issues were resolved. I am not sure what the issues were but I may have addressed them. It seems to work OK, albeit still a little odd because I need to clean them out at the start of the turn.

Diplo requests are now being written out also.

I don't think this change has much effect unless you are writing new save games in the middle of the turn, such as "double saving" or using the "mp post autosave" feature I will be submitting shortly. I guess some associated changes (sending the requests to non-active players) might prevent some weird desyncs but probably not.

I don't think I have tested this enough yet for a proper release but I have not had any issues arise. I will be monitoring the effects. I did uncomment some commented code and it does make me nervous as to why it was commented out in the first place!

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvDealClasses.cpp
+++ b/CvGameCoreDLL_Expansion2/CvDealClasses.cpp
@@ -5861,32 +5861,6 @@ FDataStream& operator>>(FDataStream& loadFrom, CvGameDeals& writeTo)
 	loadFrom >> uiVersion;
 	MOD_SERIALIZE_INIT_READ(loadFrom);
 
-#if defined(MOD_ACTIVE_DIPLOMACY)
-	if(GC.getGame().isReallyNetworkMultiPlayer() && MOD_ACTIVE_DIPLOMACY)
-	{
-		// JdH => savegame compatible load
-		loadFrom >> iEntriesToRead;
-		for (int iI = 0; iI < iEntriesToRead; iI++)
-		{
-			loadFrom >> tempItem;
-			if (CvPreGame::isHuman(tempItem.GetFromPlayer()) && CvPreGame::isHuman(tempItem.GetToPlayer()))
-			{
-				// only load human to humand deals until other problems are fixed
-				writeTo.m_ProposedDeals.push_back(tempItem);
-			}
-		}
-	}
-	else
-	{
-		writeTo.m_ProposedDeals.clear();
-		loadFrom >> iEntriesToRead;
-		for(int iI = 0; iI < iEntriesToRead; iI++)
-		{
-			loadFrom >> tempItem;
-			writeTo.m_ProposedDeals.push_back(tempItem);
-		}
-	}
-#else
 	writeTo.m_ProposedDeals.clear();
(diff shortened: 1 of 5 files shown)


davenye:
I think that might be just the way it is due to the implementation. JdH comments on how they think they should have implemented the system using notifications rather than diplo requests and planned on doing so in a new version.

Maybe it can be fixed but it might not be pretty. I'll have a looksee.

davenye:
@LoneGazebo, I don't know if this is what you want but if you change:
`NOTIFICATION_PLAYER_DEAL_RECEIVED		DiplomacyPlayer`
to:
`NOTIFICATION_PLAYER_DEAL_RECEIVED		Diplomacy`
in:
`(6a) Community Balance Overhaul - Compatibility Files (EUI) (v 1)\LUA\NotificationPanel.lua`

...then in my _extremely_ brief test, all the previously orange (whatever they are) icons become the gold globes with wreaths, including:

- War declaration diplo ("you should have known" etc) notification
- Protection pledge notification
- Trade notification

LoneGazebo:
I did the notification system after JdH disappeared, as we needed a way to keep the player from skipping pending deals. 

The orange is just the default texture for civ icons. Anything is better than that, so if you’re happy with it that’s fine by me.

davenye:
I actually think prefer the orange thing but maybe I am just used to it now.

I had a bit more of an investigation into the icons and I now believe it works by magic and reaffirmed my vow to avoid Civ Lua/UI at all costs. Might be a snap for bc1 though.

LoneGazebo:
In essence, the notification passes a bit of extra data through to the UI - even though I'm certain I'm passing the correct information on to the UI, it isn't able to use the extra bit to get the player's ID for their country icon. Oh well.

davenye:
I took another look since it sounds like you and I were having different problems with the code.

I appear to have been successful but I will do some more testing before rejoicing. It does look better:

image

You can see:
- War Diplo (fixed)
- War Declaration
- Trade (fixed)
- Protection Pledge

LoneGazebo:
Agh, you fixed it! What did you do?

raphaell666:
Just an update, the code for it to show the actual civ icons when humans sent diplo reqs between them has not been commited yet, right? 

In the current version we are just getting the orange ball thing (which is a huge improvement compared to the game default of literally nothing, thanks a lot!).

davenye:
Nah, not yet. I will send in through in the next few days along with a C++ update. The code is a little different too.

In MP, the orange ball was already there for AI initiated diplo thanks to G, as he informed us above.

In other earlier changes I got the human-human deal notifications working but they are just the gold globes with the wreaths.

In this change I have the icons with the proper civ background (although tbh there was existing code elsewhere that worked, so I am not sure about who fixed what and when).

In my dev branch I now have code to get human-human deal proposals with the icons shown as above. Massive hack, though. No idea why it wouldn't already work. Human deal accept/reject/counter-offer will remain as gold globes.

PR #4256, opened 2018-04-22
Enabled Events in MP

davenye:
Added code to send Event related messages to remote players. However, this is done via the foul means of hijacking the DiplomacyFromUI message and giving it a dual purpose. Event messages are have a flag encoded into one of the arguments and this is detected and handled appropriately on the clients. I understand this is unappealing but I believe it to be safe and the nastiness is tucked away.

Seems to work fine but maybe some events are problematic as I wouldn't have encountered all variations of mechanisms in limited testing (only 2 players!).

Added code is nothing fancy and can be improved if we want to send more
of messages in this manner, such as making the Trade Route cache invalidation when opening the TR popup propagate to other clients.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvCity.cpp
+++ b/CvGameCoreDLL_Expansion2/CvCity.cpp
@@ -44,6 +44,8 @@
 #if defined(MOD_BALANCE_CORE)
 #include "CvBarbarians.h"
 #endif
+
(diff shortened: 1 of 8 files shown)


LoneGazebo:
Neat! Never got around to tinkering with this, excellent work.

PR #4257, opened 2018-04-22
Option to keep unmet players unknown in pregame when loading MP game

davenye:
If enabled, details of players that have not been met by the local player will be hidden on the Staging Room screen when loading a save. This is in keeping with the details of unmet players not showing during the game itself. If you have not met the players then the icons will show as the question marks and the leader will be "An Unmet Player". I did make it so that the tooltip revealed the actual leader - I thought I may as well because if someone really wants to discover the truth can do so through other means anyway (such as switching selected player or checking the logs). The information is tracked independently for each player and works even when the player is switched.

This was implemented since I don't like catching a glimpse of the real players when loading a save.

I could not see how to extend the PreGame Lua API. Instead, I had to appropriate an existing PreGame function and give it a (nasty and confusing) second purpose in order to be able to get the Lua to query the known civs. Due to this version of Lua not appearing to have bitwise ops I had to encode the request in a different manner to how I did the MP Events but it is safe too. 

Not sure where you want the Lua file. I have merged in the EUI icon cache fix also.

I don't know where to add the game option text key either. I have it
done in my modpack but I can't find it in the project. How do I go about this?

I will post a screenshot or two later.


davenye:
Example:
image

If not already met, the other human would be "An Unmet Player" also.

raphaell666:
If I understood it correctly, this would only be for the AI? Would it also be possible to also hide the civilization that the other human players are choosing?

davenye:
@raphaell666 , you will need to add update your XML to show the game option on your game setup screen. I don't know how to check in such changes.

Also, this only affects loaded games. You will still be able to see who the humans select while setting up the game. It would be a useful feature to hide who humans select and I think quite easy to achieve. Do you think such a feature should be incorporated into this change and enabled using the same game option or a different one?

davenye:
OK, I have added code to hide other players details when setting up the game (if this option is enabled, didn't add a new option).

However, if the human selecting Random Leader, it will show up as such. I figured it couldn't hurt and will show that they are selecting Random and not picking their favourite. Of course, anyone wanting to engage in unscrupulous behavior could still do so.

I also made it so teams can see each other regardless.

AND I fixed some bad merging. I am blaming my G303 mouse double clicking when clicked once (and my poor QC).

davenye:
@LoneGazebo, I see you merged into the mp_bugfixes branch. How are we to check the required XML additions so that new Game Options show up in the UI. Currently I just hack them myself as I can't see anywhere to do it in the project.

Also, are we sure I have these .lua/.xml files in the right place and all the right places?

raphaell666:
Update: This is working perfectly well. Really great QoL addition, I will even set it up with enabled as default so that I don't forget to activate it in new games.

I was always annoyed when loading games with big maps that we could see what all the other undiscovered civilizations were in the game. @LoneGazebo I think this could easily be added into the master single-player branch as well (unless in single-player it already works like this, which I doubt).

davenye:
Yeah, it annoyed me too (obviously!) for the same reason.

I don't think there are any places in single player where this is useful - I thought it would have been but when I went to go code it I couldn't find anywhere that the civ details were present in SP pregame.

I noticed when I tried a dedicated server that you can see all the civ details for all players. I quickly coded a fix to just hide them if the option was enabled but then I saw all the player details shown anyway in the in game in game MP List and wondered if I was doing the right thing. In the end I decided to just leave it as it was.

PR #4258, opened 2018-04-22
Added second autosave point and save events

davenye:
Due to the barrier on the finishing of AI processing, it looks like it *might* be safe to slot in a second autosave point after AI has finished but before humans. This means that when loading, players do not need to wait while they watch the AI have their turns and (from memory) has the added bonus of not allowing the AI to have two turns.

Unfortunately, saving at this point can result in a slightly unpleasant stuttering of the game. Nothing too bad but some may not like it and not want it enabled. Also because of the was MP saves work (copying the most recent autosave) the post AI save will be used for manually saved games also.

In order to give people better choice as to when to have the post AI save enabled I have also added a couple of GameEvents which allow the various autosaves to be enabled/disabled on an individual basis, turn by turn. One event is a TESTANY which allows Lua code to request that an autosave be made at this point. The other is a HOOK which just notifies listeners that an autosave has been made. I think that with both of these simple events available modders can have a good degree of control.

I have implemented a separate Lua UI mod that gives fine grained control over saves as a proof of concept, at least. It is pretty disgusting and wouldn't be included in VP in the current state. It is something that is probably best left to the mod modders and now the tools are there. I found developing Lua Civ mods an excruciating experience and do not wish to engage in much more of it.

Unfortunately, much to my surprise I discovered that single player autosaves work in a different manner and this code doesn't have the desired effect. It seems silly saying it (so much so that I will need to go back and check) but I think there are autosaves in single player that are done by the EXE itself. My code doesn't break in single player but I need to see what I can do to make it more sensible.

I am interested in seeing if this is of any use to anyone...

### DIFF
--- a/CvGameCoreDLLUtil/include/CvEnums.h
+++ b/CvGameCoreDLLUtil/include/CvEnums.h
@@ -1312,6 +1312,10 @@ enum GameOptionTypes
 #define GAMEOPTION_RANDOM_VICTORY			"GAMEOPTION_RANDOM_VICTORY"
 #endif
 
+#if defined(MOD_SAVE_CONTROLLER)
+#define GAMEOPTION_POST_AUTOSAVES			"GAMEOPTION_POST_AUTOSAVES"
+#endif
+
 enum MultiplayerOptionTypes		
 {
 	NO_MPOPTION = -1,
--- a/CvGameCoreDLL_Expansion2/CustomMods.h
+++ b/CvGameCoreDLL_Expansion2/CustomMods.h
@@ -40,6 +40,11 @@
 //END MULTIPLAYER INSTRUCTIONS
 ////////////////////////////////////////
 
+// Sorry, I am putting this here since I don't quite know where to put it! Please advise and/or move
+// Enables autosaving GameEvents and allows and extra autosave after Barbs/AI turn in
+#define MOD_SAVE_CONTROLLER	(true)
+
+
 ///////////////////////
 // BATTLE ROYALE CODE
 //////////////////////
@@ -1292,6 +1297,9 @@ enum BattleTypeTypes
 #define GAMEEVENT_CityFlipped				"CityFlipped", "iii"
 #define GAMEEVENT_CityFlipChance			"CityFlipChance", "ii"
 #define GAMEEVENT_CityFlipRecipientChance	"CityFlipChance", "iii"
+// AutoSave
+#define GAMEEVENT_WantAutoSave				"WantAutoSave",		"i"
+#define GAMEEVENT_AutoSaved					"AutoSaved",		"ib"
 
 // Serialization wrappers
 #define MOD_SERIALIZE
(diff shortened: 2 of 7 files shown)


davenye:
For illustrative purposes, here is a couple of screencaps of the mod I made using this feature.

The Save Dialog has some more options to let people play around with various ways to get their game saved and also avoid getting double turned by the AI:
image

A new Advanced Save Options Dialog allows autosaves to be configured differently depending on their role in the game (not that I have run PitBoss/Observer mode yet):
image

As I understand it, some people find the autosaving process bothersome so with this they should be able to make it bend to their will. I admit some of this is of questionable utility.

I am not suggesting that any of these features be put into VP itself.

LoneGazebo:
Any particular reason it shouldn’t be in VP, though? We don’t want to fracture things too much.

raphaell666:
Interesting, I will see if I can test this out.

raphaell666:
Aren't there too many #endif for some reason at the end of the CvGame.cpp file?

raphaell666:
@LoneGazebo 

Are these `#endif` supposed to be at the end of the file?
https://github.com/LoneGazebo/Community-Patch-DLL/blob/4f2f4745e6382c59a1e76c16b7ea837eb998f377/CvGameCoreDLL_Expansion2/CvGame.cpp#L14186-L14187

One of the endifs closes this if:
https://github.com/LoneGazebo/Community-Patch-DLL/blob/4f2f4745e6382c59a1e76c16b7ea837eb998f377/CvGameCoreDLL_Expansion2/CvGame.cpp#L13958

Which is within this one (which in turn is closed by the second endif):
https://github.com/LoneGazebo/Community-Patch-DLL/blob/4f2f4745e6382c59a1e76c16b7ea837eb998f377/CvGameCoreDLL_Expansion2/CvGame.cpp#L13942

In other words, the new free city functions (e.g. `GetPotentialFreeCityPlayer`) are within the `#if defined(MOD_BALANCE_CORE_RESOURCE_MONOPOLIES)` condition. I am not sure but this does not make much sense to me.

Iamblichos:
@raphaell666 I think you are right about those #endif's.

LoneGazebo:
accidentally merged this...whoops.

davenye:
The autosave events should be in VP. I say my mod shouldn't be since I figure it is kind of weird and could generate unnecessary burden supporting.

Importantly, my mod code is absolutely terrible and I can't see myself wanting to clean it up and maintain it. I don't want to dump that on anybody.

@raphaell666, I will put my mod up on my github for you if you are wanting to try it. Without it, you can only turn post saves on and off via a game option. Additionally, the game option won't even be there since I can't seem to find out where I can check new game options in. I had to manually modify some when building my modpack for testing.

raphaell666:
Thank you, I will see if I can get it working. If necessary I will open issues at that repository, then.

PR #4269, opened 2018-04-23
PolicyGEorGM fixes no 2

HungryForFood:
- Now uses a common function.
- Added back to DoChooseIdeology.


HungryForFood:
@Iamblichos I tested in-game, DoChooseIdeology seems to be only for the AI to evaluate which ideology to take, I didn't get the effect when I adopted an ideology myself. I left the PolicyGEorGM stuff in there though.

PR #4285, opened 2018-04-24
Revert "Added second autosave point and save events"

LoneGazebo:
Reverts LoneGazebo/Community-Patch-DLL#4258


LoneGazebo:
I merged whatever was on mp_bugfixes. If that wasn’t the latest it needs to be pushed there and then I’ll update next week. Probably would be worthwhile to start a testing thread on civfanatics to get n+1 feedback.

davenye:
I have few fixes/changes to go in before the civfanatics testing. Sorry, I have been on the road a lot recently but that is out of the way and will be able to get on and finalise things now.

Also, I found and fixed an existing bug that would have been causing a **lot** of trouble for a lot of people (or none, depending on how you play). Will provide an update soon.

PR #4330, opened 2018-05-03
Added extra condition for the ending of the first MP turn to prevent double AI turn bug with Pitboss

davenye:
Addresses double turn issue for pitboss games when players are not
initially present when the host starts.

Implemented a new message that intends to do some of the same job that sendTurnComplete is supposed to do and will prevent human turns on remote machines ending before they should.


davenye:
Getting some strange behavior (can't end turn sometimes) that I am pinning on this change, gonna revert it until I can figure it out.

PR #4331, opened 2018-05-03
fix bug where plot reveal was dependant on active player

davenye:
Issue when revealing plots during plot visibility count update.

Return value from CvPlot::setRevealed is more nuanced than simply
returning whether the plot is being revealed to the team.

Now using a simple check for when deciding on whether the explore plots list needs amending.

This will be responsible for a large number of desyncs for people who autoexplore. When autoexploring the unit could choose a different plot to explore since it is selecting from a different set of potential plots as the active player has reveal plots that the other machines do not.

This bug causes all sorts of problems, particularly desyncs in MP.
While the position is possibly synced by host, the visibility of the
plots is not. This means things like barb camps get spawned in different
locations on different machines, which is a big problem. I think bug is the catalyst for a great amount of drift of all sorts.

Also, it causes the AI to explore sub optimally...gonna double check this though!

Will result in slightly more processing but it is the right thing to do.

EDIT:
Just remembered that I need to add the preprocessor code for conditional compilation of the bugfix.

Also attempted to be more coherent in the above text.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CustomMods.h
+++ b/CvGameCoreDLL_Expansion2/CustomMods.h
@@ -971,6 +971,8 @@
 #define MOD_BUGFIX_EXTRA_MISSIONARY_SPREADS			gCustomMods.isBUGFIX_EXTRA_MISSIONARY_SPREADS()
 // Workaround for the AI double turn when loading MP games with simultaneous/hybrid turns
 #define MOD_BUGFIX_AI_DOUBLE_TURN_MP_LOAD (true)
+// Fixes issue where explore plot list was not being consistently updated when a plot was revealed
+#define MOD_BUGFIX_EXPLORE_REVEALED_PLOT	(true)
 
 #endif // ACHIEVEMENT_HACKS
 
--- a/CvGameCoreDLL_Expansion2/CvPlot.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlot.cpp
@@ -11196,13 +11196,28 @@ PlotVisibilityChangeResult CvPlot::changeVisibilityCount(TeamTypes eTeam, int iC
 	if (bOldVisibility == false && isVisible(eTeam))
 	{
 		eResult = VISIBILITY_CHANGE_TO_VISIBLE;
-		
-		bool alreadyRevealed = isRevealed(eTeam); // result from setRevealed is not just a simple "we reveled a new plot to the player" and was causing all sorts of issues (poor ai explore decisions, desyncs, etc).
-#if defined(MOD_API_EXTENSIONS)
-		if (!setRevealed(eTeam, true, pUnit))	// Change to revealed, returns true if the visibility was changed
-#else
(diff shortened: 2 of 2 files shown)


LoneGazebo:
@davenye not a great place to post this but there’s a 32-64 bit sign issue when you compile your DLL changes for 43 civs - compile for 43 and you’ll see it.

LoneGazebo:
I didn’t get a chance to fix it before I left

davenye:
@LoneGazebo, I am pretty sure I know where the problem is going to be but I cannot see how to compile for 43.

LoneGazebo:
@davenye custommodsglobal.h in the assets part of the project viewer.

Iamblichos:
@davenye  this is where it happens if you haven't had a chance to look.....need something like this....

```
void updateKnownPlayersTable()
{
	// Playing it safe here, don't want to introduce a bug due to not knowing all the code paths and trying to save a few cycles/bytes
	s_knownPlayersTable.clear();
	s_knownPlayersTable.resize(MAX_MAJOR_CIVS);
	for (int i = 0; i < MAX_MAJOR_CIVS; i++) {
		KnownPlayersBitArray bitarray = 0;
		const CvTeam& kTeam = GET_TEAM((TeamTypes)i);

		for (int j = 0; j < MAX_MAJOR_CIVS; j++) {
			if (i == j || kTeam.isHasMet((TeamTypes)j))
#if defined(MOD_GLOBAL_MAX_MAJOR_CIVS)
				bitarray |= 1LL << j;
#else
				bitarray |= 1 << j;
#endif
		}
		s_knownPlayersTable[i] = bitarray;
	}
}
```
```
bool isKnownPlayer(PlayerTypes eA, PlayerTypes eB) {
	// table will be empty if GAMEOPTION_KEEP_UNMET_PLAYERS_UNKNOWN, i.e. not keeping track of unmet players and consider all to be met for the purposes of PreGame UI.
	if (s_knownPlayersTable.empty()) return true;
	if (eA < 0 || eB < 0) return true; // erring on the side of caution
	if ((size_t)eA >= s_knownPlayersTable.size() || (size_t)eB >= s_knownPlayersTable.size()) return true; // erring on the side of caution
#if defined(MOD_GLOBAL_MAX_MAJOR_CIVS)
	bool bKnown = s_knownPlayersTable[eA] & (1LL << eB);
#else
	bool bKnown = s_knownPlayersTable[eA] & (1 << eB);
#endif
	return bKnown;
}
```

davenye:
My Visual Studio is still half busted (IntelliSense won't function) but I inspected the code and believe the fix is sound based on the fact that setRevealed was always returning true if the plot revealed status for the activeplayer was changing. Now I am doing that test for all players instead,

A strange thing I did notice was that a explore list update would be triggered if the plot went from revealed to *not* revealed. I am not doing that, maybe I should (easy change, obviously) but can this even happen? Gonna merge this anyway.

I have finally checked in the conditional compilation version (appears much higher up in this thread).

PR #4346, opened 2018-05-05
Franklin Roosevelt 6/5/2018

HungryForFood:
Not save game compatible, so merge when you're ready for a new version.

Summary of changes:

## Trait_YieldChangesPerImprovementBuilt
- Grants yields per turn to every city per number of the specified improvement built. Intended for great person improvements, as regular improvements can be rebuilt unlimited times.
- Can be set to odd era scaler.
- Adds to the code a method to count the number of times an improvement type has been built by a player.

## Trait_GreatPersonProgressFromPolicyUnlock
- Grants the specified value as percent of a great person progress in the capital when a policy is unlocked.
- Uses the new DoInstantGreatPersonProgress function, a counterpart to DoInstantYields. Should be simple to expand the function for more cases.
- Added Description and IconString columns to the GreatPersons data table, so that text is easier to do.


Hokath:
Cool, there might be other people interested in these functions over in modmod land.

Tell me, how easy would it be for me to follow what you did to make a sister function of Trait_GreatPersonProgressFromPolicyUnlock that behaves as Trait_YieldFromPolicyUnlock?

LoneGazebo:
Flavors matter but now I used some collated tables to the tell the AI if it is a warmonger, nerd, etc.

LoneGazebo:
These are done in the DLL.

Hokath:
@HungryForFood 
The code is commented that the Trait_YieldChangesPerImprovementBuilt can be set to only affect the capital, specifically to set IsCapitalOnly presumably in Traits along with IsOddEraScaler. Is this correct, because I don't see an IsCapitalOnly field in Traits.

HungryForFood:
Oops, outdated comment I missed. No that's not true (anymore), if you're making a commit, can you help me remove it, else I'll correct it later.

Hokath:
If it's not too much trouble I would very much like it!
I just noticed while browsing so I'll leave it to you.

PR #4364, opened 2018-05-10
Iamblichos

Iamblichos:
For review and testing if you have time.


LoneGazebo:
@Iamblichos it's easier for me to track merges if each branch only contains the changes it is making - I know you're merging this for your own testing, but it'd be easier for me to keep track of what each user is doing if this branch was purely your edits.

Iamblichos:
Yeah I shouldn't of sync'ed up with shiny and mp_bugfixes especially the latter. ilteroi a lot of code swapping so that's why we were syncing with each other.

Iamblichos:
This will get resolved when the master and my branch get on the same level and then I’ll only make my commits to this branch. I didn’t realize that syncing with other branches would cause such a hassle.

PR #4370, opened 2018-05-11
Revert "Added extra condition for the ending of the first MP turn to prevent double AI turn bug with Pitboss"

davenye:
Reverts LoneGazebo/Community-Patch-DLL#4330

I am getting some strange behavior where the first turn does not progress under certain conditions. I believe it is due to this change as I have not encountered it before now.

Reverting until I can figure it out.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvDllNetMessageExt.cpp
+++ b/CvGameCoreDLL_Expansion2/CvDllNetMessageExt.cpp
@@ -12,7 +12,6 @@ namespace NetMessageExt
 			None = 0,
 			DoEventChoice,
 			DoCityEventChoice,	// Could use the same id but I doubt we will use more than 254		
-			PlayerFirstEndTurn,
 		};
 	}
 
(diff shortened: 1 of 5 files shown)


davenye:
OK, I fixed the problem. I had removed some "extra" code I had originally had in during most of my testing but it turns out it was needed.

However, I am still concerned as to why the problem was actually occurring as it may have ramifications on the way I am handling events in MP. I am going to hang onto this change (since I think we are close to release) and investigate further. The Pitboss turn skip problem can be worked around by not hot joining in the first turn so it isn't a huge deal.

davenye:
It seems that the sendTurnComplete was getting sent (across the network) before the FromDiplomacyFromUI, due to avoiding a queue or something. I improved the sequencing of the calls to properly avoid the issue. I don't think would be an issue with MP Events in the end, unless there is a more widespread problem related to this stuff also.

I will test leave it in my dev branch for a bit for testing before submitting a new pull request.

PR #4379, opened 2018-05-13
Victoria 13/5/2018

HungryForFood:
Lots of changes to CvTradeClasses.cpp. May need a quick scan to make sure I don't break anything. Everything seems fine in-game though.

## Gold internal trade routes
- Very similar to international trade routes, with some small differences.
- Enable by setting BALANCE_CORE_GOLD_INTERNAL_TRADE_ROUTES to 1 in the CustomModOptions table (turned off by default).
- Can only be used by civilizations with CanGoldInternalTradeRoutes set to 1 in Traits.
- Includes changes to UI to accommodate for this.

## Some traits for increasing trade route slots
- These are TradeRoutesPerXOwnedCities and TradeRoutesPerXOwnedVassals, both in the Traits table.


LoneGazebo:
Let me know when you've confirmed that this works as intended (and doesn't break existing functionality).

HungryForFood:
Hey @LoneGazebo, is science from international trade routes supposed to be affected by distance? Code says yes, but the UI lua is done in a way that suggests no, and I somehow have memory of no too.

I am very sure now its just a UI bug by the way, and I know what to change to fix it.

LoneGazebo:
It is not supposed to be affected.

HungryForFood:
Culture should not be affected too then right?

Fixed the issues now, UI looks fine, and distance modifier no longer affects science and culture for international routes.

Not sure why Github is showing some files as changed even though not by me. These are the changes in 43 Civ, UIText.xml, the various CityStateDiploPopup.lua and .xml, and CityStateStatusHelper.lua.

LoneGazebo:
manually merged

Issue #4448, opened 2018-05-31
City attack preview.


MoiMagnus:
(see Vortok's screenshots here: https://forums.civfanatics.com/threads/tradition-guide-from-beginner-to-intermediate.631263/page-5)

The preview is misleading, since the strength displayed is the combat strength of the city, and not the effective ranged attack combat strength of the city.
(Which is combat strength, not counting the defensive buildings, and times 40%, as said here https://forums.civfanatics.com/threads/low-end-city-attack-damage.632942/page-2)

Iamblichos:
Well I don't think we've touched CvCity::rangeCombatDamage() in a long time? You sure this is a problem the ratio between that city and the unit is almost flat.

MoiMagnus:
I think that's a vanilla issue. The strength use for city attack is not the same as the strength used for city defense in the code. In the preview, the resulting damages is correct, but the "strength" displayed is the wrong one.

Iamblichos:
Nothing has changed their from vanilla in the EnemyUnitPanel.


local myCityMaxHP = myCity:GetMaxHitPoints();
	local myCityCurHP = myCity:GetDamage();
	local myCityDamageInflicted = myCity:RangeCombatDamage(theirUnit, nil);
	local myCityStrength = myCity:GetStrengthValue();

MoiMagnus:
So that's a vanilla issue. There is two different city strength:

> //    --------------------------------------------------------------------------------
int CvCity::getStrengthValue(bool bForRangeStrike) const
{
    VALIDATE_OBJECT
    // Strike strikes are weaker
    if(bForRangeStrike)
    {
        int iValue = m_iStrengthValue;
        iValue -= m_pCityBuildings->GetBuildingDefense();
        CvAssertMsg(iValue > 0, "City strength should always be greater than zero. Please show Jon this and send your last 5 autosaves.");
        iValue *= /*40*/ GC.getCITY_RANGED_ATTACK_STRENGTH_MULTIPLIER();
        iValue /= 100;
        if(HasGarrison())
        {
            iValue *= (100 + GET_PLAYER(m_eOwner).GetGarrisonedCityRangeStrikeModifier());
            iValue /= 100;
        }
        // Religion city strike mod
        int iReligionCityStrikeMod = 0;
        ReligionTypes eMajority = GetCityReligions()->GetReligiousMajority();
        if(eMajority != NO_RELIGION)
        {
[...]

Issue #4444, opened 2018-05-30
Citadels and other improvements that cause nearbyenemydamage()


Iamblichos:
bool CvPlot::IsNearEnemyCitadel(PlayerTypes ePlayer, int* piCitadelDamage) const
{
	VALIDATE_OBJECT

	int iCitadelRange = 1;
	CvPlot* pLoopPlot;

	ImprovementTypes eImprovement;
	int iDamage;

	// Look around this Unit to see if there's an adjacent Citadel
	for(int iX = -iCitadelRange; iX <= iCitadelRange; iX++)
	{
		for(int iY = -iCitadelRange; iY <= iCitadelRange; iY++)
		{
			pLoopPlot = plotXYWithRangeCheck(getX(), getY(), iX, iY, iCitadelRange);

			if(pLoopPlot != NULL)
			{
				eImprovement = pLoopPlot->getImprovementType();

				// Citadel here?
				if(eImprovement != NO_IMPROVEMENT && !pLoopPlot->IsImprovementPillaged() && GC.getImprovementInfo(eImprovement)->GetNearbyEnemyDamage() != 0)
				{
					iDamage = GC.getImprovementInfo(eImprovement)->GetNearbyEnemyDamage();
					if(pLoopPlot->getOwner() != NO_PLAYER)
					{
						if(GET_TEAM(GET_PLAYER(ePlayer).getTeam()).isAtWar(pLoopPlot->getTeam()))
						{
							if(piCitadelDamage)
								*piCitadelDamage = iDamage;
							return true;
						}
					}
				}
				// Unit here that acts like a citadel?
				else if (pLoopPlot->getNumUnits() != 0)
				{
					for (int iZ = 0; iZ < pLoopPlot->getNumUnits(); iZ++)
					{
						CvUnit* pLoopUnit = pLoopPlot->getUnitByIndex(iZ);
						if (pLoopUnit != NULL && pLoopUnit->getNearbyEnemyDamage() != 0 && GET_TEAM(GET_PLAYER(ePlayer).getTeam()).isAtWar(pLoopUnit->getTeam()))
						{
							iDamage = pLoopUnit->getNearbyEnemyDamage();
							if (piCitadelDamage)
								*piCitadelDamage = iDamage;
							return true;
						}
					}
				}
			}
		}
	}
	return false;
}

[...]

PR #4450, opened 2018-05-31
Fix CvLuaCity and EnemyUnitPanel to display the ranged combat strengt…

Iamblichos:
…h and approx damage for combat odds of City bombardment of enemy units. #4448 and #4444


Iamblichos:
Added commit adc3e78

Issue #4507, opened 2018-06-26
PUPPET_GOLDEN_AGE_MODIFIER


supermerill:
Hi

I was watching the puppet yield reduction stuff, and i see that PUPPET_GOLDEN_AGE_MODIFIER isn't modified inside the CoreDefines, and it's so at the default value of -50.

Intended?

I think, even if you want to keep it at -50, it's wise to put it also behind the other ones in the CoreDefines.sql for consistency.

Issue #4492, opened 2018-06-18
Incorrect puppet yields with Imperialism/Martial Law


Omen-of-Peace:
_1. Mod version (i.e Date - 4/23):_ 6/14 (modpack from the forum)

_2. Mod list (if using Vox Populi only, leave blank):_

_3. Error description:_ the text of the Martial Law policy (Imperialism tree) indicates that the puppet penalty is reduced to -50%. However, when checking a puppet, the penalty is -55%.
(I think it used to reduce penalties by 25% - maybe it was never changed even though the text was?)

supermerill:
20% + 25% = 45%

I don't understand your problem

PR #4508, opened 2018-06-26
#4492 and #4507: puppet yield inconsistency bugfix

supermerill:
I don't know what is the good value: the one in the text or the one in the database?


LoneGazebo:
Your fix looks right to me.

PR #4516, opened 2018-07-03
Tojo 3/7/2018

HungryForFood:
Not save game compatible.

## Trait_YieldFromTilePurchaseTerrainType

- Instant yield in the city when gold buying a tile of a specified terrain type.

## Trait_YieldFromTileConquest

- Instant yield in capital when conquering a city, scaling based on the number of tiles of the specified terrain type, rather than population.

## Trait_YieldFromTileCultureBomb & Trait_YieldFromTileStealCultureBomb

- Instant yield in the appropriate city when gaining neutral or enemy tiles respectively of the specified terrain type by culture bombing (citadel).

## Trait_YieldFromTileSettle

- Instant yield when settling, based on the number of tiles of the specified terrain type gained.


HungryForFood:
Found a few bugs. :(

PR #4584, opened 2018-07-24
Iamblichos

Iamblichos:


HungryForFood:
```
+	const std::vector<std::pair<int, int>>& possibleUnits = GET_PLAYER(getOwner()).GetAreaEffectPromotionUnits();
+	for (std::vector<std::pair<int, int>>::const_iterator it = possibleUnits.begin(); it != possibleUnits.end(); ++it)
+	{
+		//first quick check with a large, fixed distance
+		CvPlot* pUnitPlot = GC.getMap().plotByIndexUnchecked(it->second);
+		if (plotDistance(pUnitPlot->getX(), pUnitPlot->getY(), getX(), getY()) > 4)
+			continue;
+
+		CvUnit* pUnit = GET_PLAYER(getOwner()).getUnit(it->first);
+		if (pUnit && pUnit->getNearbyHealFriendlyTerritory() != 0 && pUnit != this)
+		{
+			iRange = pUnit->GetNearbyUnitPromotionsRange();
+			if (plotDistance(getX(), getY(), pUnit->getX(), pUnit->getY()) <= iRange)
+			{
+				if (pUnit->getGiveDomain() != NO_DOMAIN && (pUnit->getGiveDomain() == getDomainType()))
+				{
+					iHeal = pUnit->getNearbyHealFriendlyTerritory();
+				}
+			}
+		}
+	}
```

@LoneGazebo by vector table, you mean this part yes? Specifically the std:vector parts?

I don't know much about memory management, so how much memory does adding a new trait table cost compared to a vector table?

supermerill:
After looking a bit, no new vector. The ones used are here since more than a year.
@HungryForFood, your vector is just a reference on the one stored inside the CvPlayer. No worry, it does nothing.

> how much memory does adding a new trait table cost compared to a vector table?

the 3 entries you added 
```
	FAutoVariable<int, CvUnit> m_inearbyHealEnemyTerritory;
	FAutoVariable<int, CvUnit> m_inearbyHealNeutralTerritory;
	FAutoVariable<int, CvUnit> m_inearbyHealFriendlyTerritory;
```
cost each at least 64bits/8bytes of memory (maybe double of that, a sizeof() should give a more precise answer)
Multiply that by the number of units a huge map can have (~24*42 ~=1000) -> 24 kB of memory minimum

supermerill:
sorry, i just read the pr changes, without verifying.

Issue #4828, opened 2018-10-18
Tradition - Yields to buildingclass added via free building instead of by policy directly


pineappledan:
_1. Mod version: 10-10

_3. Error description: Tradition uses Building_BuildingClassYieldChanges instead of Policy_BuildingClassYieldChanges to give +2 culture to monuments/baths/gardens, and +1 science to councils and herbalists.  This causes an issue if a civ has the ability to capture buildings (Rome), because they can receive the full, global benefit of the landed elite and legalism policies. I  propose tying the buildingclass yields to the policy, rather than the building, so Rome can still get the bonus specialists and yields in city, but can't capture the full power of the policy.  

INSERT INTO Building_BuildingClassYieldChanges
	(BuildingType, BuildingClassType, YieldType, YieldChange)
VALUES
	('BUILDING_PALACE_GARDEN', 'BUILDINGCLASS_GARDEN', 'YIELD_CULTURE', 2),
	('BUILDING_PALACE_GARDEN', 'BUILDINGCLASS_BATH', 'YIELD_CULTURE', 2),
	('BUILDING_PALACE_GARDEN', 'BUILDINGCLASS_MONUMENT', 'YIELD_CULTURE', 2),
	('BUILDING_PALACE_ASTROLOGER', 'BUILDINGCLASS_GROVE', 'YIELD_SCIENCE', 1),
	('BUILDING_PALACE_ASTROLOGER', 'BUILDINGCLASS_HERBALIST', 'YIELD_SCIENCE', 1);

New version:

INSERT INTO Policy_BuildingClassYieldChanges
	(PolicyType, BuildingClassType, YieldType, YieldChange)
VALUES
	('POLICY_LANDED_ELITE', 'BUILDINGCLASS_GARDEN', 'YIELD_CULTURE', 2),
	('POLICY_LANDED_ELITE', 'BUILDINGCLASS_BATH', 'YIELD_CULTURE', 2),
	('POLICY_LANDED_ELITE', 'BUILDINGCLASS_MONUMENT', 'YIELD_CULTURE', 2),
	('POLICY_LEGALISM', 'BUILDINGCLASS_GROVE', 'YIELD_SCIENCE', 1),
	('POLICY_LEGALISM', 'BUILDINGCLASS_HERBALIST', 'YIELD_SCIENCE', 1);

LoneGazebo:
I don’t think Rome can capture those anyways.

[...]

PR #4829, opened 2018-10-18
Tradition.sql - small alteration

pineappledan:
Change Landed Elite and Legalism policies so the global bonus yields to garden/monument/baths/council/herbalist are tied to the policy instead of the policy's free building. 

No gameplay change, except it prevents civs with the ability to capture buildings from being able to grab the full benefit of these policies


pineappledan:
#4828

LoneGazebo:
They wouldn't be able to capture buildings anyways?

PR #4879, opened 2018-11-11
Overlooked Enginseer small changes?

johanbest69:


TechpriestEnginseer:
@ilteroi What's his changes? I can't seem to comprehend them, but it seems similar to your merges recently.

PR #5082, opened 2019-01-19
Meiji 19/1/2019

HungryForFood:
Tried using the map container instead of a fixed size array to save data for the new traits. Seems to work well, and reduces memory usage.

Not sure what's up with the commits.

## Trait_UnitCombatProductionCostModifiers

- Allows you to change the production cost of a unit combat (which will then change gold purchase costs and upgrade costs).
- Zero = no change; negative = less cost; positive = more cost.
- Can be toggled to only be in effect during golden ages.

## Trait_GoldenAgeYieldModifiers

- Simple trait counterpart to Building_GoldenAgeYieldMod.


LoneGazebo:
@HungryForFood if it saves memory, we might consider doing the same to all traits.

LoneGazebo:
Hard to say. Probably not relative to the big hitters like unit or buildingclass arrays.

HungryForFood:
@supermerill Aside from our labour time, there's a slight performance cost in accessing a map container as opposed to a vector/array, based on my Googling. Not sure if significant.

@LoneGazebo I think it should be possible to replace unit and buildingclass vectors/arrays with map containers as well?

PR #5233, opened 2019-03-20
Pius IX 20/03/2019

HungryForFood:
## Bugfix
- Update check for pseudo Great Artists to include new GAP mechanic.

Save game compatible.


Voker57:
What does "Pius IX" here mean?

HungryForFood:
That's just my branch name, which is named as such because the changes are for JFD's Pius IX for Vox Populi.

PR #5291, opened 2019-04-14
Ekeuhnick 14/04/19

HungryForFood:
## Trait_NonSpecialistFoodChange 
- Added Trait_NonSpecialistFoodChange (changes the food consumed by each non-specialist citizen).


pineappledan:
Yaaaaaay!

pineappledan:
So what do I need to change in the Inuit mod so that the trait still works?

HungryForFood:
When I coded it in, there was a small possibility that the food discount might depend on terrain, so I didn't want to redo everything if that's the case.

Another small thing is that the debug database looks neater in SQLite Spy.

Does it make a big difference?

LoneGazebo:
@HungryForFood,

Yes it takes up more memory to do it this way. 

G

PR #5440, opened 2019-06-15
Allow AI to make & use improved Approach guesses

HeathcliffWarriors:
**Note:** I tried my best to ensure I used the correct syntax but am not 100% confident, so I'd appreciate if someone could verify as I'm still new to this.

What this pull request does:

- Rewrites the DoUpdateApproachTowardsUsGuesses function to allow the AI to make an educated guess as to another player's approach towards them, without conflicting with any other function which sets a given guess. Even though this function isn't used anywhere to my knowledge, I was aiming to make the AI a bit smarter in terms of potential, since the original DoUpdateApproachTowardsUsGuesses function was pitiful. If you like the function, perhaps it could see use elsewhere :)

- Adds a new function to access the memory value for the AI's guess, GetTrueApproachTowardsUsGuess, since GetApproachTowardsUsGuess is used to show the AI's visible approach (originally, GetApproachTowardsUsGuess returned this memory value, but that was apparently commented out).

- When a human player makes a demand or chooses the insulting dialogue option, the AI will now guess that their approach towards them is HOSTILE.

Let me know if you have any feedback!

Edit: Also fixes the extra space in the Workshop description, and a couple comment fixes.

### DIFF
--- a/Community Balance Patch/Balance Changes/Text/en_US/BuildingText.sql
+++ b/Community Balance Patch/Balance Changes/Text/en_US/BuildingText.sql
@@ -516,7 +516,7 @@ SET Text = 'Provides a free Garden in the city in which it is built.'
 WHERE Tag = 'TXT_KEY_WONDER_HANGING_GARDEN_HELP' AND EXISTS (SELECT * FROM COMMUNITY WHERE Type='COMMUNITY_CORE_BALANCE_BUILDINGS' AND Value= 1 );
 
 UPDATE Language_en_US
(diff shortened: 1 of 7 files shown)


MoiMagnus:
> When a human player makes a demand or chooses the insulting dialogue option, the AI will now guess that their approach towards them is HOSTILE

Do demands include "asking for help to a friend"? Or are they treated differently? Or is it assumed that the player only ever do aggressive demands?

HeathcliffWarriors:
They are treated separately.

Issue #5488, opened 2019-07-09
Combat Preview Panel Bonuses Overflow Fix


Aqinu:
_1. Mod version (i.e Date - 4/23): 6/12

_2. Mod list (if using Vox Populi only, leave blank): 

_3. Error description: 

I've noticed that when you have above a certain amount of combat bonuses in the preview panel they begin to show over the dedicated area and you can barely see them.

To fix this problem, I don't know if it is possible, from what I've noticed the panel can hold up to 6 rows of bonuses (5 if the description of a bonus takes two rows) without overflowing, and to fix this it needs to be made to work something like this:
- if there are 5 bonuses then to be displayed as usual
- if there are more than 5 bonuses (even the panel can hold 6 rows I've accounted for a possible bonus with two rows which occupies), than all the bonuses to be added together and displayed into a single bonus (i.e. if you have 6 bonuses than the 5th and 6th are added together and displayed under a single bonus)
- and, of course, if in the event that the bonuses that were added together happens to give a negative bonus then to be colored accordingly in red

See the **screenshots** to see how I would like to look.

_4. Steps to reproduce (optional):_

---------------------------
Supporting information:
Please note that you can attach .zip files by dragging-and-dropping them. If possible, zip up all supporting data and post that way.

1. Log files (always attach your Logs folder, located at My Documents/My Games/Sid Meier's Civilization 5. Make sure you have enabled logging before experiencing an error! Go here to find out how: http://forums.civfanatics.com/showthread.php?t=487482):

2. Save game (always attach a save that was made a turn before the error; located at My Documents/My Games/Sid Meier's Civilization 5/ModdedSaves):

3. CvMiniDump.dmp file (attach if experiencing a game crash. Located at Program Files/Steam/steamapps/common/Sid Meier's Civilization V):

4. Screenshots (optional):

**Here how is currently, which doesn't help me too much cause I barely see them**
![Civ5Screen0034](https://user-images.githubusercontent.com/49090452/60897355-9fa5f080-a270-11e9-9a31-f898b3c659ea.png)

**Here is with the fix and at least, in this instance I can see how much the bonuses help me (I added the last three bonuses and displayed them under a single name**)
![Untitled](https://user-images.githubusercontent.com/49090452/60897127-332af180-a270-11e9-83f0-0c5dffcf0346.png)

ilteroi:
nice idea but hard to implement. 

[...]

PR #5492, opened 2019-07-12
Fix for issue #5488

Aqinu:
Fix for Combat Preview Panel Bonuses Overflow, I tested it and it does work.

**Additional changes:**
Also changed YOUR STRENGTH value to be rounded to one decimal after the point instead of two (**see screenshots 1 and 2** from **145.75** in **145.8**) this way doesn't go out of the panel.

**Known issues:**
One single problem I saw is that not always in the preview panel are shown 4 bonuses then the misc. one (as far as I tested didn't go less than 3 bonuses and one misc. bonus but remember that **misc. one doesn't appear when there are 4 bonuses or lower**), thats because some bonuses stack together into one name but each one of the bonuses is counted (i.e. you have Bonus 1 = 10% and Bonus 2 = 15% but they are displayed as a single bonus in form of Bonus 3 =25% but Bonus 1 and Bonus 2 as counted as two bonuses, even though are displayed as one).

I reduced this problem as much as I could by prioritizing some bonuses to be displayed like this:
1 Empire Unhappiness
2 GA/GG
3 Policy Bonuses
4 Monopoly
...
n Combat Bonus (I put is the last one so that doesn't cause these issues, cause this bonus stacks with multiple bonuses)

**Regardless now it will not go more than 6 rows in the panel (it will not overflow).**

**Here some screenshots:**

**Here test UnitvsCity with two decimal points, see that overflows**
Civ5Screen0037

**Here test UnitvsCity with one decimal point, now is better and rounded**
Civ5Screen0043

**Here test UnitvsCity**
Civ5Screen0048

**Here CityvsUnit (didn't manage to give bonuses to the AI but should still work)**
Civ5Screen0042

**Here test UnitvsUnit**
Civ5Screen0047


ryanmusante:
Coolio

TechpriestEnginseer:
But you have to make the change for all EnemyUnitPanel.lua tho. CP, CBP, the 43-civ version (EUI), the 43-civ version (non-EUI), 6a, 6b....

Aqinu:
Yeah, all of those are all identical and I don't know if these has to be changed as well, wait for @LoneGazebo to confirm if this is the case then I will push the rest of them (just copy and paste no big deal...)

TechpriestEnginseer:
All UI files of the same file name have to have similar changes for it to be approved.

Aqinu:
So you are 100% sure that all of them must be pushed? OK then...

Aqinu:
**Now the bonus Strategic Resource Monopoly is correctly aligned (with Fix for issue #5488 Part 3), see screenshot 3 from above where is not properly aligned**

Civ5Screen0049

Aqinu:
Conflict caused, I guess by Terracotta Army Fix from previous @HeathcliffWarriors  pull request (just keep my change of it so that the wonders have the same format, in case you accept my improvements)...

Issue #5423, opened 2019-06-08
Human/AI Diplo Inequalities for Coop War Requests


HeathcliffWarriors:
_1. Mod version (i.e Date - 4/23):_ 6/2

DoSendStatementToPlayer function, when a coop war request is rejected:
```
#if defined(MOD_BALANCE_CORE)
				{
					if(eAgainstPlayer != NO_PLAYER && ePlayer != NO_PLAYER)
					{
						if(GetLoyalty() >= 5 && GetMajorCivOpinion(eAgainstPlayer) > MAJOR_CIV_OPINION_FAVORABLE)
						{
							GET_PLAYER(eAgainstPlayer).GetDiplomacyAI()->ChangeRecentAssistValue(ePlayer, -200);
							GET_PLAYER(eAgainstPlayer).GetDiplomacyAI()->ChangeNumTimesIntrigueSharedBy(ePlayer, 1);
							GET_PLAYER(eAgainstPlayer).GetDiplomacyAI()->SetApproachTowardsUsGuess(GetPlayer()->GetID(), MAJOR_CIV_APPROACH_WAR);
							GET_PLAYER(eAgainstPlayer).GetDiplomacyAI()->SetApproachTowardsUsGuessCounter(GetPlayer()->GetID(), 0);
						}
						else if(GetDiploBalance() > 5 && GetMajorCivOpinion(eAgainstPlayer) > MAJOR_CIV_OPINION_NEUTRAL)
						{
							GET_PLAYER(eAgainstPlayer).GetDiplomacyAI()->ChangeRecentAssistValue(ePlayer, -200);
							GET_PLAYER(eAgainstPlayer).GetDiplomacyAI()->ChangeNumTimesIntrigueSharedBy(ePlayer, 1);
							GET_PLAYER(eAgainstPlayer).GetDiplomacyAI()->SetApproachTowardsUsGuess(GetPlayer()->GetID(), MAJOR_CIV_APPROACH_WAR);
							GET_PLAYER(eAgainstPlayer).GetDiplomacyAI()->SetApproachTowardsUsGuessCounter(GetPlayer()->GetID(), 0);
						}
					}
					SetCoopWarAcceptedState(ePlayer, eAgainstPlayer, eAcceptedState);
				}
```
**Problem 1.** The AI does not change the # of times a coop war was denied for its approach calculation. To be equal with the human it'd need to be +1 for denying it and +2 for reporting it to the other AI.

[...]

PR #5493, opened 2019-07-13
Improve approach guesses & add coop war warning logic

HeathcliffWarriors:
Makes the following improvements to my reworked approach guesses:

- Code optimization, removed useless lines;

- AI will now reset its approach guess to WAR if the other player's military deployment is threatening enough (follows similar rules as deciding to ask for a military promise, including comparing the current turn and previous turn's postures);

- If the AI has no other guess and the human has a military aggression posture of at least MEDIUM, the AI will assume they are GUARDED;

- When deciding if an opponent is threatening war, AI will consider a war possible if there have been 0 previous wars, not just the peace counter

- To prevent exploits, the human making a demand will only set the guessed approach to HOSTILE if (a) the approach isn't already guessed to be war or (b) the AI gave the human what they demanded - it will still revert back to WAR the next turn if the human's military deployment is threatening enough.

Text changes:
- Updated "comment" opinion values for the thresholds; the old values were from base Civ V, not even BNW, plus the BNW values are listed above the code in CoreChanges.sql which modifies them. Having these values here will be more useful for modders/people trying to understand the code.

- Moved the victory dispute and victory block checks in GetBestApproachTowardsMajorCiv next to each other; not a code change, just looks better that way.

- Added comments showing the FAVORABLE and COMPETITOR opinion thresholds to GetBestApproachTowardsMajorCiv.

Other changes:
- CvDiplomacyAI: When a human breaks a coop war promise with an AI, now adds 300 recent assist value and +2 coop war denials, the same as if you warn the target AI. I think this is fair in addition to the broken promise penalty, as it's at least as disrespectful as the breach of diplomatic secrecy caused by warning the other AI *and* constitutes denying the coop war request in an especially nasty way.

- CoreChanges.sql: Removed duplicate entries modifying OPINION_THRESHOLD_FAVORABLE and OPINION_THRESHOLD_COMPETITOR if the policy tweaks are enabled; no other thresholds are modified by the policy tweaks, all of them are modified by the diplomacy/war tweaks later in the same file, and this looks like a simple oversight since I can't see any reason why policy changes would affect this (if I'm wrong, please revert).


HeathcliffWarriors:
When I raised issue #5423, it was pointed out that the reason the AI had diplo inequalities when it came to the penalties for denying a coop war request was because the AI's choice to warn the other AI was largely RNG.

This commit fixes #5423 by equalizing the approach/opinion penalties between humans and the AI, and adding the function IsCoopWarRequestUnacceptable to determine whether or not to warn the target AI, with smarter logic and much less RNG; feel free to build on this decision-making process.

Also fixes what seems to be a bug in the DoSendStatementToPlayer function where the AI would never warn the target AI when asked by an AI to start a coop war, and cleaned up the confusing syntax which set the CoopWarState.

PR #5525, opened 2019-07-30
Fixed some typos

Aqinu:
For the promotions fixes I used this format:
1) If range unit/siege unit has only RCS, without other requirements, I added that is only when attacking (because when a range unit is attacked by another range unit, the strength is calculated using the RCS of the defending unit, not the CS, and in this case the bonus doesn't apply for the defending unit)
2) If range unit/siege unit has bonus for defense and offense I used the "[ICON_STRENGTH] CS"
3) I specify in multiple cases when the bonus was only for attacking or for defense and offense (in the latter case using "[ICON_STRENGTH] CS", see **targeting**, **infiltrators** promotions where the first ability is for offense and defense and the second only offense)


Aqinu:
How did I know that some promotions are offense only and other for offense and defense, simple I've looked at the Extended information for the promotions and if specifies 'attack' somewhere in the promotion name is clear that is for attack only and the ones that didn't have attack in their name is clear for offense and defense.

And, of course, I've made some testing to be sure that everything is correct using IGE.

Aqinu:
**Added a new paragraph for nearby resources in buildings description!**

Aqinu:
@LoneGazebo I've noticed something weird with the pull request that I did the last time, which involved fixes for World Wonders from Word Congress, I can see that the changes that I made are present in the current version (7/29) but in the LoneGazebo/Community-Patch-DLL repository the changes aren't there... (I said this because I've made another change to the Grand Canal Wonder so that you know what to do with it, from what I can tell the changes are in another branch)

PR #5534, opened 2019-08-02
Fixed Some Typos

Aqinu:


Aqinu:
**One change** I've made to the Indonesia UU Promotions, that is to the "Sneak Attack" one because, after I did some testing the bonus is increased by 10% not 50% the way it was described.

**Here promotion description**
Civ5Screen0007


**Here the unit with the promotion**
Civ5Screen0008

**Here combat status for flanking is 25% (10 (adjacent unit) + 5 (1 adjacent unit from shock) + 10 (UU bonus but 10 not 50))**
Civ5Screen0009

**Here another combat status with 50% (10 (one adjacent) + 10 (second adjacent) + 10 (2 x adjacent units from shock, 5+5) + 20 (2 x adjacent units from UU, 10+10, not 50+50))**
Civ5Screen0010

**Here normal unit without promotions confirms correct calculations from above, here 30% (20 (2 x nr. adjacent units) + 10 (2 x nr. adjacent units from shock, 5 + 5))**
Civ5Screen0011

PR #5539, opened 2019-08-05
Fixed Some Typo

Aqinu:


Aqinu:
I've changed these three promotions because were incorrect described...

Civ5Screen0003
Civ5Screen0004
Civ5Screen0005

Aqinu:
I've changed these promotions because were incorrect described (the flanking)...
Civ5Screen0005
Civ5Screen0007
Civ5Screen0008
Civ5Screen0009

Aqinu:
The way I arranged the abilities within a promotion is like this:
[ICON_STRENGTH] --> [ICON_RANGE_STRENGTH] --> [ICON_STRENGTH_REQUIREMENT][ICON_RANGE_STRENGTH_REQUIREMENT] --> [OTHER_ABILITIES]

Community Balance Patch/Balance Changes/Text/en_US/CivText.sql
```diff
@@ -738,7 +738,7 @@ WHERE Tag = 'TXT_KEY_UNIT_MAYAN_ATLATLIST_STRATEGY' AND EXISTS (SELECT * FROM CO
 -- Mongols
 --------------------
 UPDATE Language_en_US
-SET Text = 'Mounted Ranged Units gain +2 [ICON_MOVES] Movement and ignore Zone of Control. City-States surrender via Heavy Tribute. Receive 250% of Heavy Tribute yields in [ICON_CAPITAL] Capital when a City-State is acquired, scaling with Era.'
+SET Text = 'Mounted Ranged Units gain +2 [ICON_MOVES] Movement and ignores [COLOR_POSITIVE_TEXT]Zone of Control[ENDCOLOR]. City-States surrender via Heavy Tribute. Receive 250% of Heavy Tribute yields in [ICON_CAPITAL] Capital when a City-State is acquired, scaling with Era.'
```
HeathcliffWarriors: Should stay in singular form (ignore, not ignores) to be grammatically correct.

Aqinu: Done.

Aqinu:
Just wanted to say that I wished to add more default promotions (e.g. those from the submarine) but I noticed that if I change the description of some of these, the changes occurs everywhere (in the title and in the left panel with the promotion list), so I let those alone.

One more thing, @LoneGazebo, I suggest that when another release of the VP arise to specify something like this:

"In the case for the all Ranged Units (Land, Naval, Air), if the ability within a promotion states a bonus with **melee shield** and it doesn't specify if it is for **attack/defense**, that means that is for **both**. Everything else beside this case should be quite clear."

PR #5558, opened 2019-08-17
Some Typo Fixes

Aqinu:


Aqinu:
**Previously Requested Changes**

Civ5Screen0016

**General Typo Fixes - Scholars In Residence**

Civ5Screen0002

Aqinu:
**Fix for Combat Preview Panel Misc. Bonuses**

I've noticed that when the unit has 4 bonuses the misc. bonus appears, which it shouldn't (one bonus didn't have "**iModifier ~= 0**" condition. Also "**Ignores Great General penalty**" bonus should reduce the misc. bonus, I fix that as well (you can see the use of "**-iModifier**").

**Before the fix**
Civ5Screen0001


**After the fix**
Civ5Screen0002

Issue #5600, opened 2019-09-07
8/31 Reproducible CTD on Inca turn -> possible interplay between CPDLL and 3UC/4UC


JanSobus:
_1. Mod version (i.e Date - 4/23):_ 8-31-1

_2. Mod list (if using Vox Populi only, leave blank):_

3/4UC + Wonders Expanded + Enhanced Naval Warfare+ Unique City States + collection of UI mods , all combined into Modpack

_3. Error description:_ 
Reproducible CTD happening on Inca turn (save 1), started another game, Inca in game again, another reproducible CTD on their turn (save2).
I think the problem has been identified in the 3UC/4UC thread with the culprit code being somewhere in CPDLL , see discussion by ridjack and pineappledan :  https://forums.civfanatics.com/threads/3rd-and-4th-unique-components-for-vp-official-thread.631657/page-93 

_4. Steps to reproduce (optional):_
Load savegame, end turn

---------------------------
Supporting information:
Please note that you can attach .zip files by dragging-and-dropping them. If possible, zip up all supporting data and post that way.

1. Log files (always attach your Logs folder, located at My Documents/My Games/Sid Meier's Civilization 5. Make sure you have enabled logging before experiencing an error! Go here to find out how: http://forums.civfanatics.com/showthread.php?t=487482):
[Logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/3586142/Logs.zip)
2. Save game (always attach a save that was made a turn before the error; located at My Documents/My Games/Sid Meier's Civilization 5/ModdedSaves):
[Boudicca1_0042 BC-2950.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/3586140/Boudicca1_0042.BC-2950.zip)
[Boudicca2_0061 BC-2475.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/3586141/Boudicca2_0061.BC-2475.zip)
3. CvMiniDump.dmp file (attach if experiencing a game crash. Located at Program Files/Steam/steamapps/common/Sid Meier's Civilization V):

4. Screenshots (optional):

PR #5627, opened 2019-09-16
Bugfixes 16/09

HungryForFood:
## Misc

- Added missing null check in CvPlot::setImprovementType.


HungryForFood:
Potential fix for #5600.

PR #5684, opened 2019-10-07
Some Typo Fixes

Aqinu:


Aqinu:
**Withdraw Chance Bonus Fix**

**Before Fix (this bonus should count as modifier just as info and in this instance it shouldn't appear even)**
Civ5Screen0006

**After Fix (the bonus it shown only when attacking a unit who has this promotion)**
Civ5Screen0005

**Another Example Before**
Civ5Screen0001


**And After**
Civ5Screen0007

Aqinu:
**Small Typo Fix**

I wanted to make it clear that "Creating Great Works or gaining Cities" grants food and gold in all Cities each time these thing occur, so it's not read like a different bonus who isn't related to the first bonus.

Aqinu:
**Logistic Combat Penalty Fix and Withdraw Chance Fix Part 2**

For some reason this bonus wasn't shown in the enemy panel, now it's fixed. 

**Before**
r1


**After**
r2

PR #5694, opened 2019-10-09
Some Typo Fixes

Aqinu:


HeathcliffWarriors:
Perhaps worth noting in the Settler description that you cannot settle a city next to someone's border. I didn't know that - it explains a lot.

PR #5751, opened 2019-10-28
AI should attack barbarian cities again

LoneGazebo:


HeathcliffWarriors:
Conflicts with my edit to the function, here's the full code with ilteroi's change if you want to copy + paste:

```
// Does the AI even want to conquer another player if they are at war?
// Since there is no "defensive war" flag, this seems to be the best way to differentiate
bool CvDiplomacyAI::IsWantsToConquer(PlayerTypes ePlayer) const
{
	if(GC.getGame().isOption(GAMEOPTION_ALWAYS_WAR) || GC.getGame().isOption(GAMEOPTION_NO_CHANGING_WAR_PEACE))
		return true;
	
	//we don't have a target value for the barbarians ... could change it but that would break savegames
	if (ePlayer == BARBARIAN_PLAYER)
		return true;
	
	// Locked into war by a coop war/trade agreement
	if (GetNumTurnsLockedIntoWar(ePlayer) > 0)
	{
		// Fiercely competitive with this guy? Who cares if he's tough to kill?
		if (GetVictoryDisputeLevel(ePlayer) == DISPUTE_LEVEL_FIERCE || GetVictoryBlockLevel(ePlayer) == BLOCK_LEVEL_FIERCE || GetBiggestCompetitor() == ePlayer)
		{
			return true;
		}
		// Otherwise, AI is bolder if backed up by someone else
		else if (GetPlayerTargetValue(ePlayer) >= TARGET_VALUE_AVERAGE)
		{
			return true;
		}
	}
	
	if (GetBoldness() > 6)
	{
		//bold players
		if (GetPlayerTargetValue(ePlayer) < TARGET_VALUE_AVERAGE)
			return false;
	}
	else
	{
		//less bold players
		if (GetPlayerTargetValue(ePlayer) < TARGET_VALUE_FAVORABLE)
			return false;
	}

	return true; //by default we conquer everything
}
```

PR #5764, opened 2019-10-31
Add Diplo Aggression Options, Calculate GPT Loss from DoW & Misc Improvements

HeathcliffWarriors:
**Adds Advanced Diplo AI Aggression Options (new memory values)**, because of the many people who have complained about the AI being too competitive.

So, if I did this correctly and didn't screw up, they should now be able to disable AI war declarations VS humans, disable AI war declarations globally, or remove victory competition penalties.

These are all disabled by default, of course.

Options are in CoreChanges.sql.

DIPLO_AI_NO_VICTORY_COMPETITION: Set to 1 to disable the AI's victory competition.

DIPLO_AI_WAR_DISALLOWED_HUMAN: Set to 1 to disable AI civs initiating wars with human players.

DIPLO_AI_WAR_DISALLOWED_GLOBAL: Set to 1 to disable AI civs initiating wars with other civs or City-States.

Default for all of them is 0.

Enabling either of the War Disallowed settings will cause the AI not to pursue Domination Victories, and prevent the AI requesting/offering third party wars to humans (although humans can still ask them).

(I copied what ilteroi did with his combat knobs due to my unfamiliarity with adding globals, feel free to change the placement/data type for the options)

-------------------------------------------

**Calculate GPT Loss From DoW**
There doesn't seem to be a function which accurately calculates this (unless you count the woefully inadequate vanilla check in GetCoopWarScore), so I made one.

Will be useful in the event that a sanity check is added for the AI to avoid entering wars that'll harm their economy too much.

bool bOtherPlayerEstimate allows for ePlayer to call the function to see how much GPT the other player would lose in the event of war, but doesn't use information (TR/deal details) that ePlayer wouldn't have access to.

bool bIgnoreDPs is used if you want to calculate how much GPT would be lost if ePlayer declared war, rather than this player - since that wouldn't trigger DPs.

-------------------------------------------
Improves diplo mechanics for stolen artifacts **(adds new memory value)**:

- AI will no longer give Open Borders to players who have stolen artifacts from them, if there are still artifacts in their lands that can be stolen by the other player.

- AI will now always have an opinion penalty for stealing artifacts, regardless of if you make a promise or not. The penalty will decay over time, and making a promise will reduce the penalty, provided you keep it.

- (Hopefully) fixed bug where AI would never obey its promise to stop digging up artifacts.
-------------------------------------------
Other changes:
- AI should be **much less likely** to stupidly backstab their friends and DPs, or to start wars that cause their teammates to backstab _their_ friends and DPs.

- AI should likewise be much more hesitant to attack powerful neighbors via a DP over insults/etc.

- Increased AI backstabbing penalties personally and factored it in for breaking military/CS promises, but removed the way too harsh IsUntrustworthyFriend logic that gave large permanent friendship weight reductions for ending a DoF with any civ.

- Increased AI hostility for killing or repeatedly attacking their protected City-States.

- Reworked IsUntrustworthyFriend function a bit: changed into a function that evaluates the player's trustworthiness from the evaluating AI's perspective, rather than a rigid global true/false value. Will work on this more soon.

- Spies will no longer consider repeatable projects intrigue-worthy. Rationale: If it's repeatable, then it's not noteworthy.

- IsPlayerRecklessExpander will return false if the player doesn't have more cities than the AI; will now count City-States with 2+ cities towards the global average (to appropriately handle any weird situations or modmods); fixed issue where it would always return false when AI checked themselves; invalid players will now always return false.

- Made AI less susceptible to being tricked into starting coop wars or giving/taking Open Borders in situations where it's likely a trap.

- AI teammates will now be willing to start coop wars immediately if they're bold enough and the target value is at least FAVORABLE, just like normal AIs.

- AI teammates of humans will no longer agree to coop or 3rd party war bribes on their own.

- Made war weariness no longer apply to wars where the other player has 0 cities, as they're insignificant. Also set war weariness to 0 for the dead player when a player is killed.

- Tweaked approach guess sensitivity for WAR approach (will only consider military posture if it's extremely threatening).


supermerill:
Imo, it's an information that could also be displayed to the human player. It's nearly impossible to calculate manually and if the AI has access to this information, then why not the human?

HeathcliffWarriors:
@supermerill There is a popup that appears when you try to declare war which displays deals, trade routes, masters/vassals and DPs; but it doesn't calculate the total GPT lost from all of these.

I'm no UI wizard :)...but once this function is complete - I suspect there are other things still to consider, like religion, for now it's just an estimate - I'll open an issue to see if anyone's able to display it to the human player.

There's other checks like this that could be made, for Happiness and yields besides Gold.

PR #5791, opened 2019-11-03
Some Typo Fixes

Aqinu:


Aqinu:
**Fix Combat Panel Overflow**

This bonus applies to both range and melee mounted units, but for range has different description which caused overflow in this situation.

a

**Here for melee mounted same bonus applied**
b

**Here after fix**
c

Aqinu:
**Combat Preview Fix**

Misc. shows 0%. Enemy interception is in theirPanel so it is not needed to be counted for myPanel.

**Before Fix**
a

**After Fix**
b

LoneGazebo:
@Aqinu make another push

PR #5816, opened 2019-11-10
Some Typo Fixes

Aqinu:
I made duplicates here, hope is not a problem...


Aqinu:
**Small Combat Panel Fix**
Another instance with overflow...

**Before**
a

**After**
b

**After (even though occupies in this instance just 1 row unlike 2 as above, i can't otherwise to fix this...)**
c

Issue #5818, opened 2019-11-10
Terracota Army still providing old bonus


Guynemer:
_1. Mod version (i.e Date - 4/23):_Nov 8

_2. Mod list (if using Vox Populi only, leave blank):_3/4 UC, Dan's various civs for 3/4 UC, Dan's building changes and China/Spain rebalance

_3. Error description:_ TCA is still providing the extra copies of units, in addition to culture when winning a battle 

---

ASCII-Guy:
I have put a pull request #5819 to fix this issue.

PR #5819, opened 2019-11-10
Remove Military Unit Duplication From Terracotta Army

ASCII-Guy:
It has been reported that the wonder is still duplicating military units.


ASCII-Guy:
This should fix #5818.

PR #5886, opened 2019-11-24
Diplo Fixes & Improvements

HeathcliffWarriors:
**Note:** The enormous amount of changes isn't notable, I just did Find "if(" and "switch(" and Replaced them with "if (" and "switch ("

Bugfixes

Cooldown on demands will now only trigger if the AI has accepted a demand (this removes the unfun mechanic of guessing how long it'll be before the AI will accept a demand and having to wait another 20-ish turns if you're wrong; this should also make demands more fun and useful, since you don't have to wait before asking again if they rebuffed your demand)

Nerfed AI friendship willingness slightly

Nerfed opinion bonus for having no contested borders, maximum is now +15 (was +20)

AI should be more sane in targeting City-States

AI will no longer consider a player an easy target if their war against them is in a defensive state

AI will no longer attack its friends' protected/allied City-States if its approach towards the friend (real or fake) is FRIENDLY

Nerfed opinion bonus for vassals of the same master to +30 (was +50)

AI will now update its opinion and approach towards all players immediately when any of the following events happens:
- It makes a Declaration of Friendship with another player
- It ends a Declaration of Friendship with another player
- It denounces another player
- It declares war on someone
- ...or has any of these things happen to it

Simplified & reworked bonus/penalty to warmongering score based on global politics

Slight nerf to penalty for refusing to start a coop war

Removed double opinion penalty for stealing territory from warmongers (approach penalty should be enough)

Small improvement to AI approach estimates

Resurrecting a player should now also give you the temporary bonus for liberating a city, in addition to other diplo bonuses with them

The following notifications should no longer be sent:
- An Unmet Player has denounced CivName!
- CivName has denounced An Unmet Player!

...because the idea that you would know of a denouncement but not who's involved is ridiculous.


ilteroi:
again, this breaks the build. can you please make sure your stuff at least compiles before creating a PR? also, i'm not sure whether 3500 whitespace changes really are a step forward ...

PR #5913, opened 2019-11-30
Promotion: Fortification Yield

ASCII-Guy:
Another request from pineappledan. "A promotion table that lets you define a yield that the player gets every turn if that unit is garrisoned in a city, Fort, or citadel." Considering there's already a promotion table for garrisons, I simply added one for fortifications.

I'm not sure if I did the correct logic for getting the combat unit on the plot. Specifically line 9815 of "CvPlot.cpp".


LoneGazebo:
I'll fix it.

PR #5914, opened 2019-11-30
Promotion: Free Embark

ASCII-Guy:
Another request from pineappledan. "The ability to give units free movement points on embarkation, or to not lose movement points if they embark. Whichever is easier."

In this case I made no movement point loss on embarking. I put a counter in like the double movements. Hopefully that's what prevents an infinite movement loop.


LoneGazebo:
Let's take it easy on additional API requests - we don't have a lot of wiggle room.

ASCII-Guy:
Understandable. Holding out for that 64-bit update.

ryanmusante:
@ASCII-Guy you're funny.

PR #5919, opened 2019-12-01
Diplo Fixes & Improvements

HeathcliffWarriors:
**Changelog**
- Fixed some bugs
- Adds several new memory values to separate trade route plundering, plotting, and City-State coup penalties from each other
- Adds several new memory values to allow for some negative modifiers (the three above + demands) to disappear over time properly
- Added new memory values for better promise tracking (conversion/artifacts)
- Simplified some code thanks to the new memory values
- Added some opinion weight defines, including two that Firaxis apparently forgot to add in
- Plotting penalty now attainable by humans, replaces recent assist penalty for being caught in coop war planning/third party war bribery (although getting caught plotting will still remove any recent assist bonus)
- Bumped up opinion weight for global military promise broken penalty to 20 (was 15)
- Re-added global City-State promise broken penalty (also 20)
- Reduced opinion penalties for trade route plundering & City-State coups
- Plundering a trade route will now count against killing/capturing civilians, including if at war (temporary penalty, decreasing with era)
- Added opinion bonus of -80 for liberating a player's capital; unlike regular liberation, this one will not be removed until you declare war or capture a city; does not stack with regular liberation bonus
- Share opinion requests will now provide DECEPTIVE/WAR approach information if Opinion is Friend+ and approach is FRIENDLY (this is close to Putmalk's original intention; because VP raised the Opinion threshold for Ally, you would almost never get accurate information as intended); makes them more useful, too


HeathcliffWarriors:
@ASCII-Guy Thanks for that :)

PR #5938, opened 2019-12-06
Proposed Barbarian Spawn Changes

HeathcliffWarriors:
The current situation is as follows:
- Barbarians in uprisings are always land melee units, always the same type of unit, and never Unique or Ranged units.
- Barbarians spawn 2 units when a camp is created; 1 unit if Chill Barbarians is enabled.
- Chill Barbarians increases Barbarian spawn rates from camps & cities by 50%, and has no other effects.
- Raging Barbarians halves Barbarian spawn rates from camps & cities, and has no other effects.
- Barbarians do not spawn a garrison when they take over a city (usually as a result of a Horde/Rebellion quest), which can make them quite vulnerable.
- Barbarians can still spawn even if "No Barbarians" is enabled, from city uprisings (Unhappiness). Chill/Raging Barbarians do not have any effect on city uprisings.

I propose to make the following changes, and have included a pull request for the code work:
- Unit spawns from camps: 1 (Chill), 2 (Normal), 3 (Raging)
- Unit spawns when a city is captured: 1 (Chill), 2 (Normal), 3 (Raging)
- Horde/Rebellion quests: -1 Barb per wave (Chill), +1 Barb per wave (Raging)
- Espionage-induced rebellion: do not occur (No Barbs), -1 Barb (Chill), +1 Barb (Raging)
- Uprisings from Unhappiness: do not occur (No Barbs), are delayed by 50%, and -1 Barb per uprising (Chill), occur twice as often, and +1 Barb per uprising (Raging)
- Barbarian units in uprisings: Each unit in an uprising is of a random type + can be the player's Unique Units, or ranged units.

Rationale for Changes
- Unit spawns from camps: 1/2/3 is a linear progression and allows the player to dynamically adjust how many Barbarians they want to deal with.
- Unit spawns from cities: Same as camps + makes the strategy of waiting for a City-State to be captured, then rushing in to liberate it a little more challenging (currently it's often very easy, particularly with the CS nerf for cities).
- Horde/Rebellion quests: I feel that Chill/Raging Barbarians should impact these quests somewhat, since they deal with Barbarians.
- Espionage-induced rebellion: Same, and if the player has selected No Barbarians, then it makes sense that Barbarians should not be spawned in from espionage.
- Uprisings from Unhappiness: Chill/No Barbarians would now make it easier for new players who are getting a handle on the happiness system, allowing them to manage rebellions more easily. I've seen some complaints from new players about happiness management, so this would give them another tool. Also, since it deals with Barbarians, I feel the spawn rate should be affected by this, and once again, if the player has selected No Barbarians, then it makes sense that Barbarians should not appear. For Raging Barbarians, this would act as a counterbalance for the strategy of "go Authority and turn on Raging Barbarians for massive yield bonuses" - you can still do that, but if you go on a conquering spree and don't make your citizens happy, prepare to deal with more rebels.
- Barbarian units in uprisings: I feel this would be fun and less boring, but it makes karmic sense that warmongers should occasionally have to deal with a rebelling Jaguar or Impi, if they're not managing their empire's happiness well. :)

Also fixes a typo (super & very unhappy text keys were swapped).

### DIFF
--- a/43 Civs CP/43 Civs CP/EUI/TopPanel.lua
+++ b/43 Civs CP/43 Civs CP/EUI/TopPanel.lua
@@ -1155,10 +1155,14 @@ if civ5_mode then
 
 		-- CBP EDITS HERE
 
-			if g_activePlayer:IsEmpireSuperUnhappy() then
+			if g_activePlayer:IsEmpireSuperUnhappy() and not Game.IsOption(GameOptionTypes.GAMEOPTION_NO_BARBARIANS) then
 					tips:insert( "[COLOR:255:60:60:255]" .. L("TXT_KEY_TP_EMPIRE_SUPER_UNHAPPY")  .. "[ENDCOLOR]" )
-			elseif g_activePlayer:IsEmpireVeryUnhappy() then	
-				tips:insert("[COLOR:255:60:60:255]" .. L("TXT_KEY_TP_EMPIRE_VERY_UNHAPPY") .. "[ENDCOLOR]" )
+			else if g_activePlayer:IsEmpireSuperUnhappy() and Game.IsOption(GameOptionTypes.GAMEOPTION_NO_BARBARIANS) then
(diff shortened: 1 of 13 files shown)


HeathcliffWarriors:
Closed pending build error fixes.

PR #5944, opened 2019-12-08
Diplo Fixes & Improvements

HeathcliffWarriors:
**! Adds new memory value (DoFBrokenTurn) !**

**Changelog**
- Removed cooldown on DoF requests.

- Typo fixes.

- Bugfixes and optimizations.

- Improvements to deal AI valuation for Defensive Pacts, third party war, third party peace.

- AI can now accept bribes normally again, not just against major competitors, since there's enough other checks to prevent issues.

- Fixed bugs and made improvements to DP logic; should now make DPs more often, ties are broken in favor of the human (also applies to DoF/competitor weight), and AI will prioritize obtaining DPs with players that have the highest DP value.

- However, AIs with hostile Approaches may backstab players they only have a DP with.

- Improvements and bugfixes for DoRelationshipPairing (experimental).

- AI will now end Declarations of Friendship if the DoF value for a player reaches -32 or below (32 is required to make a DoF).

- AI diplo malus for breaking a DoF now timed, lasts 1.5x deal duration (75 turns on standard speed).

- AI should be more aggressive in certain circumstances.

- Improvements to IsEasyTarget, now strategically targets unhappy players (if it would be fair for the AI to do so, i.e. they have a way of knowing that information that humans can also use)

- AI now less likely to peace out when winning, target is close to victory, or target captured their capital/Holy City (experimental, may need adjustment).

- Nerfed penalty for plundering trade routes when not at war to 5 per TR (was 10; still 2x when Morocco is caught plundering by the TR owner, however)


HeathcliffWarriors:
Closed pending build error fixes.

PR #5962, opened 2019-12-14
Fix API Additions

ASCII-Guy:
UnableToCancelRazing and AnyBodyOfWaterMakesValid are working. FortificationYield has been fixed with this.
I have removed FreeEmbark and have overhauled it. Now split into EmbarkFlatCost and DisembarkFlatCost. A unit with the EmbarkFlatCost promotions embarks for one movement point. A unit with DisembarkFlatCost disembarks for one movement point.


ASCII-Guy:
I was under the impression that AnyBodyOfWaterMakesValid wasn't working, but it turns out it was in error in the mod's XML.

PR #5977, opened 2019-12-16
Diplo Fixes & Improvements

HeathcliffWarriors:
**NOTE**
Adds new memory values (values to keep track of # times war was declared and # times city captured, and an "opinion towards us guess" value - will be used shortly).

Includes some miscellaneous diplomacy-related fixes & improvements.


HeathcliffWarriors:
@LoneGazebo Welcome back :)

FYI, am now using the IDE to make my pull requests, and testing for build errors before pushing, so no more need to worry about build errors from my commits. Sorry for the trouble I put you through beforehand :)

PR #5982, opened 2019-12-18
Small fixes & add more religion/ideology weight

HeathcliffWarriors:
Just a few small miscellaneous fixes.


Iamblichos:
@HeathcliffWarriors you might want to check your code in CvDiplomacyAI. I think you have a variable that isn't defined right when it saved and then reloaded from kstream. Game crashes when trying to reload an autosave or any save. All was working well with 12-1 release.

Iamblichos:
CC: @LoneGazebo

HeathcliffWarriors:
@Iamblichos My first excuse to run the debugger, how excellent. :)

PR #6031, opened 2019-12-28
Jose Rizal 28/12/19

HungryForFood:
## Trait_TradeRouteProductionSiphon
- Adds production to your trade route origin cities, based on the production in the destination cities.
- AI configured to value open borders more with a civilization with this trait,

## Trait_YieldFromRouteMovementInForeignTerritory
- Yields from trade unit movement when they are in foreign territory.
- Can be configured so that your trade partner gains the yields when they send a trade route to you.
- AI will attempt to consider the yields in this trait when choosing trade routes.

Not save game compatible.


HungryForFood:
That is correct. Actually I forgot to consider barbarians in the logic, maybe I need to work in that.

On 28 Dec 2019 5:27 pm, HeathcliffWarriors <notifications@github.com> wrote:

@HeathcliffWarriors commented on this pull request.

@HungryForFood<https://github.com/HungryForFood> For IsInForeignOwnedTerritory, keep in mind that Barbarians can own a tile if they capture a City-State.

HungryForFood:
Yep, should be good now.

PR #6047, opened 2020-01-03
BattleRoyale 04/01/20

HungryForFood:
- Enable for compiling main VP build.
- Now to use CustomModOptions.xml to enable instead.


HungryForFood:
To enable, use this line in SQL:

UPDATE CustomModOptions SET Value = 1 WHERE Name = 'BATTLE_ROYALE';

PR #6290, opened 2020-02-25
Small Fix

Aqinu:


Clutterbutter:
I'd say "can only" rather than "must."'

Issue #6294, opened 2020-02-26
Trouble debugging build errors


HeathcliffWarriors:
I've been working on the revisions to the diplo AI's approach selection process I described in #6255. My current progress has been uploaded to the diplomacy branch of the repository.

Edited functions:
- DoUpdateMajorCivApproaches
- GetBestApproachTowardsMajorCiv
- Changed the call to GetBestApproachTowardsMajorCiv when meeting a new major to a DoUpdateMajorCivApproaches call.

Added functions:
- IsTeammate
- IsStrategicTradePartner
- IsTeamUntrustworthy

However, while checking the code for build errors, I encountered a type of build error I've never seen before, and I don't understand how to debug it.

The first pass of the compiler works just fine, but on the second pass:
![Screenshot](https://user-images.githubusercontent.com/38794241/75328570-38af6900-584c-11ea-95a7-a81cbaec7d48.png)
```
1>Finished pass 1
1>
1>CvLuaPlayer.obj : error LNK2001: unresolved external symbol "public: enum MajorCivApproachTypes __thiscall CvDiplomacyAI::GetMajorCivApproach(enum PlayerTypes,bool)const " (?GetMajorCivApproach@CvDiplomacyAI@@QBE?AW4MajorCivApproachTypes@@W4PlayerTypes@@_N@Z)
1>CvReligionClasses.obj : error LNK2001: unresolved external symbol "public: enum MajorCivApproachTypes __thiscall CvDiplomacyAI::GetMajorCivApproach(enum PlayerTypes,bool)const " (?GetMajorCivApproach@CvDiplomacyAI@@QBE?AW4MajorCivApproachTypes@@W4PlayerTypes@@_N@Z)
1>CvTradeClasses.obj : error LNK2001: unresolved external symbol "public: enum MajorCivApproachTypes __thiscall CvDiplomacyAI::GetMajorCivApproach(enum PlayerTypes,bool)const " (?GetMajorCivApproach@CvDiplomacyAI@@QBE?AW4MajorCivApproachTypes@@W4PlayerTypes@@_N@Z)
1>CvUnitProductionAI.obj : error LNK2001: unresolved external symbol "public: enum MajorCivApproachTypes __thiscall CvDiplomacyAI::GetMajorCivApproach(enum PlayerTypes,bool)const " (?GetMajorCivApproach@CvDiplomacyAI@@QBE?AW4MajorCivApproachTypes@@W4PlayerTypes@@_N@Z)
[...]

PR #6296, opened 2020-02-26
Rewrite approach selection functions

LoneGazebo:
Code doesn't currently build, still needs work.


HeathcliffWarriors:
See #6294, @LoneGazebo.

LoneGazebo:
Yeah I wanted to see why it was balking.

HeathcliffWarriors:
Ah, that makes sense. Thanks for any help, I'm confused here. :)

PR #6353, opened 2020-03-16
Fix idiotic typo

HeathcliffWarriors:
I can't believe I missed that. *facepalms*

Strange that the compiler didn't catch it.

Sorry for all the pull requests today.


LoneGazebo:
compiler won't catch division by zero errors if the variable isn't a known value.

PR #6493, opened 2020-04-17
Heritage Ideology choice update / Golden Age Points to Prestige hook

iskaandismet:


LoneGazebo:
Do you still need this? If so, resync and merge.

PR #6586, opened 2020-05-06
Diplo Fixes & Improvements

RecursiveStar:
@ilteroi To avoid merge errors with your changes since we've edited the same files, I rebased my pull request on the ole branch and am sending the request here instead.


RecursiveStar:
@ilteroi hmm, seems that Github doesn't actually make the destination of a pull request very clear since it groups them all together for the repository - but in case it isn't clear, the request is _to_ the ole branch for you to merge, since I based it on the ole branch to start with :)

I tested for build errors already, of course. Most of this is not new, just transferring diplo-related code from other files over to the diplo AI and reducing the amount of redundant code.

RecursiveStar:
Edit: Fixed a memory issue; however, should still be savegame compatible as the value was not serialized.

PR #5592, opened 2019-09-05
Some Typo Fixes

Aqinu:

PR #6606, opened 2020-05-22
 Check First If It Works, Fix Wonder Popup Text Overflow 

Aqinu:
I don't know how to compile it, but theoretically should work, just test it first to be sure. 

It's a fix for #5592


Aqinu:
@LoneGazebo 

Check it if it works wihout IGE, cause IGE has a similar file so that your are sure that it doesn't overwrite the one in VP, I checked it by making a wonder have 1 turn left, then I disable IGE, reload the game and hit next turn and then it worked (but as a separate mod not integrated as this one...)

PR #6652, opened 2020-06-07
Fix vassal protect & failed protect mechanics

RecursiveStar:


RecursiveStar:
Whoops, screwed this one up. :)

PR #6818, opened 2020-07-19
Various Fixes & Improvements

RecursiveVision:
@LoneGazebo This pull request fixes all the merge errors in the current repository and includes all of my and @ilteroi's pushed changes so far - I've carefully merged all of that into a single commit. To prevent further merge errors I'd suggest force-merging this, redownloading your local repository, making sure the added WonderPopup.xml files are in your list of files that get updated (I think you mentioned you had such a list to Aqinu once) and then working from there...


LoneGazebo:
Excellent work as always!

PR #6871, opened 2020-08-06
Events fixes #6836

### LINKED ISSUE #6836: Events: Bugs, Improvements and Suggestions
I have been looking into the implementation of Events and have a few observations to report. I've been working with my very old version but from a quick look at the code from a recent version, I don't think anything significant has changed. I have made a few changes/corrections to my old version which I think the current version would benefit from.

Bugs:

1) Where a choice has a probability of success, for example Hurricane Approaching, the test is backward so that a 33% chance is actually a 67% chance (2 occurrences)

2) When an event or choice is made obsolete by the discovery of a tech, the code checks the Prereq tech instead, so the event/choice is never valid (4 occurrences)

3) The Plague City Event is missing an entry in the table CityEvent_ParentEvents, so although it triggers, nothing happens.

4) The knock-on City Events for the Ice Age global event (e.g. FROZEN_WATER) don't trigger due to issues with entries in the table CityEvent_EventLinks. Not sure what's intended here. My solution was to add a fourth choice (sort of 'grin and bear it') and make the knock-on Events depend on that, so that the first two choices (giving up gold or production in all cities) mitigate the problem - I can post details if you think that's a good solution.  

Improvements (probably bugs):

5) For global events, the human player will only experience the event if it triggers on his turn. All the AI's will still experience the event whoever triggers it. This is due to a check for isEndTurn, apparently to avoid some exploit, which causes the event to fail if it's not the player's turn. The check may also be related to preventing popups occurring at the wrong time. My solution to this is to only allow global events to trigger on the human's turn (this would need further thought for multiplayer). I thought I would also need to increase the chance of the event happening to compensate for only one player potentially triggering it, but in practice I see the events more frequently than before - if anything I might lower the chances.

6) When multiple events are possible, the current implementation can only trigger the event with the highest probability. This means that certain low probability events become even rarer as they can only trigger when they are the only possible event. I changed this so that any one of the possible events can trigger. It appears that the code was designed to work allow any event to trigger, as it puts all the possible events into a vector and checks each one - if it was only meant to check the most probable event the code could be much simpler. Because it uses the same random number for all events and checks the most likely event first, there is no chance of a lesser event happening. I don't know the full history here so perhaps I'm just seeing an evolution of the code, but it looks to me like a bug. Because the basic support was already there, my change just required the insertion of a plus (I also removed the now unnecessary sort of the table and added the random number to the log to help validate my solution). Of course this change made events more frequent overall, so I also increased the range of the random numbers from 1000 to 2000 for player events and 2500 for city events, which seems to give events at a pleasing (to me) rate. I have subsequently seen some events which I don't think I've seen before in thousands of hours play.

My changes (marked afw) to DoEvents in CvPlayer (similar changes to CvCity):

```
	if(veValidEvents.size() > 0)
	{
//afw		veValidEvents.SortItems();
		int iRandIndex = GC.getGame().getJonRandNum(2000, "Picking random event for player."); //afw
		if(GC.getLogging())
		{
			CvString strBaseString;
			CvString strOutBuf;
			CvString strFileName = "EventLogging.csv";
			CvString playerName = getCivilizationShortDescription();
			FILogFile* pLog = LOGFILEMGR.GetLog(strFileName, FILogFile::kDontTimeStamp);
			strBaseString.Format("%03d, ", GC.getGame().getElapsedGameTurns());
			strBaseString += playerName + ", ";
			strOutBuf.Format("Found %d Events for seeding. Random=%d", veValidEvents.size(), iRandIndex);
			strBaseString += strOutBuf;
			pLog->Msg(strBaseString);
		}

//afw		int iRandIndex = GC.getGame().getJonRandNum(1000, "Picking random event for player.");

		//which one is it?
		int iWeight = 0;
		for (int iLoop = 0; iLoop < veValidEvents.size(); iLoop++)
		{
			EventTypes eEvent = (EventTypes)veValidEvents.GetElement(iLoop);
			CvModEventInfo* pkEventInfo = GC.getEventInfo(eEvent);
			if (!pkEventInfo)
				continue;

//afw			iWeight = veValidEvents.GetWeight(iLoop);
			iWeight += veValidEvents.GetWeight(iLoop); //afw
			if (iRandIndex < iWeight)
			{
				eChosenEvent = eEvent;
				break;
			}
		}
```

Possible Bug/Change:

7) There is a facility in the events system for checking the event status of other players. When checking for False, none of the other players can have the event active for it to be valid for you, which is fine. When checking for True, however, this requires the event to be active for ALL the other players. Reading the documentation, I don't think that's intended. I certainly tried to use it in a way where I wanted it to allow the event if ANY of the other players had the trigger event. Probably can't change this now in case existing additional events (the standard ones wouldn't be effected) rely on the existing implementation, although I think it's actually more likely that they are assuming the 'any' case and don't actually work currently. There is a similar situation with checking all cities.

Suggestion:

8) Regarding the rate of events, I wonder if this should scale with game speed to that you could expect a similar number of total events irrespective of the speed. I wonder also if a define (or even a game option) could be added giving the player control over the rate - it would just need to act as a multiplier for the random number range - so, for example, the user could choose to play with events, but at a lower frequency. Another point is that if you play with extra events, overall occurrences will become more frequent, especially after my change above, so perhaps the random number range could also vary somewhat based on the number of events in the game.

psparky:
Sorry, I couldn't see how to avoid CvPlayer being part of the request, but it's not needed as G has already done the code fixes. You just need the changes to CoreEvents.sql and Player_en_US.xml


psparky:
> To prevent merge errors (bringing everything together has been a complicated merge and this pull will likely revert changes), it'd be best if you saved the files you changed somewhere, reforked the repository, readded your changes as a commit to your fork's master branch (by replacing the changed files and making sure you aren't reverting changes), and made a new pull request.

I'll redo it. Does 'reforking' mean deleting and recreating the fork or can I just update my fork to the latest version (how)?

RecursiveVision:
I'd just delete and recreate (make sure to copy your changes somewhere first!).

RecursiveVision:
Normally you can update by making a pull request from the updated branch to your branch, but the auto-merge is broken here. @psparky

LoneGazebo:
I've manually grabbed the files I needed @psparky , feel free to close.

psparky:
OK great!

PR #6893, opened 2020-08-13
Small updates

Mileslewis:
Tweaks to promotion logic including for barrage change and added values for amphibious.

Vanguard (known as COASTAL_TERROR in xml) says it has 100 damage vs cities in game, but xml has value of 125. Didn't change this as I don't know which is correct.

Base Tech cost formula updated to be multiplicative rather than additive.

There are some other issues with tech cost and particularly overflow which I will mention in a separate issue.


RecursiveVision:
@Mileslewis Looks like Gazebo changed the XML value but not the text. He's also gonna make more changes to promotions, according to his forum posts.

PR #6969, opened 2020-08-26
RSI Code Cleanup

RecursiveVision:
Reuploaded to fix potential merge conflicts.


RecursiveVision:
@ilteroi Since you're back, you may want to consider updating to this to prevent later merge errors, since I changed 70 files.

RecursiveVision:
@LoneGazebo @ilteroi 

Added some experimental fixes to improve war logic to this commit, especially coop war logic. Unfortunately, because I cut and paste the coop war functions Github is recording 47,000 line changes because everything was moved around - but it builds okay and nothing's missing.

Includes a few other improvements as well, like a fix for #6990.

Notable changes:
- Cleaned up and consolidated all coop war related code
- Rewrote coop war desire calculation
- Commented out "coop war time" statement; all players will now start a coop war immediately once the 10 turns are up, or whenever either player goes to war with the target, whichever comes first (fixes various issues where players were out of sync, etc.)...also updated text on the coop war request option to reflect this
- When war is declared, all preparing coop wars on BOTH sides are launched immediately
- Players can still cancel their coop war agreements by ending their friendship with the AI or making a DoF/DP with the target player, but this results in a penalty for breaking the promise
- Removed "coop war counter", replaced with "state change turn"
- Humans can't request that AIs coop war their friends, DPs, or other invalid players (the button is simply not available) - AIs, of course, can't request this of humans either
- DoMakeWarOnPlayer() now updates approach; WAR AIs will have their approach set to a non-WAR approach based on war face if they don't "want to attack"; players that "want to attack" but don't have the WAR approach will have their approach set to WAR
- DoMakeWarOnPlayer() now always sets "want to attack" to true if a coop war is being prepared
- Added IsPhonyWar() to determine when an AI is in a "phony war" and should just make peace right away.
- Added IsAggressor()/SetAggressor(), a Diplomacy AI flag which is set to true if the AI either chose to declare war or wanted to declare war but was declared war on - this affects IsWantsToConquer() and in turn, IsPhonyWar().
- Added IsEmpireInBadShapeForWar() to CvPlayer - determines if AI should not start new wars.
- Added GetClosestCityToUsByPlots() to CvPlayer - used to determine a player's closest city to the AI's empire - the coop war desire calculation now evaluates the economic value of this closest city and scales coop war willingness based on that. May have other uses!
- Added HasAnyOffensiveOperationsAgainstPlayer() to CvPlayer, which tests if the AI is currently attacking a player (used in IsPhonyWar()).

I'm leaving for the weekend in one hour, so I can't test this myself - if someone could test this and make sure it doesn't cause crashes, disable key functions, cause other problems, etc. that would be appreciated!

RecursiveVision:
Oh, forgot to mention this also fixes the warmongering penalty for coop wars, which will be very welcome.

RecursiveVision:
I almost feel sad closing this pull request because the number is too perfect. 🤣 

Will recreate to ensure hotfix compatibility.

PR #7076, opened 2020-09-16
Various Cleanups & Bugfixes

RecursiveVision:


RecursiveVision:
@LoneGazebo I removed the following defines as they are useless (if someone tried to disable them it would likely break the DLL):
MOD_BALANCE_CORE_DEALS
MOD_BALANCE_CORE_DEALS_ADVANCED
MOD_BALANCE_CORE_DIPLOMACY
MOD_BALANCE_CORE_DIPLOMACY_ADVANCED

Where needed I replaced instances of
```
if (MOD_BALANCE_CORE_DIPLOMACY_ADVANCED)
```
with
```
if (MOD_BALANCE_CORE)
```

Also adjusted MOD_BALANCE_CORE_DIPLOMACY_ERA_INFLUENCE so it works properly and the description is accurate.

PR #7132, opened 2020-10-08
Fix "locked into war" timer not decreasing

RecursiveVision:


RecursiveVision:
@LoneGazebo You may want to merge this and push another hotfix...this bug means that you cannot make peace with someone once you start a coop war against them or get bribed into war...ever. :)

RecursiveVision:
It's savegame compatible.

Issue #7127, opened 2020-10-08
CTD on new game (probably due to Mapscript)


usadefcon1:
_1. Mod version: Oct 7th (10-7)

_2. Mod list (if using Vox Populi only, leave blank):_

![Screenshot (55)2](https://user-images.githubusercontent.com/57153924/95406260-fc66df80-08e7-11eb-8bf7-9c7e033ec0f4.jpg)

_3. Error description:_ 
CTD starting a new game with Venice. This is the last line in the Lua Log:

[93721.296] Runtime Error: C:\Users\nospa\Documents\My Games\Sid Meier's Civilization 5\MODS\(1) Community Patch\Mapscripts/Communitas.lua:8864: attempt to call method 'AttemptToPlaceStoneAtGrassPlot' (a nil value)
stack traceback:
	C:\Users\nospa\Documents\My Games\Sid Meier's Civilization 5\MODS\(1) Community Patch\Mapscripts/Communitas.lua:8864: in function 'NormalizeStartLocation'
	C:\Users\nospa\Documents\My Games\Sid Meier's Civilization 5\MODS\(1) Community Patch\Mapscripts/Communitas.lua:8909: in function 'BalanceAndAssign'
	C:\Users\nospa\Documents\My Games\Sid Meier's Civilization 5\MODS\(1) Community Patch\Mapscripts/Communitas.lua:960: in function 'StartPlotSystem'
	Assets\DLC\Expansion2\Gameplay\Lua\MapGenerator.lua:815: in function <Assets\DLC\Expansion2\Gameplay\Lua\MapGenerator.lua:779>
	=[C]: ?

[Logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/5344948/Logs.zip)

azum4roll:
Use Communitu_79a aka Communitas for VP.

I'll still fix this though, for any potential mapscripts overriding NormalizeStartLocation.

Mileslewis:
#7140

PR #7140, opened 2020-10-10
Mapscript compatibility fix + snow hills list

Mileslewis:
AttemptToPlaceStoneAtGrassPlot is being used in some mapscripts so is needed in the file.

Also added a list for snow hills to allow resources to be put on them. Snow hills are also now included in hills_list so should get some of the strategics on them which appear on other hills. hills_open_list is used for sheep so snow hills not included in this.

------------------------------------------------------------------------------------------

Btw AssignStartingPlots also has a file in the Non-EUI compatibility files which was not the same as the one in EUI compatibility files. Don't see any way in which EUI or Non_EUI affects this file so I'm guessing they should be the same and perhaps AssignStartingPlots should be in a different location?


Mileslewis:
#7127

azum4roll:
> Btw AssignStartingPlots also has a file in the Non-EUI compatibility files which was not the same as the one in EUI compatibility files. Don't see any way in which EUI or Non_EUI affects this file so I'm guessing they should be the same and perhaps AssignStartingPlots should be in a different location?

43-civ still has the old AssignStartingPlots since I have no experience on it to attempt to balance it. EUI and non-EUI should have the same file.

azum4roll:
Looks like I forgot to change 6b in my own pull request (for the current version). Oops! Can you include the 6b version in your PR? Just copy the 6a one over is ok.

Hope not many people are using 6b right now...

PR #7158, opened 2020-10-19
EvenMoreResources 19/10/20

HungryForFood:
### AssignStartingPlots.lua
- Backend code for Even More Resources mod.
- No changes if mod not activated.
- Adding unrelated mod stuff here to avoid conflicts, due to how AssignStartingPlots.lua requires overrides to be modded.


HungryForFood:
If mod is not activated, the only extra internal feature is the new plot list coconut_list.

This list includes all coastal plots with a jungle feature, or of latitude which is less than 15.

Issue #7284, opened 2020-12-13
Releasing installer automatically using github actions.


bukowa:
It's possible to execute a script using [github actions](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions) based on git actions (push/commit/release). For example [this action](https://github.com/actions/create-release) could be used to automatically create installer and make a release on [release page](https://github.com/LoneGazebo/Community-Patch-DLL/releases) of the project. Github hosted / self hosted runners [can run on windows server 2019](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#github-hosted-runners).

I don't know the process of building and creating installer for this project - with some help i could make a commit for this.

RecursiveVision:
@bukowa https://github.com/LoneGazebo/Community-Patch-DLL/issues/7018

bukowa:
Great it looks like there's a CLI https://www.kymoto.org/documentation/scriptstudio/index.html?CommandLine.html

Issue #7018, opened 2020-09-04
Installer Wizard


RecursiveVision:
@LoneGazebo Any chance you could share the software you use to create the executable launchers for new versions?

I don't intend to supplant you as project lead or anything, however I can think of several potential uses:
- If you ever become inactive again for an extended period of time due to real life obligations or problems, ilteroi or I could fill in for you with interim patches. I often make database changes and as a result some of ilteroi's interim builds didn't include my changes to reduce complication to the end user.

- It can be somewhat challenging to debug or test when a lot of changes to LUA files have been made, and since you already have a program that installs everything in the correct location, I could delete my MODS folder and use the special version I created to make sure all the changes get in correctly.

- It's useful to have auto-installer software if I want to send a friend a modified version of the game, or make a modmod, or get beta test feedback for a specific feature, etc.

Thoughts on this? I'm not sure whether the software you use costs money, but you seem to be able to post new versions relatively quickly once the code is straightened out.

LoneGazebo:
No worries! Here's what I use:

https://www.kymoto.org/products/inno-script-studio/

The .iss file is already on github, it is called CPPSetupData.

Note that there's a specific structure you need to follow for the folders. They have to look like this:

![image](https://user-images.githubusercontent.com/7849458/92254724-edc77b80-ee96-11ea-9ce4-a5a19d0d7013.png)

Note that the CP LUA and CBO LUA folders have to be moved out of the CP and the CBO into those external folders. If you do that, it should compile out of the box. :)

RecursiveVision:
Thanks! :)

bukowa:
@LoneGazebo Looks like there are some files missing in the github repository:
````
LicenseFile=C:\Users\Jeff\Documents\Community Patch Project Installer Files\License.rtf
InfoBeforeFile=C:\Users\Jeff\Documents\Community Patch Project Installer Files\Opener.rtf
SetupIconFile=C:\Users\Jeff\Documents\Community Patch Project Installer Files\installer.ico
````
Can you provide them?

bukowa:
@LoneGazebo There's also folder named "UI_bc1" that is not included. Is this a specific version? Is there anything else i am missing that is not included?

LoneGazebo:
Easier and safer to just download the inno files locally, compile, and then reload. In any case I've uploaded the requested files (except for the .modinfo files, as those come from compiled mods).

bukowa:
[...]

PR #7294, opened 2020-12-15
github action for automatic release & and installer #7284 #7018

bukowa:
This is a test working version, you can see it's output here, it compiles OK and the installer works: https://github.com/bukforks/Community-Patch-DLL/runs/1557261918?check_suite_focus=true

But before i proceed finishing i would love feedback from @RecursiveVision @LoneGazebo

There are files missing:
````
LicenseFile=C:\Users\Jeff\Documents\Community Patch Project Installer Files\License.rtf
InfoBeforeFile=C:\Users\Jeff\Documents\Community Patch Project Installer Files\Opener.rtf
SetupIconFile=C:\Users\Jeff\Documents\Community Patch Project Installer Files\installer.ico
````
#7284 #7018


RecursiveVision:
G would likely have more useful comments than I would, as he's set up the project. :)

Would I be able to create an installer .exe from a separate build using a forked/cloned repository using this script? That might be useful, especially to make an intermediate release which includes database changes.

I think at one point there was an attempt to upgrade to the latest EUI version, but a historical version seems to be in use, with VP modifying the files. If this upgrade ever happens the link would likely need to be changed.

That's all I can think of at the moment, but thanks for the contribution!

bukowa:
Yes, the github action can run on forks. There are multiple triggers that can execute an action. In the draft pull request it's for the "push" action for now. Later we can add "release" that will be executed if you manually trigger a release on github page - compiled installer will be uploaded.

> 
> Would I be able to create an installer .exe from a separate build using a forked/cloned repository using this script? That might be useful, especially to make an intermediate release which includes database changes.

Just right now i added "upload artifacts" action - it makes the compiled installer available to download, take a look [here](https://github.com/bukforks/Community-Patch-DLL/actions/runs/423603615) - at the bottom there's an artifact. If you include the files from my commits into your own fork - the action should run just fine.

If the pull request gets merged and someone makes a pull request to this repository - it will automatically build an artifact so other people can test it right away.

We can also include environment variables in the `CPPSetupData.iss` - like; commit hash, version number etc. to automatically have them in the installer based on commit / release version.

**Keep in mind** these settings should be configured aswell to avoid usage limits https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#artifact-and-log-retention-policy

bukowa:
Looks like i am wrong because you have to compile the files first - it needs CIV 5 SDK more info in #7018. I will close for now.

Issue #7293, opened 2020-12-15
Colonialism policy gives all Great Person Points (GPPs) in all cities? Not sure if bug


CokolwiekUnni:
Why does Colonialism (monopolies policy in Imperialism) does that? I've been looking for this in here and CivFanatics. Searched for Colonialism bug and other keywords, but I found nothing. Is this a recurring bug or intended effect? If intended is it additional and not mentioned in the policy info or is it due to some effect? If it is intended I think it should be completely removed.
I don't work any specialist, I still have flat 3 GPPs towards any great person in all of my cities as soon as I open this policy.

I rarely play Imperialism so don't remember exactly? But I've done a search on Milae's videos as I found it (Spain and Netherlands playthrough) and it provided the same bonus to him, so this bug (?) must be present since at least some versions.

And just to clarify Colonialism other bonuses works the way that flat counts multiple (so +9 happiness from three +6 happiness monopolies) but only once for percentage ones (so only 10% percent for having both sugar and salt monopolies, not 20%)?

My save is on 10-07-2 with Ilteroi hotfix, along with Historical Religions Edit (BNW or GK) (v 45)

[AutoSave_0193 AD-1330.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/5693539/AutoSave_0193.AD-1330.zip)

Mileslewis:
Yea, I think you also get a +10% modifier to GP generation from the policy which says monopoly bonus even though I don't think there is a monopoly which gives that.

CokolwiekUnni:
Alright guys, I made a tiny modmod to fix the issue for now

I didn't want to be locked into other policies by this bug so I simply edited out both flat and percentage modifiers for monopolies from Colonialism policy to get rid of both GPP bug and combat bonus bug. The tree is still strong if you know what you're doing and if you happen to possess mostly flat yield monopolies the nerf is negligible. You can easily throw some other thing into it like copying the line from Progress' Expertise that gives you +10% production if you miss original effect so much. Progress is in the Liberty folder. Changes are savegame compatible so you can add desired effect manually after gaining a particular monopoly, with text, if you so desire. Tested with Ingame Editor, everything works perfectly. I also added text for logging camps bonus from Industry's Entrepreneurship policy because it was missing the info previously, while providing the effect.
Just replace the files in (2) Community Balance Overhaul. All are in Balance Changes folder, one Policies, the other Text folders. Don't have CivFanatics account so can't post it there.

[...]

Issue #7215, opened 2020-11-22
Strategic Monopolies don't work after reloading the game


Mileslewis:
Patch: 10-07-02

Strategic Monopolies aren't doing anything when you reload a game. Seems like you need to either gain or lose a monopoly (of any kind) in order for it to update so that you start getting the benefits you should have.

Mileslewis:
Also this is still quite a major issue: #7120 
And I've just realised it only happens when you have Colonialism (Imperialism Policy).

bukowa:
I also cannot see strategic monopolies. 
![mspaint_L1e9oq5Sa5](https://user-images.githubusercontent.com/38087302/101238603-93ca9200-36e1-11eb-956a-bb316a779ec2.png)

![mspaint_ieVg3KXX02](https://user-images.githubusercontent.com/38087302/101238606-962cec00-36e1-11eb-931a-4ab59006fd7c.png)

![mspaint_mw6xgu8za7](https://user-images.githubusercontent.com/38087302/101238608-9927dc80-36e1-11eb-8c1f-cfae8e7dba9e.png)


**Save game:**
[AutoSave_0274 AD-1808.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/5646774/AutoSave_0274.AD-1808.zip)

RecursiveVision:
@bukowa You need at least 25% to have a strategic monopoly...

bukowa:
Yes but it's not the case in my post. You can see the coal and horses
visible in the first screen but after switching to strategic monopolies
there are only horses. Also all the other resources like iron etc are not
visible at all.

sob., 5 gru 2020, 16:10 użytkownik RecursiveVision <notifications@github.com>
napisał:

RecursiveVision:
Looks like it only shows up under Strategic Monopolies if a player actually has a monopoly.

bukowa:
Yes is this a big or a feature? Also other resources are not shown at all.

sob., 5 gru 2020, 18:32 użytkownik RecursiveVision <notifications@github.com>
napisał:

bukowa:
Oh sorry they are indeed visible my bad. Should i create different issue
about the 25% visibility?

sob., 5 gru 2020, 18:41 użytkownik ama bamo <pythoncontrol@gmail.com>
napisał:

Issue #7151, opened 2020-10-14
10/7 - CS influence not decaying


silentreply:
_1. Mod version (i.e Date - 4/17):_ 10-7-2

_2. Mod list (if using Vox Populi only, leave blank):_

_3. Error description:_ 
Something has stopped my influence from decaying with CS.  I don't have any beliefs/policies that should enable this such as statecraft trade routes. 

![image](https://user-images.githubusercontent.com/59859911/96021784-e42e1d80-0e1d-11eb-94a8-f0b66cbe2efb.png)
![image](https://user-images.githubusercontent.com/59859911/96021814-f0b27600-0e1d-11eb-9e3e-7679fac4518b.png)

Seems to apply only to allied CS, friendly are still decaying at usual rate:
![image](https://user-images.githubusercontent.com/59859911/96022607-19873b00-0e1f-11eb-9996-e188b9b745f0.png)

_4. Steps to reproduce (optional):_

---------------------------
Supporting information:
Please note that you can attach .zip files by dragging-and-dropping them. If possible, zip up all supporting data and post that way.

1. Log files (always attach your Logs folder, located at My Documents/My Games/Sid Meier's Civilization 5. Make sure you have enabled logging before experiencing an error! Go here to find out how: http://forums.civfanatics.com/showthread.php?t=487482):

2. Save game (always attach a save that was made a turn before the error; located at My Documents/My Games/Sid Meier's Civilization 5/ModdedSaves):

3. CvMiniDump.dmp file (attach if experiencing a game crash. Located at Program Files/Steam/steamapps/common/Sid Meier's Civilization V):

4. Screenshots (optional):

LoneGazebo:
Cold War?

silentreply:
@LoneGazebo Negative, world congress has not even met yet.

randomnublet:
I noticed for this to happen you only have to be allied and share religion. Also, sharing religion is enough for the influence to raise by 4 every turn if you angered the CS in past and its opinion is below 0.

silentreply:
@randomnublet That is indeed what's happening. I think the influence recovering faster when religion is shared was an intended change several versions ago, but I don't recall anything about freezing influence. The latter is probably too OP.

JanSobus:
Can confirm, noticed it in my game too

ricanuck:
Confirmed: any CS sharing religion stops decaying influence, be it as Friend or Ally.

WAD or BUG?

[...]

Issue #7120, opened 2020-10-04
Gaining a Monopoly gives you a temporary Combat bonus


Mileslewis:
Patch 8/16/2

Quite a strange one but it seems like when you gain a monopoly on a resource you then get a combat strength bonus on all your units equal to the number of monopolies you currently have x10%. It doesn't show up on tooltip but you can see it in the damage numbers.

Bonus is lost upon reloading the game.

Mileslewis:
New Info: This only happens when you have Colonialism (Imperialism Policy).

PR #7313, opened 2020-12-22
Fixes

Mileslewis:
Changed formula in CvPlayerTechs::GetResearchCost so that tech cost displayed should be closer to actual cost #6894

Removed tech overflow modifier by removing calculateresearchmodifier( from various places #6894

No longer count people you haven't met when working out tech modifier from known techs in calculateresearchmodifier(
-- Couldn't get it to ignore your own civ/team which is a bit annoying. Means you can never get up to 10% discount. -1 to iPossibleKnownCount would do the trick but it's pretty ugly.

Moved getUnitExtraCost in CvPlayer::getProductionNeeded(UnitTypes so that the flat production is added before modifiers. Affects settler costs on different gamespeeds.

Fixed Imperialism GPP bug. CvPlayer::getSpecificGreatPersonRateChangeFromMonopoly and CvPlayer::getSpecificGreatPersonRateModifierFromMonopoly #7293 

Fixed Imperialism Combat Strength bug in CvPlayer::UpdateMonopolyCache() #7120 
I just got rid of any extra combat bonus from imperialism. Not sure if you are meant to get a bonus 10% to combat bonuses you already get, doesn't mention this in the policy tooltip though.

added UpdateMonopolyCache into CvPlayerManager::Refresh so that monopolies work on reload. #7215 

Changed MINOR_FRIENDSHIP_RATE_MOD_SHARED_RELIGION to 50 to stop no friendship decay with shared religion. #7151
Literally just an xml value that was too high. Though perhaps the formula should be changed to multiplicative rather than additive modifiers to prevent this.


RecursiveVision:
Mmm, commits from June...and changing files ilteroi and I have also changed. I foresee merge issues!

I'll just merge your changes into my own branch. :)

RecursiveVision:
Done.

Mileslewis:
It was all good right? Also how do I get rid of the old commits which aren't relevant anymore?

RecursiveVision:
@Mileslewis There were no build errors. You *did* test it first, right?

To remove old commits just reclone master branch by deleting the fork and reforking/redownloading.

Issue #7295, opened 2020-12-15
Tooltip showing wrong Max Tribute %


MIS-g:
_1. Mod version (i.e Date - 12/1):_12/1

_2. Mod list (if using Vox Populi only, leave blank):_

_3. Error description:_ Tooltip showing wrong Max Tribute %

_4. Steps to reproduce (optional):_ Try to max heavy tribute.  Tooltip will say, e.g. can get 115 Faith because that is 93% of available.  But when get up to 100%, still only 115 Faith.  See: https://forums.civfanatics.com/threads/max-tribute.665415/

---------------------------
Supporting information:
Please note that you can attach .zip files by dragging-and-dropping them. If possible, zip up all supporting data and post that way.  See: https://forums.civfanatics.com/threads/max-tribute.665415/

1. Log files (always attach your Logs folder, located at My Documents/My Games/Sid Meier's Civilization 5. Make sure you have enabled logging before experiencing an error! Go here to find out how: http://forums.civfanatics.com/showthread.php?t=487482):

2. Save game (always attach a save that was made a turn before the error; located at My Documents/My Games/Sid Meier's Civilization 5/ModdedSaves):

3. CvMiniDump.dmp file (attach if experiencing a game crash. Located at Program Files/Steam/steamapps/common/Sid Meier's Civilization V):

4. Screenshots (optional):

See: https://forums.civfanatics.com/threads/max-tribute.665415/

Issue #7101, opened 2020-09-22
Tribute Policy Culture from tributing Scaling with Era


Mileslewis:
Patch: 8-16-2:
Currently seems like the culture bonus does scale with era for when you tribute any yields other than food. (not sure about gold actually).

I did mention before that tribute policy culture amount probably shouldn't scale with gamespeed because the tribute amount you are getting should already scale with gamespeed.

This also applies to scaling with era imo, but should be all yields or none.

PR #7325, opened 2020-12-24
Tribute fixes

Mileslewis:
- GetBullyValueModifier no longer applied twice to tribute values
- removed the rounding to the nearest 5 (it wasn't working correctly anyways and is more trouble than it's worth imo)
- Value now shows 0 when you can't tribute
- iGlobalMilitaryScore now actually factored into tribute score
- Heavy tribute now includes the extra negative modifier it should have
- Culture yield from tributing Food and Gold with Tribute policy now scales with era like other yields. Tbh it probably shouldn't for all of them but should be all or none and most already did.
- Culture yield from tribute policy no longer scales with gamespeed (the tribute value already does).

#7295 #7101 done. Pretty sure there is now no more weirdness from tribute values or tooltip. (tested).

Didn't want to change game balance but:
- Gold tribute should probably scale the same as heavy tribute (Heavy tribute currently increases more as game goes on) or at least have lower penalties or noone will ever do it.
- iGlobalMilitaryScore should be based on your military compared to average instead of yours as a percentage of total military so that it is fair with different numbers of civs.
- Actually wouldn't be too hard to include Civs protecting a City State's military into iOurPower so that iLocalPowerRatio was reduced by protective civs in the area.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvMinorCivAI.cpp
+++ b/CvGameCoreDLL_Expansion2/CvMinorCivAI.cpp
@@ -15074,13 +15074,6 @@ int CvMinorCivAI::GetBullyGoldAmount(PlayerTypes eBullyPlayer, bool bIgnoreScali
 	iGold *= (100 + GET_PLAYER(eBullyPlayer).GetPlayerTraits()->GetBullyValueModifier());
 	iGold /= 100;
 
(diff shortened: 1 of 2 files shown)


RecursiveVision:
Merged.

PR #7333, opened 2020-12-30
City State Unit Fixes

### LINKED ISSUE #7332: Military City-States not selecting their unique unit gift randomly
_1. Mod version (i.e Date - 12/1):_

12/1.

_2. Mod list (if using Vox Populi only, leave blank):_

Usually play with 3/4th component but still happens if it's left out.

_3. Error description:_ 

Military City-States don't seem to be selecting their unique units randomly, as I was constantly seeing the same units selected, notably Musketeer/Samurai/Musketman.

I did some experimenting and it seems to be selecting them in an order, as putting France/Japan/America in the game would consistently make them take other vanilla civ units like Jannisary/Cossack/Naresuan's. Adding those civs aswell they would select more vanilla civ units and also mix in some DLC ones like Denmark/Spain. Adding all of those to a game would make them select G&K civ units like Babylon's Bowmen.

Mileslewis:
#7332 
- Each unit now has a different random seed (before they all had the same I think)

- PseudoRandomChoiceByWeight( now puts them in descending order before using the first 3 to choice randomly from. This means those with highest weight will now be chosen instead of 3 lowest.

- Now pick from top 5 choices instead of 3.

- Included units from the same era as game start. Given they are valid until they obsolete you will still always get at least until Medieval and usually to renaissance to get ancient era units.

- EDIT: sorted unique unit not always being picked on initialization, needed a higher iNum in getSmallFakeRandNum( for some reason.

Seems like there was some attempt to get removed Civ 5 units to spawn from city states, was that abandoned or just doesn't work? If we want them in they should probably be buffed a bit as well...

FOR RECURSIVE: You already have the changes from CvMinorCivAI and CvPlayer merged


RecursiveVision:
@Mileslewis You can't change globals from CvGlobals.cpp like that, any entry in the SQL/XML database will override it. You'd need to add this code to a SQL file:
```
UPDATE Defines
SET Value = '5'
WHERE Name = 'UNIT_SPAWN_NUM_CHOICES';
```

RecursiveVision:
To save you some trial and error, if searching for the database value of a value, just use Notepad++'s "Find in Files" feature to search through both the Community Patch DLL and C:/Program Files (x86)/Steam/steamapps/common/Sid Meier's Civilization V, using file filter: *.sql *.xml

When updating a define from the base game, the above syntax (UPDATE DEFINES) is correct.

When adding a define into the database (e.g. if UNIT_SPAWN_NUM_CHOICES didn't exist already), the correct syntax is:
```
INSERT INTO Defines (Name, Value)
SELECT 'UNIT_SPAWN_NUM_CHOICES', '5';
```

Mileslewis:
Okay thanks, tbh I couldn't find it anywhere so I thought it might just be being initialised there. Not a big deal if it's changed or not but i'll add it anyways.

RecursiveVision:
Merged into rsi branch.

PR #7375, opened 2021-01-09
Misc 09/01/21

HungryForFood:
- Feature parity for the 43 civ version of AssignStartingPlots.lua (compatibility with Communitu_79, Even More Resources).

Should be easy to merge for the next beta version since no one changed this file.

On a side note, there are so many versions of AssignStartingPlots.lua in the repository. Which actually gets included in installers?


RecursiveVision:
merged

PR #7461, opened 2021-02-05
Misc 06/02/21

HungryForFood:
- Added new Lua method, Player:DoInstantYield. Works with the Community Patch's instant yield system for a more convenient way to add yields, especially for the newly added ones.
- Added null check for several CvUnit::plot() calls.


RecursiveVision:
Merged into #new-beta

PR #7488, opened 2021-02-11
City state unit spawn changes

Mileslewis:
- Stop spawning units option now freezes the counter
- Naval unit gifts from minors will now spawn near the minor capital instead of nearest major's city if major's city is not coastal. (Minors should only offer naval units if their capital is coastal)
Maybe someone else could make this cleaner but I have tested it and it does the job.


RecursiveVision:
There's CvCity::isCoastal() :)

I think it'd be preferrable to iterate through cities by distance, this iteration would only be called when a unit is actually spawned under these special circumstances so the performance impact is negligible.

Mileslewis:
Yea that was my intention but I didn't really understand the GET_PLAYER(eMajor).GetClosestCityByPathLength(pMinorCapitalPlot); function or how to modify it easily to just consider coastal cities or to iterate through

RecursiveVision:
I'll see if I can fix it to do that...

RecursiveVision:
I rewrote this function in new-beta.

PR #7489, opened 2021-02-11
Misc 11/02/21

HungryForFood:
NOT save game compatible.

- Fixed IMPROVEMENTS_EXTENSIONS in CustomModOptions not working.
- Fixed Route_QuantityRequirements and QuantityRequirement column in Improvement_ResourceTypes not fully working (did not change resource count correctly in every case, did not check if player had enough resource before building). Requires IMPROVEMENTS_EXTENSIONS.
- Moved all route/improvement maintenance checks/changes into CvPlot::SetPlayerResponsibleForImprovement and CvPlot::SetPlayerResponsibleForPlot.
- The responsible player is now always set, instead of just when the route/improvement is built in neutral territory (memory already allocated, let's just use it).
- Total number of improvement/route a player is responsible for now cached, and can be accessed via Lua (disabled if IMPROVEMENTS_EXTENSIONS not enabled).
- Necessary UI changes to display improvement/route resource requirements.

### DIFF
--- a/(6c) 43 Civs CP/EUI/TopPanel.lua
+++ b/(6c) 43 Civs CP/EUI/TopPanel.lua
@@ -2430,6 +2430,100 @@ local function ResourcesToolTip( control )
 					end
 				end
 			end
+			for improvement in GameInfo.Improvements() do
+				local improvementID = improvement.ID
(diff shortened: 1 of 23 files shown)


HungryForFood:
Wrong branch to merge into.

PR #7490, opened 2021-02-11
Misc 11/02/21

HungryForFood:
NOT save game compatible.

- Fixed IMPROVEMENTS_EXTENSIONS in CustomModOptions not working.
- Fixed Route_QuantityRequirements and QuantityRequirement column in Improvement_ResourceTypes not fully working (did not change resource count correctly in every case, did not check if player had enough resource before building). Requires IMPROVEMENTS_EXTENSIONS.
- Moved all route/improvement maintenance checks/changes into CvPlot::SetPlayerResponsibleForImprovement and CvPlot::SetPlayerResponsibleForPlot.
- The responsible player is now always set, instead of just when the route/improvement is built in neutral territory (memory already allocated, let's just use it).
- Total number of improvement/route a player is responsible for now cached, and can be accessed via Lua (disabled if IMPROVEMENTS_EXTENSIONS not enabled).
- Necessary UI changes to display improvement/route resource requirements.


HungryForFood:
## Misc
- Added new column 'FightWellDamaged' to UnitPromotions. Units do not get a combat penalty from being wounded, but also does not gain the combat bonus of 'StrongerDamaged'.
- Added ability to have processes unique to a civilization.
## UI
- EUI top panel no longer incorrectly includes free buildings when counting resources used.
- Civilopedia now shows the Civilization if there is a define in the 'CivilizationRequired' column in table 'Buildings'.
- Buildings unavailable due to the 'CivilizationRequired' column in 'Buildings' will no longer show up in the tech tree.
- When checking for corporations to populate the corporation tech, the tech tree now checks if the player is forbidden from building the headquarters due to the selected civilization ('CivilizationRequired' in 'Buildings', 'Civilization_BuildingClassOverrides' and a null replacement).
- The available corporations part of the UI (shown before founding a corporation) now fades out corporations which cannot be founded due to the player's civilization not allowed to build the headquarters ('CivilizationRequired' in 'Buildings', 'Civilization_BuildingClassOverrides' and a null replacement).
- Adjusted height of each row for the available corporations UI to remove text clipping.

HungryForFood:
- The ideology tenets which grant a new unit type now replaces the relevant generic version (eg Heavy Bombers will become B17). This will convert all existing units, and units will upgrade into the new unit type.

RecursiveVision:
Merged into #new-beta.

PR #7497, opened 2021-02-14
Fix: UnculturedHappinessTraitMod not being applied

daniel1709:
Was reading (or rather trying to read) from a non existent - and therefore assumingly "wrong" - column...


RecursiveVision:
Merged.

PR #7514, opened 2021-02-21
Fixed bug where global event buildings were not being added properly

Stickerbrush:
I encountered an issue with the civ specific events from engisneer, apparently the dummy buildings were not being built at all after making a choice.

Let me picture exactly how I encountered the bug:

While playing Celts, there's a civ specific choice that lets you get a promotion on melee units and also +33% great musician rate in all cities. The first part was working, but the second half wasn't. After messing around with possible culprits, I came to the conclusion that the dummy building that boosts the great musician rate by +33% was not being built when selecting that choice.
Upon inspecting the cpp files of the dll and doing some trial and error, I found out that the reason why this happened was because the type of the building that was being passed was 'NO_BUILDING' because the function that was getting the building type, was trying to do it under the assumption that a building from the same class was already present in the city, which is not the case for most dummy buildings in the E&D core.

This should at the very least fix all the civ-specific events that create dummy buildings in cities to boost stuff/yields


RecursiveVision:
Merged.

PR #7519, opened 2021-02-25
Revise Casus Belli & Global Peace Accords voting logic

RecursiveVision:


RecursiveVision:
@LoneGazebo Just finished this one. Should be the last of my changes! :)

LoneGazebo:
Awesome! Do you have a running list of balance changes?

RecursiveVision:
@LoneGazebo Don't think much was changed in regards to balance as that's mostly your purview, but here's what I remember. :)

```
[Celtic Pantheons]
Morrigan, the Harbinger
Loses +5 HP healed per turn in friendly territory
Gains +Gold, Culture, GAP instant yields if you pillage a non-road improvement

Dagda, the All-Father
Gains +5 HP healed per turn in friendly territory

[Spain]
A few changes from PAD: 49461e57060ee572a3bac9cb176aada355045902

[Diplo]
Human now uses the base AI WarmongerHate flavor for their chosen civ (for calculating warmonger score and anti-warmonger penalty)

[General]
Cities now heal faster when no enemy unit is within 3 tiles of the city (visible or not)
```

Think that's all.

RecursiveVision:
Edit: Added a few more things.

Issue #7479, opened 2021-02-08
(For Gazebo) Issues trying to build the mods from github repository


Mileslewis:
So since you were away I had a go at making the installer using the instructions you provided in #7018.

Actually got it to work after a few attempts (my first involved making a new User on my comp to try and sync up the file source locations lol). Pretty sure it works fine but I could only get it to work by using the pre-existing built Mod folders from a previous version and then adding in the changed files.

From the thread I mentioned it seemed like the idea was to be able to download the Github repository, then build the individual mods (using ModBuddy) and then put them all into the correct folders for the installer but these were the issues I ran into while trying to build the mods:

No Project files or solution files for the (Edit): '(3), (4) or (5)' folders on Github as far as I could see.

The project for (1) threw an error when I tried to build it in ModBuddy. This was due to missing mp3 files which were supposed to be in the (1) -> Modules -> MoreGP -> Extra_GPs_items folder.
The others mods did build, though I wasn't sure whether the project files were up to date or not.

One other issue I had was that the UI_bc1 folder from github did not have the EUI_0, EUI_1 and EUI_2 CIV5PKG files which were in the UI_bc1 folder I had installed from a previous VP version. EUI did not work without these.

LoneGazebo:
I'll work on this this week.

LoneGazebo:
On the same topic, @RecursiveVision is @ilteroi's mil branch in your latest merge?

RecursiveVision:
@LoneGazebo A handful of recent changes are missing, I'll merge them for you though, just give me a couple minutes :)

RecursiveVision:
Merged.

ilteroi:
hate to mention it again but you did not merge anything as far as git is concerned. you created a branch with conflicting changes!

git is the best tool on the planet. make it work for you, don't work against it guys!

RecursiveVision:
@ilteroi Yes, I know, but as I mentioned, mil and mrg would not merge automatically into master. I'll avoid rewriting history in the future, I was not anticipating that you were going to merge again...

Your changes are all included in master, if you rebase off of that it should be good.

ilteroi:
you just did it again :)

tip: if you want a commit from my branch without merging you cherry-pick it. you do not need to "transfer" it manually.

RecursiveVision:
[...]

PR #7529, opened 2021-03-01
Fix broken merge & a few bugs

RecursiveVision:
@LoneGazebo This should fix the issues with the merge.

Not sure why you added a second 43 Civs CP folder to the repository, though?


RecursiveVision:
If you can create a new installer I can do testing...however as mentioned in #7479 I can't create the installer on my own as there are missing files or something...

PR #7531, opened 2021-03-02
Fix other merge errors

RecursiveVision:
@LoneGazebo There are still a few merge snafus as well as one build error from your changes; this should fix them. WinMerge for the win.

The 6 deleted files can be deleted; I integrated the bugfixes and moved the config stuff to the core files for easier organization (I did a huge cleanup of "zombie" settings, although there is more work to be done).

If you're planning to include a separate EnemyUnitPanel.lua file for 43 Civs CP rather than using the version in CoreLua, then don't delete that one, but otherwise those two should be deleted too.

P.S. Your added checks to CvMinorCivAI::DoSpawnUnit() are somewhat puzzling...I already added checks at the start of the function so that City-States with no capital or a null plot cannot spawn units :)


LoneGazebo:
>P.S. Your added checks to CvMinorCivAI::DoSpawnUnit() are somewhat puzzling...I already added checks at the start of the function so that City-States with no capital or a null plot cannot spawn units :)

I caught a null pointer crash in that function where I added the check. Still able to get a null pointer through without that last check.

RecursiveVision:
Strange.

PR #7543, opened 2021-03-06
Fix various issues & Consolidate enums

RecursiveVision:


RecursiveVision:
@LoneGazebo I found what was causing the bug...multiple initialization of the pUnitPlot variable within the brackets :)

Removed your extra checks as they are now redundant (already a != NULL check for every city and their plot).

LoneGazebo:
@RecursiveVision anything else before I issue a new version?

RecursiveVision:
@LoneGazebo Nah, due to personal issues I'll be out of commission for a while.

RecursiveVision:
@LoneGazebo Having to stop a medication with harsh withdrawal symptoms. Not fun!

RecursiveVision:
@LoneGazebo Not in a healthy modding mindset so can't do it myself atm, but prior to the next release you may want to do something to address the voluntary vassalage complaints, there's a lot of them.

My eventual intention is to make it so voluntary masters are dragged into any wars their voluntary vassal is in (without kicking out troops etc) and that the AI understands when and when not to accept voluntary capitulation. I already reduced their overeagerness to become a vassal.

A simple interim patch would be to block voluntary (not capitulated) vassalage if the civ is at war with any major and/or any of their cities are in danger, though.

LoneGazebo:
@RecursiveVision i'll take a look, yep.

LoneGazebo:
@RecursiveVision my guess is that all I need to add is this line (the 'else' bit) in DealAI for vassalage value:

```
if (m_pPlayer->IsAtWarWith(eOtherPlayer))
		{
			return (iItemValue / 2);
		}
		else
		{
			if (GET_PLAYER(eOtherPlayer).IsAtWar() || GET_PLAYER(eOtherPlayer).HasCityInDanger(false, 0))
				return INT_MAX;
		}
```

RecursiveVision:
@LoneGazebo That should work, yeah.

I'd choose one of "is at war with any major" or "has city in danger from any major", whichever you think best.

Not sure IsAtWar filters out minors?

LoneGazebo:
It doesn't filter minors, but either way I think fully at peace should be the norm for voluntary vassalage. Keeps hiccups from happening.

PR #7585, opened 2021-03-14
Rsi

RecursiveVision:
@LoneGazebo Feeling unexpectedly better after lots of sleep, so here's something. :)


RecursiveVision:
Also, ilteroi fixed a "nasty crash", so you may want to merge that as well...

PR #7596, opened 2021-03-18
World congress changes

Mileslewis:
Changes are not done yet but since this has some fairly major changes I thought I'd make a pull request so you can check what I'm doing.

My main idea is to have a civ first evaluate a proposal based purely on how it directly affects them. So if it helps them win the game it's positive or if it hinders them directly it's negative. 
Next it will loop through all the other participating civs and check how the proposal directly affects that civ. Then in some way combine that with our feelings towards them (if we want to help or hinder them and by how much) and then add those in to create the final score.

The overall structure has been done, along with removing a lot of extra calls to the score function as I know these changes could decrease performance if score function gets called a lot. Also now there is no randomness in the score and scores do not get changed (apart from being occasionally zero'd) outside of the main score function. Full changes are listed in the Github commits.

Still to be done:
- Update the values for when a civ is evaluating how a proposal directly affects them.
- Find a good formula for figuring out how much we want to help or hinder each civ.
- Update the evaluateDesire function to better reflect new scores
- Small changes to vote allocation function for rare cases where we shouldn't keep adding votes to a proposal that will 100% pass
- Probably change to proposal allocation function to choose the top scored proposal rather than one of the top ones.
- Update values for the less used ScoreVoteChoicePlayer function which is used for host and world leader sessions, if necessary (I haven't really looked at it)


Mileslewis:
Okay I think it's good to go. I have playtested it and it all seems to work well but I'm sure it will need some further changes with feedback.
Main changes:
-  When a proposal is being scored by a civ it will first score how the proposal affects itself, then loop through all other civs and add on their score * DiploModifier which is how much the civ wants to help (+) or hinder (-) the other civ.
- When a civ is choosing host or world leader it will only choose either civs on the same team or civs they have a good alignment (from EvaluateAlignment) with who also have many more votes than them.
- Vassals will no longer automatically support their masters but their vassal relationship is factored into the alignment.
- Votes are now allocated to the highest scored proposal/choice. The score of this choice is then decreased by 15% before the next vote is allocated.
- Proposals are now allocated to highest scoring proposal (instead of choice of one of the highest).
- Many small changes within scoring proposal function and evaluatealignment function to bring them up to date.
- Updated vote trading value function. World Leader votes can no longer be bought.

RecursiveVision:
Merged into RSI

PR #7641, opened 2021-04-02
Adding GetOriginalCapitalPlot, ChangeYieldFromTraits

inkyant:
I've made a function `city:ChangeYieldFromTraits(eYield, iChange)`, both for lua and DLL.
Like the name says, it gives the yield but it shows as up as coming from the trait.

I've made a lua function `player:GetOriginalCapitalPlot()`
Obviously returns the plot of the player's original capital.

Tested both and it looks like they work.


RecursiveVision:
Merged into RSI

Issue #6657, opened 2020-06-08
better differentiation between tercio and musketman


ilteroi:
i don't if i'm the only one but i constantly confuse tercios and musketmen.

which is which? would it be possible to use more distinct icons?

![image](https://user-images.githubusercontent.com/1358966/84036621-b9c6e200-a99d-11ea-8254-b99f2e14b5e8.png)

LoneGazebo:
I don't have the art, but someone might make it for us. They're both vanilla assets.

ASCII-Guy:
@LoneGazebo here you go. I extracted the original from Expansion1UITextures.fpk

DDS download: https://drive.google.com/file/d/1bWUnzPw_3wqwX3-2gcRK-IScfOL58FDk/view?usp=sharing

![sv_tercio](https://user-images.githubusercontent.com/14044545/84089338-8dae5f80-a9a3-11ea-879a-52615332d940.png)

azum4roll:
I confuse between Spearman and Tercio more.

ilteroi:
spearman and tercio are different eras ...

pineappledan:
making the halberd heard slightly bigger would resolve the issue, no?

pineappledan:
https://github.com/LoneGazebo/Community-Patch-DLL/pull/7648

PR #7648, opened 2021-04-03
Tercio flag change

pineappledan:
Changes the tercio flag to a single, slightly larger Halberd. 
This should help differentiate it a bit from the musketman
https://github.com/LoneGazebo/Community-Patch-DLL/issues/6657


RecursiveVision:
Merged into RSI

PR #7702, opened 2021-04-14
Fix issues in new dogpiling check + CS military ranking %

RecursiveVision:


RecursiveVision:
@LoneGazebo Also I'm going to change it so your Pledge is annulled if you go below 0 Influence.

RecursiveVision:
Actually, since I fixed it so breaking a PtP doesn't give you the jerk status anymore, the only loss from that is a small amount of Influence (if above -60) and the inability to make a PtP for a while. Wouldn't be too bad of a penalty. What do you think?

RecursiveVision:
@LoneGazebo What are your thoughts on the effect on Great Diplomats?

PR #7912, opened 2021-05-25
Text when Proposing Resolution Update

Mileslewis:
- The screen now shows how all civs agree each civ is affected by a resolution. This is not necessarily the same as their voting as they may want to hurt enemies or help friends.
- They are now colour coded based on how positive or negative the proposal is to each civ. To know how a civ will actually vote you need a Diplomat/Ideology.
- If Transparent Diplomacy is on they will be able to see the actual score and those civs who are neutral will be shown in white.
- Some functions moved from private to public as needed.
- Text changes to be clearer about what is being shown.


Mileslewis:
Screenshot (34)_edited
Screenshot (35)_edited

Some pics to show how it looks with the changes (with Transparent Diplomacy ON)

PR #7922, opened 2021-05-26
Remove FAutoVariable from CvUnit

IanE9:
### **Please provide feedback, it is appreciated!**

Appears to be save game compatible 👍
Heavily untested 👎

This replaces all instances of FAutoVariable with a new system I've written using CvSyncVar.
This is a rather striking change so please provide feedback.

The primary advantage here is the reduction of CvUnit's size from **5220** bytes to **2272** bytes.
A secondary advantage is that the system allows CvSyncVars to be tagged as automatically saved or not.

The system uses a whole lot of template and macro wizardry to generate "trait" structures describing variables.
The variable trait structures are organized into a sync object structure which exists primarily for multiplayer data similar to Fraxis' archive structure.
The variable traits are referenced by CvSyncVar templates to automatically generate behavior.

Variable deltas for multiplayer are stored as a std::set of indices referencing each changed variable. The deltas are written using an automatically generated virtual dispatch table (see CvSyncObject<CvUnit>::s_SyncWriterTable).
In order for deltas to be tracked, a sync var object must be provided an appropriate sync object.

Example:
```
void CvUnit::ChangeRangeAttackIgnoreLOSCount(int iChange)
{
	VALIDATE_OBJECT
	if(iChange != 0)
	{
		m_iRangeAttackIgnoreLOSCount.dirtyGet(m_syncObject) += iChange;
	}
}
```

Otherwise CvSyncVar is just a wrapper around another type.

It should be noted that m_syncArchive has been maintained as a member of CvUnit because the engine wants to access it even if it doesn't do anything.


LoneGazebo:
Into master? Not yet! Let's test first. Master is sacred!

IanE9:
@LoneGazebo
Apologies. Is there another branch you'd prefer I create a pull request for?

LoneGazebo:
just make a new branch off of master and pull to that.

IanE9:
I don't have permission to create branches in the original primary repo.

RecursiveVision:
@IanE9 I made the autovariables branch for you and pulled.

Issue #7753, opened 2021-04-22
Suspicious voluntary vassalage


LukaSlovenia29:
_1. Mod version (i.e Date - 4/14):_
4/16/21

_2. Mod list (if using Vox Populi only, leave blank):_

_3. Error description:_ 
The save file is on the turn before Poland (#3/4 in score) becomes a voluntary vassal to #2 Ethiopia.

_4. Steps to reproduce (optional):_
Click next turn.
[Turn B4 Poland Vassalage.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/6359562/Turn.B4.Poland.Vassalage.zip)

---------------------------
Supporting information:
Please note that you can attach .zip files by dragging-and-dropping them. If possible, zip up all supporting data and post that way.

1. Log files (always attach your Logs folder, located at My Documents/My Games/Sid Meier's Civilization 5. For instructions, go to the repository's main page, under "To enable logging for bug reports"):

2. Save game (always attach a save that was made a turn before the error; located at My Documents/My Games/Sid Meier's Civilization 5/ModdedSaves; you can change autosave frequency in the game's Options menu):

3. CvMiniDump.dmp file (attach if experiencing a game crash. Located at Program Files/Steam/steamapps/common/Sid Meier's Civilization V):

4. Screenshots (optional):

RecursiveVision:
Can't load the save due to the broken savegame system.

RecursiveVision:
I dropped voluntary vassalage willingness a little more but not much else I can do at the moment.

PR #7923, opened 2021-05-26
Autovariables pull

RecursiveVision:


LoneGazebo:
@IanE9, once we test CvUnit, do you want to apply the same method to CvCity and CvPlayer? I know they won't have as much of an impact, but the consistency would be nice.

LoneGazebo:
I also wonder if the portability of these memory changes will improve multiplayer stability re: sync issues.

IanE9:
I would like to do the same with the other classes, I'll need to test multiplayer pretty extensively though. I got around to examining Fraxis' interfaces for syncing multiplayer just before I went to sleep last night and ended up deeming it pretty dreadful and I'm going to have to retrofit some stuff to make sure all the new code I have remains compatible with them. It probably wont impact sync stability much, but understanding the system may make it easier for me to debug desyncs.

RecursiveVision:
It would be awesome if we could play VP in multiplayer. :D 

Also if you're fixing serialization on a larger scale maybe you'd be able to find why switching between release and debug configs makes saves impossible to load, sometimes but not always. I have no clue why that happens but it sure makes debugging a pain.

Serialization in general is pretty bad...

LoneGazebo:
You can reproduce the release->debug issue by compiling release, starting game, saving, then trying to load on a debug DLL..

RecursiveVision:
I've sometimes been able to get it to work, and it seems more likely in the early game. That's what makes it so annoying...

Perhaps I can ask for unmodded late game saves on the forum and test them out until the behavior is identified.

RecursiveVision:
#7777, #7753

These are from the 4/16 version...

IanE9:
> You can reproduce the release->debug issue by compiling release, starting game, saving, then trying to load on a debug DLL..

I only asked for a save because the saves I've been using for testing have loaded no problem in both configs ^^'

> These are from the 4/16 version...

Recent saves compatible with the current version are best. I tried #7777's save and it claims to have DLC I don't anyway, so I'll try to reproduce the issue locally later.

RecursiveVision:
@IanE9 Do you own all of the Civ V DLC?

Might be other mods affecting savegame compatibility...

IanE9:
The only DLCs I don't have are map packs. I hope that's not impacting it but maybe.

PR #7936, opened 2021-05-27
Merge Latest

IanE9:
Mostly changes related to making the new stuff work with Fraxis' interfaces for sync.

CvSyncObject has been renamed to CvSyncArchive and consolidated it with CvUnit's m_syncArchive since it needs to derive from Fraxis' FAutoArchive in order to work properly with multiplayer. FAutoVariableBase is also required to actually track changes, so I've setup a scheme in which generated interfaces are allocated for CvUnit but **ONLY** in a multiplayer game.
Because we rely on engine interfaces in order to actually maintain multiplayer compatibility have dropped my generated vtable for writing out sync deltas (a shame because it's a much lighter weight scheme than Fraxis').
Overall these changes resulted in CvUnit's size shrinking further.

The changes I've made _may_ result in marginally faster singleplayer loading because of reduced levels of indirection but I'm not sure of the best way to benchmark it.

I did have to bump up VC's PCH memory limit to 150mb because memory was being exhausted otherwise. Because of this I encourage others to try building the branch and be vocal about any concerns the may have about the complexity of the template stuff going on in CvSerialization.h so that I can tackle it sooner rather than later.


LoneGazebo:
@ilteroi has a memory debugger somewhere....
also you can use [VerySleepy](http://www.codersnotes.com/sleepy/) for general performance modeling

PR #7937, opened 2021-05-27
Fix multiplayer never trying to sync CvUnit

IanE9:
Oops. Multiplayer is working though.


RecursiveVision:
@IanE9 So you said you found really bad sync code for multiplayer, does that mean the constant resyncs that make multiplayer difficult to play could be fixed?

IanE9:
@RecursiveVision Sorry, my comment just meant that I felt the interfaces were poorly designed and expensive as a result. Desyncs are going to keep happening until someone takes the time to play a lot of multiplayer and do a lot of debugging. It looks like a good place to start some debugging would be to just log any failures on FAutoVariableBase::compare(). It's the function that the engine seems to call to decide if the game is actually in sync or not.

RecursiveVision:
Unfortunate! Oh well.

PR #7941, opened 2021-05-28
Misc 28/05/21

HungryForFood:
- Fixed potential crash in CvPlot::isCoastalLand (null Area pointer).


HungryForFood:
@ilteroi since you changed this code last, I'm not familiar with how the area stuff works. This makes sense?

PR #7986, opened 2021-06-05
Defer database processing to engine load

IanE9:
CvGlobals::init() currently caches and processes the game database, but this operation appears to be unnecessary and serves to make startup times worse.

Immediately after CvGlobals::init() is invoked the engine retrieves and requests a flush of the database and then performs the cache and process operations a second time.

This reorders some operations in order to try and make the cache that the engine performs the only time the database gets cached and subsequently processed.

I haven't observed anything unusual from these changes and I see no reason why this would cause problems. I've tested this as both a standard mod installation and as a modpack; **however if you do believe this might cause problems let me know and I'll do more testing in any areas requested.**


LoneGazebo:
valhalla

PR #8016, opened 2021-06-09
Remove FAutoVariables in CvPreGame

IanE9:
s_preGameArchive isn't exported or referenced. As far as I can tell all of this is residual.


LoneGazebo:
This likely breaks savegames so I'll close here, but we can merge into :autovariables.

IanE9:
Gotcha. I'll merge it locally with my most recent serialization changes and make a pull request.

PR #8029, opened 2021-06-11
Minor clean-ups in PreGame

IanE9:
No fixes, just cleanliness


LoneGazebo:
fun fact- the pregame code is legacy code from Civ4!

IanE9:
I was certainly getting that feeling

PR #8037, opened 2021-06-13
Fix WB scenarios never loading in modpack installations

IanE9:
**This only fixes the bug if a user installs EUI**
**An identical bug exists in the vanilla GameSetupScreen**

When loading a world-builder map, there is a tick box that is meant to control whether or not the associated scenario script should be loaded. For some reason someone decided to override the behavior of this tick box (among other wack things I can't begin to comprehend) when no mods are loaded. This means that if a player uses a modpackesque install, world-builder scenarios cannot be loaded.

This removes the erroneous check for mods being loaded from the EUI GameSetupScreen so that world-builder scenarios will actually load! Hooray! \o/

### DIFF
--- a/UI_bc1/GameSetup/GameSetupScreen.lua
+++ b/UI_bc1/GameSetup/GameSetupScreen.lua
@@ -36,7 +36,7 @@ Controls.StartButton:RegisterCallback( Mouse.eLClick,
 function()
 	PreGame.SetPersistSettings( not bIsModding ); -- Whether or not to save settings for the "Play Now" option.
 
-	if bIsModding and isWBMap( PreGame.GetMapScript() ) then
+	if isWBMap( PreGame.GetMapScript() ) then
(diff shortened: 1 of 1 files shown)


RecursiveVision:
@IanE9 bc1 doesn't like us modifying the EUI files directly.

IanE9:
> 
> 
> @IanE9 bc1 doesn't like us modifying the EUI files directly. Wants us to do it only through the (6) mods.
> 
> That said, I can't see a way to fix this otherwise. Thoughts, @LoneGazebo?

Ah, my apologies, I was unaware. If all else fails could bc1 not just implement the fix themself? There's no way for the DLL to know why or where from ``PreGame.SetLoadWBScenario()`` is being called by Lua so I agree that there likely exists no clean way to fix this other than changing the Lua files themselves.

RecursiveVision:
Screw it, there's no other way to fix it.

Merged into rsi.

PR #8043, opened 2021-06-14
Add TRTurnModGlobal/Local to Buildings

inkyant:
TRSpeedBoost doesn't actually do much, this decreases the turns for a TR to complete.


RecursiveVision:
Merged into rsi.

PR #8048, opened 2021-06-14
AdjacentResourceYieldChange only works for revealed resources

inkyant:


CvGameCoreDLL_Expansion2/CvBuilderTaskingAI.cpp
```diff
@@ -2320,7 +2320,7 @@ int CvBuilderTaskingAI::ScorePlotBuild(CvPlot* pPlot, ImprovementTypes eImprovem
 			for(int iJ = 0; iJ < GC.getNumResourceInfos(); iJ++)
 			{
 				ResourceTypes eResource = (ResourceTypes)iJ;
-				if(eResource != NO_RESOURCE)
+				if(eResource != NO_RESOURCE && GET_TEAM(getTeam(m_pPlayer->GetID())).GetTeamTechs()->HasTech((TechTypes)(GC.getResourceInfo(eResource)->getTechReveal())))
```
IanE9, line 2323: This ought to be done inside ``CvPlot::ComputeYieldFromAdjacentResource()``

``iAdjacentResourceValue`` only impacts whether or not ``CvPlot::ComputeYieldFromAdjacentResource()`` is called. In fact all of the ``iAdjacentValue*`` variables should be booleans and the loops should break at the first true condition.

Doing this inside of ``CvPlot::ComputeYieldFromAdjacentResource()`` should ensure that force revealed resources are taken into account if ``CvPlot::getResourceType()`` is used.

inkyant: Oh I hadn't thought about forced revealed resources, I guess they do some sort of workaround? I don't see any calls to CvPlot::ComputeYieldFromAdjacentResource() other than this?

I didn't want to do it in ComputeYieldFromAdjacentResource() because it doesn't have a player parameter,  I could get the owner of the plot but that wouldn't work in case it's an improvement that can be built outside of your territory. So I guess I should add a new parameter then?

IanE9: Yeah, it seems like it only exists for this specifically so adding a parameter is the best solution 👍

RecursiveVision:
Merged into rsi.

PR #8050, opened 2021-06-15
Make NoFollowUp for improvements

inkyant:


Community Patch/Core Files/Core Values/CoreValues.xml
```diff
@@ -315,4 +315,18 @@
...
+		</Update>
+		<Update>
+			<Where Type="IMPROVEMENT_FORT" />
+			<Set NoFollowUp="1" />
+		</Update>
+		<Update>
+			<Where Type="IMPROVEMENT_BARBARIAN_CAMP" />
+			<Set NoFollowUp="1" />
+		</Update>
+	</Improvements>
```
inkyant, line 331: Is this the right place to put this?

RecursiveVision: @inkyant Is MOD_GLOBAL_NO_FOLLOWUP_FROM_CITIES enabled in the Community Patch or the CBP? That should determine where to put this.

inkyant: I think it is part of the PNM mods, which are in the Community Patch, so that's why I put it there. But I couldn't find where it was set to be enabled or even the code for it (not in Core Files\PNM Mods DB\GLOBAL) so I wasn't sure. Additionally this file I put it in is a little strange.

RecursiveVision: Well it's an okay file choice anyway. :)

RecursiveVision:
Merged into rsi.

PR #8060, opened 2021-06-20
Quests now cancelled when City-State killed

Mileslewis:
- This is necessary because you can only have one quest of each type at once. So if a city-state was killed with an active quest it was never getting returned to the pool and you would never get that quest again.


RecursiveVision:
Merged into rsi.

PR #8063, opened 2021-06-20
Add new Trait and Policy stuff, tiny clean up

inkyant:
MonopolyModFlat and MonopolyModPercent to Traits:
works the same as the policy, boosts the yields of monopolies

 Add Policy_YieldForLiberation, InfluenceAllCSFromLiberation, NumUnitsInLiberatedCities, BuildingClassInLiberatedCities:
gives yields/CS influence when you liberate a city, or gives the liberated city units/ a building

Combined some redundant files in Community Patch\Core Files\Core Tables


RecursiveVision:
Merged into rsi.

PR #8064, opened 2021-06-20
Bugfixes 20/06/21

HungryForFood:
- FreePolicy column in PolicyBranchTypes now works correctly with ideology branches.


RecursiveVision:
Merged into rsi.

PR #8085, opened 2021-06-22
Bugfixes 22/06/21

HungryForFood:
- Units granted by Building_FreeUnits now gain starting experience correctly (eg Jet Fighters from Pentagon).


RecursiveVision:
Manual merged

PR #8100, opened 2021-06-27
Add mountain and snow bias

azum4roll:


RecursiveVision:
If we're adding a Mountain start bias, we should assign it to Inca IMO.

PR #8103, opened 2021-06-29
All achievement code now locked behind MOD_API_ACHIEVEMENTS (default 0)

RecursiveVision:
@LoneGazebo This deletes the NO_ACHIEVEMENTS and ACHIEVEMENT_HACKS defines, replaces them with MOD_API_ACHIEVEMENTS in all instances and sets the default value to 0.

The one thing that confuses me is that in CustomMods.h, the syntax was:

```
#if defined(ACHIEVEMENT_HACKS)
#undef(NO_ACHIEVEMENTS)
#else                            
(all of the main defines like MOD_BALANCE_CORE etc. are defined)
#endif
```

I'm unsure if this means there's a compatibility issue between ACHIEVEMENT_HACKS and the main DLL. 

The only case where only ACHIEVEMENT_HACKS (and not MOD_API_ACHIEVEMENTS) defines enabled achievement code was with the UnlockFromDatabase() function in CvAchievementUnlocker.

So you may want to inspect this to make sure all is okay before merging.

### DIFF
--- a/Community Patch/Core Files/PNM Mods DB/API/Achievements.xml
+++ b/Community Patch/Core Files/PNM Mods DB/API/Achievements.xml
@@ -1259,7 +1259,8 @@
 		</Row>
 	</Achievements>
 
+	<!-- Recursive: Default is now 0. Change to 1 to enable all achievements-related code in the DLL. Functionality not guaranteed. -->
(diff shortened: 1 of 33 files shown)


LoneGazebo:
Should be fine, looks good to me!

PR #8124, opened 2021-07-03
Cleanup & sanity check for vassal taxation

RecursiveVision:


RecursiveVision:
@LoneGazebo The large number of changes may look intimidating but it's just copying and pasting functions so they're in different positions in the file for organization.

PR #8131, opened 2021-07-06
Freedom Tweaks

inkyant:
See the [Freedom Tweaks Poll Thread](https://forums.civfanatics.com/threads/poll-freedom-tree-adjustments.671074/)
Urbanization merged into Civil Society
Universal Suffrage now T1
New Deal gets +2 Tourism
Transnationalism gives a Franchise for each Town
Self-Determination gives stuff for liberating cities

BuildingClassInLiberatedCities also changed to a table.


Community Balance Patch/Modular Elements/Corporations/Corporations.xml
```diff
@@ -823,10 +823,6 @@
 			<YieldChange>2</YieldChange>
 		</Row>
 	</Policy_BuildingClassYieldChanges>
-	INSERT="" INTO Policy_BuildingClassYieldChanges
```
inkyant, line 826: so I found this in the middle of the file, was this file broken this entire time? In that case, is it even needed?

LoneGazebo: Possibly? Could be a recent merge issue.

inkyant: I'm no github expert, but I think it says those changes were made 2 years ago? The last time it was updated was 2 months ago so at the very least it's been like that for 2 months.
Why are there two different places for the corporation code anyway? There's `Community-Patch-DLL\Community Balance Patch\Modular Elements\Corporations` and also `Community-Patch-DLL\Corporations\Corporations`

LoneGazebo: second is a standalone project that was aborted by someone else

inkyant: Should we still have the files then? It hasn't been updated for a while anyway, and if needed it can be accessed but for now I think we can delete it.

inkyant:
The poll has closed and all the changes [are here.](https://forums.civfanatics.com/threads/formal-proposal-freedom-tree-updates.671630/)
This PR has all of those changes so it's ready to be merged.

Mileslewis:
The extra franchises from towns won't be counted if you have Nationalization (order tennet) because CvPlayerCorporations::RecalculateNumFranchises() returns before it gets to that part.

Also most things use CountAllImprovement((ImprovementTypes)iI, true) 
Instead of: CountAllImprovement((ImprovementTypes)iI, false) 
as then it only counts improvements you have created yourself rather than conquered ones.

inkyant:
> The extra franchises from towns won't be counted if you have Nationalization (order tennet) because CvPlayerCorporations::RecalculateNumFranchises() returns before it gets to that part.

Right for some reason I thought nationalization didn't count other sources, good catch. 

> I also think you might need to increase CvPlayerCorporations::GetMaxNumFranchises() for each Town as again otherwise it won't work with nationalization.

When I was making it it's intended that it only increases you franchise count, not increasing your maximum. Of course it could do that but that would be a lot stronger. In the description of the max amount of franchises for the player, it says, that some policies can make you have more franchises than your max. Sorry if I'm missing something but I believe it would still work with Nationalization. The max amount of franchises would be equal to your offices, and you would have some above that max from improvements. 

> Also most things use CountAllImprovement((ImprovementTypes)iI, true)
> Instead of: CountAllImprovement((ImprovementTypes)iI, false)
> as then it only counts improvements you have created yourself rather than conquered ones.

I feel like it's more clear to the player to be all the towns that they have, I also specified this on CFC and no one raised an issue so it seemed fine. But if you think its an issue I'll change it.
I'll make a comment on CFC asking people their opinion.

inkyant:
Ok well for now I'll also have it increase the max franchises as that is clearer. We'll see what people think and how it turns out.

Mileslewis:
Wait I just remembered this is a Freedom policy so it doesn't matter how it works with nationalization.

I guess it is still fine like this but sorry for causing stress man lol.

inkyant:
oh no problem lol.
Actually went the opposite for me, when I was first thinking about it I didn't think Order stuff would affect it, bu then you brought it up so I remembered someone suggested using this for an Order tenet and I assumed that's what your were talking about haha.

PR #8188, opened 2021-08-15
Bugfixes 15/08/21

HungryForFood:
- Fixed resources not being revealed correctly when its reveal technology has completed researching.


HungryForFood:
Considering rewriting the resource revealing code to use memory values in CvTeam, surely it is worth the memory rather than iterating through the players in a team every time we want to check the resource reveal status.

PR #8197, opened 2021-08-20
Block non-founder AIs from building Hagia Sophia, Borobudur, St. Basil

RecursiveVision:


RecursiveVision:
(I'm not certain I did this correctly, so @LoneGazebo I'm marking this as a draft for your approval.)

PR #8234, opened 2021-09-03
RSI Fixes & Improvements

RecursiveVision:
Add recent assist bonus for agreeing to a coop war
No positive recent assist decay during the waiting period for a coop war to begin


LoneGazebo:
_I can't believe you've done this_

RecursiveVision:
In fairness, I just got the idea from ridjack on the forums.

It's not mission critical or anything 😉

RecursiveVision:
Considering the population issue it looks like no harm is done anyway 😂

PR #8372, opened 2021-11-01
Game_Era Serialised

Mileslewis:
- Use a serialised m_eGameEra to allow m_eGameEra to be accessed without need for calculation from CvGame::getCurrentEra()
- m_eGameEra changed in CvGame::UpdateGameEra() which is called whenever a major civ reaches a new era.

I've used this in my mod for a while and it works well. It reduces calculations for quite a few calls to CvGame::getCurrentEra() and I used CvGame::getCurrentEra() instead of CvPlayer::GetCurrentEra() in some cases to stop the civs which are already further ahead getting even more bonuses than the rest.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvGame.cpp
+++ b/CvGameCoreDLL_Expansion2/CvGame.cpp
@@ -807,6 +807,7 @@ void CvGame::setInitialItems(CvGameInitialItemsOverrides& kInitialItemOverrides)
 
 	m_iEarliestBarbarianReleaseTurn = getHandicapInfo().getEarliestBarbarianReleaseTurn() + GC.getGame().getJonRandNum(GC.getAI_TACTICAL_BARBARIAN_RELEASE_VARIATION(), "barb release");
 
+	UpdateGameEra();
 	// What route type forms an industrial connection
 	DoUpdateIndustrialRoute();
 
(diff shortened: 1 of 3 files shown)


CvGameCoreDLL_Expansion2/CvGame.cpp
```diff
@@ -4383,26 +4385,32 @@ bool CvGame::canTrainNukes() const
...
+void CvGame::UpdateGameEra()
 {
 	float fEra = 0;
 	int iCount = 0;
-
-	for(int iI = 0; iI < MAX_TEAMS; iI++)
+	for (int iI = 0; iI < MAX_TEAMS; iI++)
 	{
 		if (GET_TEAM((TeamTypes)iI).isAlive() && GET_TEAM((TeamTypes)iI).isMajorCiv())
 		{
```
RecursiveVision, line 4401: @Mileslewis It looks like fEra will always be 0? How does this code work?

RecursiveVision (changes requested):

CvGameCoreDLL_Expansion2/CvGame.cpp
Mileslewis: Yea good spot, must've been a merge error. Fixed now.

RecursiveVision (approved):

Issue #8307, opened 2021-10-06
(AI) Unit Upgrades seem quite low priority


Mileslewis:
_1. Mod version 1.0.3

Seems like there are a lot of reasons why units do not get considered for upgrades which means civs end up with many non-upgraded units even if they enough money.

Only units recruited in CvHomelandAI::RecruitUnits() are considered for an upgrade.

I think it should be higher priority since outdated units can't really achieve much anyways?

PR #8373, opened 2021-11-01
Unit Upgrade Priority

Mileslewis:
- All units now checked to see if they can be upgraded each turn instead of only those recruited by homelandAI

I have used this in my own games and AIs will upgrade their units much faster and I haven't noticed any visible downsides. This could potentially have issues with cancelling units' current orders though I'm not exactly sure how troop movements work.

@ilteroi 
Up to you whether you think this is a good idea or not but could solve #8307


RecursiveVision:
@ilteroi Can you look over this?

RecursiveVision:
@ilteroi ping pong

PR #8383, opened 2021-11-11
Desired vote commitment fix

### LINKED ISSUE #8371: AI offers to vote NAY on my proposal
just an example: i proposed an open door for the city state of sidon, and egypt wants to sell me his NAY. does not make sense.

it seems the code just looks at what the AI desires itself (which is fine) but does not take into account that i made the proposal and therefore would like it to succeed. a fix is probably possible somewhere in CvLeagueAI::GetDesiredVoteCommitments() but i don't know how.

Mileslewis:
- bFlippedLogic removed in CvLeagueAI::GetDesiredVoteCommitments
- human desired commitments based on  'yay' if they proposed it or their vote score with bConsiderGlobal 'false'
Fixes #8371 
- **Interest rate change for deal equilization**

@LoneGazebo You may want to increase the EACH_GOLD_PER_TURN_VALUE_PERCENT to 6 to maintain the same feel of interest rate as before.


RecursiveVision:
@Mileslewis Change to "ready for review" when ready to merge

Issue #8369, opened 2021-10-31
Feature Request for Policy ability: CityCaptureHeal and/or CityCaptureHealGlobal


pineappledan:
I need a new ability added for the proposed redesign of Order's Guerilla Warfare tenet: Units fully healed after a city is captured. 

The heal could either be global or localized to the newly captured city. Global heal is easier to implement, and I think it has the advantage since players won't try to pack as many units into the city range for the heal if it doesn't matter where they are. 

---------------------------
Supporting information:
Forum Thread for new feature:
https://forums.civfanatics.com/threads/poll-rework-of-order%E2%80%99s-geurilla-warfare-tenet.673763/

PR #8395, opened 2021-11-18
Patriotic war update

LessRekkless:
Items left to do:
- Remove Patriotic War Promotion (+15% CS in friendly territory)
- Remove +15% friendly territory CS bonus bonus in Combat Preview (from old Patriotic War); with the update, this is just a graphical glitch
- Add notification for when units are healed? Sweden doesn't get a notification.

#8369


LessRekkless:
Hrm. Looks like I just got rid of the +50% XP bonus in the SQL.  There needs to be another entry that removes the +15% CS in Friendly Territory from the **_BNW_** database.

LessRekkless:
also, the text needs to get updated

PR #8419, opened 2021-11-27
Merges for modmod: Era Restricted Generals

nobubblegums:
Adds a few changes necessary for my upcoming Era Restricted General Bonuses modmod. My first time so hopefully it's in the proper format, tried to imitate what I could from other modmods. Best wishes


nobubblegums (commented):
this is reverted in the second commit to CvPlayerAI.cpp , so you disregard this. didn't know github web didn't have sth like "undo/revert commit", sorry for the mess.

RecursiveVision:
@nobubblegums Where is the code that negates the effect of the GG/GA promotion?

nobubblegums:
Hello, those are done through Lua, I thought modmoders should keep any lua&xml files separate. I vaguely remembered you saying that about ifdefineds :D but with so many people doing it I've erred on the side of caution. I didn't rely on them as true if conditionals, they're practically comments, so should be safe to remove. I've built and tested (a lot), didn't see any problems.

RecursiveVision:
@nobubblegums Okay cool, just wanted to make sure you weren't missing other DLL integration before merging :)

PR #8428, opened 2021-11-29
Update CivText.sql

nobubblegums:
By popular demand Suleiman gets Kanuni instead of Tanzimat. Haven't touched any other lines but git seems to think I've changed 3 of them, maybe an issue with reformatting of unique characters like ü .


RecursiveVision:
Manually merged

PR #8436, opened 2021-11-30
Merges for WHoward's Military Log

nobubblegums:
I don't think these interfere with the actual military/tactical functions and should be safe to merge. That said, this still needs the addition of CvEventLog files to work, which could be too many changes for an integration modmod. I also couldn't figure out a way of making '#include CvEventLog.h' statements conditional on CustomModOptions since we now know #if defined doesn't work that way. Hope that's not a problem.


ryanmusante:
Awesome

ilteroi:
some comments ...

* in general i do not see a problem here, it should work fine
* i do not see any serialization for m_pMilitaryLog though?
* i am slightly concerned about increased filesize for savegames, do you have any idea how much this will add up to?
* you should maybe squash those commits into one

also this coding style with lots of generic default parameters (int iData1 = -1, int iData2 = -1, int iData3 = -1, int iData4 = -1) is hell. i realize you did not invent the thing ... but you ask for a review, you get a review.

nobubblegums:
That's good to know, thanks a lot for your comments! I've pretty much just begun learning c++ and I actually started this merge thinking it'd be a simple copy/paste job, I was almost done with it when realized this was something affecting saves.
I may be horribly wrong here, but I thought removing serialization would also remove any effect on saves - at the cost of losing the military log data between saves, which I thought was a fair compromise. It currently works as I guessed, hopefully it doesn't cause any issues under the hood.
While military logging doesn't seem to work in autoplay, I still did a few 125-turn AI autoplays (pangea, standard), comparing the latest CP release,  my CP build with Military Log disabled in CustomModOptions, and my CP build with MilLog enabled, I found that savegame sizes increased around 10-50kb (560kb to 600kb), seems small enough. My playtested 10ish turns didn't show any differences, but this probably needs more testing to be safe.

ryanmusante:
https://forums.civfanatics.com/threads/ui-military-log.557705/

I remember having issues trying to use this with a version of VP back in 2018.

nobubblegums:
Yeah, that didn't work because the Community Patch DLL didn't integrate necessary functionality from WHoward's DLL, which is actually very rare because almost everything else is integrated. I cannot know for sure, perhaps Gazebo knows, but maybe this issue with the log requiring a new save game format&serialization was the reason. I'm hoping that by removing serialization this provides an acceptable compromise, but the decision to integrate is up to DLL masters. If this gets integrated, it'll still need the mod you linked above for activation & UI functions though.

ilteroi:
ah so you left out the serialization on purpose. also fine, but maybe add a comment to the declaration of m_pMilitaryLog ... if the serialization code in CvEvent is never called the linker will ignore it, so no harm to leave it in.

nobubblegums:
Will do, thanks a lot for your time and patience ilteroi! I've spotted a bug causing logging to not work after re-load (I swear I remember this working), closing the request until I fix this.

gamask:
Is this included in 1.4.6?
It's crashing immediately when I click on the log icon.

nobubblegums:
It's, but requires a different version of WHoward's mod to use https://forums.civfanatics.com/threads/whowards-military-log-civ6like-era-restricted-ggs-gas-generals-spawn-at-the-nearest-city.674314/

gamask:
Thanks! I've been wishing for something like this ever since civ 5 came out!

PR #8459, opened 2021-12-10
Fixed t34 promotions / Shrunk Brazilwood Atlas / Removed unnecessary mipmaps from Community2Art atlas

nobubblegums:
Removed the old entries overwriting new promotion data. Also shrunk Brazilwood Atlas size, it used an 8x8 atlas for just one icon, using a 1x1 atlas instead saves a lot of space (18MB, unbelievable gains)


nobubblegums:
Closing because I hate git

PR #8462, opened 2021-12-14
Fix for "Venetian puppets circumventing building requirements for uni…

nobubblegums:
I think we don't need or won't ever need !bVenetianException check for BuildingClassRequirements, so removing it seems the simplest fix. Tested Venetian&NonVenetian cases with IGE, worked fine.


RecursiveVision:
@nobubblegums Not sure this is a bug? Looks intentional?

nobubblegums:
It's not easy to see this in the screenshot but ridjack explained it a bit more on discord, so the buggy part is that puppets can also make building-locked purchases without having that building, e.g. buying cargo ship or caravel without a harbor constructed. That feels like a bug to me?

RecursiveVision:
Ah yeah, that would be a problem

PR #8465, opened 2021-12-16
Final pass on textures, all standardized & optimized

nobubblegums:
A lot of changes, I know. Here's a list of what's been changed and why:
**Community Patch:**
- Extra GP art files used DXT3 compression, which is DXT1 with more bits for the alpha channel. Yet these images do not use an alpha channel, so there's no point in DXT3, using DXT1 compression instead halves the file size without any quality loss. These alone constitute more than half of all changed files.
- There were also a couple of GreatArt files that weren't compressed, since all other files (including the original game files) are compressed, I've done the same for them.

**Community Balance Overhaul:**
- Shrunk the last oversized atlases. Made a pass on the new unit & improvement textures, some of them used a lossless format, compressed them as above. While size-wise it doesn't make much difference, this could be the most important change here, since these textures can be on-screen on multiple models. Having them lossless causes unnecessary bloat, and as before, the game already uses compressed formats for such files. 
- Splashscreen for CERN was of very low quality, replaced it with a better version of the same image. Also made a minor texture improvement on Hunnic Eki.

**CSD for VP:**
Same as the above, shrunk oversized atlases and icons, compressed splashscreens for wonders&projects, which CSD had a lot. 

Also updated .civ5proj and .modinfo files, including new md5 hashes.
All in all these shrink VP by 100mbs, and I guess the previous commit shaved off 50mbs, so around 150mb total. I know no one cares about a few megabytes, and this may or may not make an impact on vram usage&loading times (maybe people with very low-end rigs can see improvements), but keeping things optimized seems like a better practice to follow with basically no downsides.


RecursiveVision:
150 MB is significant! Thank you!

Issue #7957, opened 2021-05-31
[Feature Request] Open Promotion Tree from Unit Panel


skodkim:
_1. Mod version (i.e Date - 4/14):_

_2. Mod list (if using Vox Populi only, leave blank):_

_3. Error description:_ 
JFD has made a modded version of Unit Panel where you can click on the name of a unit due to be promoted in the unit panel and it opens the Promotion Tree if the mod is activated. Real QOL feature. I've tried to mod it in myself (to the unit panel from VP's EUI) but with no luck - I'm really not good at lua or UI modding. 

I'd like to request this feature in VP's EUI. JFD's mod included for inspiration.
[JFD's Rise to Power - (2) (UI) Unit Panel (v 1).zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/6571797/JFD.s.Rise.to.Power.-.2.UI.Unit.Panel.v.1.zip)

_4. Steps to reproduce (optional):_

---------------------------
Supporting information:
Please note that you can attach .zip files by dragging-and-dropping them. If possible, zip up all supporting data and post that way.

1. Log files (always attach your Logs folder, located at My Documents/My Games/Sid Meier's Civilization 5. For instructions, go to the repository's main page, under "To enable logging for bug reports"):

2. Save game (always attach a save that was made a turn before the error; located at My Documents/My Games/Sid Meier's Civilization 5/ModdedSaves; you can change autosave frequency in the game's Options menu):

3. CvMiniDump.dmp file (attach if experiencing a game crash. Located at Program Files/Steam/steamapps/common/Sid Meier's Civilization V):

4. Screenshots (optional):

skodkim:
Should also note that it seems like really simple code but despite my best tries I end up with no unit panel showing up at all (but then again I have next to no experience with this type of modding so...)

\Skodkim

PR #8507, opened 2022-01-03
Clicking on Unit Promoted opens the promo tree, clicking on promotion…

inkyant:
…s opens their pedia

pressing screenshot removes my mouse but my mouse is on the text that says "Unit Promoted", and left clicking it opens the Promotion Tree if that mod is enabled.
image

Right clicking on a unit's promotions opens the pedia for that promotion


inkyant:
#7957

PR #8518, opened 2022-01-08
RecomputeFollowers() 

LessRekkless:
@ilteroi
RecomputeFollowers() has been added after AddReligiousPressure().  There's no point in recomputing twice in succession.


RecursiveVision:
I removed the extra calls outside of AddReligiousPressure() instead. Fewer lines of code to deal with.

LessRekkless:
@ilteroi specifically moved them outside AddReligiousPressure() on e0b6d886d92975104869d7f0cecc007917f10425 on 2021/12/31.

RecursiveVision:
shrug, don't think he'll mind

RecursiveVision:
Okay I stand corrected

PR #8542, opened 2022-01-15
Some changes for my new difficulty mod + commented questions

Mileslewis:
- HumanPerEraMod and AIPolicyCatchUpMod added
- Comments about unit upgrade cost and AI supply limit
AI Supply Limit: seems to be reduced per era as it's modifier by AIPerEraModifier which is negative
AI upgrade cost: I think it's modified by AIPerEraModifier twice since the base cost of unit upgrades are already modified by AIPerEraModifier. Might be for the best since unit upgrades get expensive late and they really should do them.


nobubblegums:
About AI supply : I'm not seeing any isHuman checks in the supply functions in CvPlayer and they already reduce supply by techs known, so that looks suspicious indeed.

Issue #8558, opened 2022-01-20
Enemy Unit Panel Health bars go wonky with enough bonus text


LessRekkless:
_1. Mod version (X.Y.Z, e.g. 1.3.4):_
1.4 dev (plus experimental comparison)

_3. Error description:_ 
If enough text is added to the modifiers section, the health bars shrink vertically:
2021/10/21 version of EnemyUnitPanel.lua
![image](https://user-images.githubusercontent.com/34615525/150377083-78a9b0b2-032b-44ba-8489-ec3ff8ddca2f.png)
my 2022/01/20 version of EnemyUnitPanel.lua, where I have removed the 100% from the right side of the table, and I think (?) I've reduced the bonusCount to +1 for the movement immunity display. My experimental changes still show the same effect when the miscellaneous bonus appears.
![image](https://user-images.githubusercontent.com/34615525/150376945-4f51f0a3-e6e7-4f66-990b-f55234e079b0.png)
My guess it has something to do with auto-size.

For reference, the unit panel with no bonuses
![image](https://user-images.githubusercontent.com/34615525/150379451-e7a4fa67-b998-403a-9e3f-26170a853c09.png)

_4. Steps to reproduce (optional):_
Force a multi-line bonus, like Status Effect Chance (Boarded), to appear, or get the Miscellaneous bonus to appear.

LessRekkless:
looking at it again, the health bars don't shrink ... the panel itself grows!

LessRekkless:
![image](https://media.discordapp.net/attachments/721543887285387314/934400538664120400/unknown.png)
This screenshot from a user on discord with presumably 1.3.4 that has none of the graphical glitches.

PR #8573, opened 2022-01-24
Ranged Support Fire & EUpanel

LessRekkless:
Main Changes:
Spear Throw Bugfixes
- visually show that units with RangedSupportFire gain no benefit from Fortifying
- units with RangedSupportFire now only spend 1 movement when attacking
- updated the promotion text

Enemy Unit Panel:
- CYAN bonuses now utilize the space available better
- Added Ranged Support Fire calculation to damage prediction, and as a CYAN bonus
- Withdraw Chance includes actual chance to withdraw, taking into account unit speeds and blocked hexes
- excluded damage % from miscellaneous bonuses, as it is multiplicative with the other bonuses (**Suggestions of differentiating this bonus from others encouraged**


LessRekkless:
I would love it if this could be checked on another PC; All of my changes work, but I also have a bugged Panel with no border and it grows unnecessarily (see #8558)

LessRekkless:
Successfully tested? Are my worries with #8558 unfounded and it can be closed?

RecursiveVision:
gonna test

PR #8592, opened 2022-02-01
Major SQL Cleanup

nobubblegums:
--Cleaned up almost all AND EXISTS, except those with actual options, like ResourceReveal, BarbarianHeal and WonderConsolidationTweaks.

--Changed "INSERT INTO SELECT"s to "INSERT INTO VALUES" where appropriate. Values method is faster afaik, most SELECTs were there to add WHERE EXISTS conditional only.

-- Removed everything unit flavor related, since that's handled in a separate file now.

-- DB returns no error, and has the same size, so possibly it's good to go. 

For future:
-- A lot of INSERT INTOs and some UPDATEs can be merged. Ideally, there should be one INSERT command for each table, and it could be better to delete/rebuild tables with a lot of UPDATEs. I'll do that for Unit_FreePromotions next.
-- Corporations.sql changes stuff related to Policy, Techs etc., would be better to merge it to base files.
-- CSD also has modularity leftovers.


nobubblegums:
goddamnit

PR #8652, opened 2022-02-16
Cppcheck

JohnsterID:
Cppcheck as a Github Action

Uses cache to save Cppcheck work folder to reuse the results if the hash for a file is unchanged.

First run would be longest to setup cache. Max configs check set to 2 due to six hour run limit.

More suppression can be added in line. Noting there are false positives and cppcheck devs seem very active as well so it could be reported to them.

Recommended to open xml results in the Cppcheck GUI. The file just needs to placed in your src directory to reference the source when viewing the issues.

[Cppcheck manual](https://github.com/danmar/cppcheck/blob/main/man/manual.md)

[Cppceheck --help dump for all commands](https://github.com/LoneGazebo/Community-Patch-DLL/files/8075639/cppcheck.help.dump.txt)


IanE9:
I support this. More automated testing means less wasted time.

ilteroi:
no harm in this one, but for the record i'll note that the legacy half-c macro-laden mess we have is not the ideal environment for cppcheck

RecursiveVision:
@JohnsterID Any chance you could make it only scan pushes to master? It's doing it for side branches as well, which I suspect is eating into the limited number of scans.

Issue #8675, opened 2022-02-20
Bazooka&Helicopter Anti-Tank Rounds promotion doesn't have a description


nobubblegums:
_1. Mod version, and DirectX version if reporting a graphical issue (X.Y.Z, e.g. 1.4.6, DX 10 & 11):_
1.4.6
Would be good to know about the exact effect
![20220220221932_1](https://user-images.githubusercontent.com/94949778/154860256-19874f0a-6e3c-4b1e-8280-9f2710da5dac.jpg)

I think these are missing associated _help keys

nobubblegums:
OH OK this is actually something that crept in with stackedicons thingy

PR #8677, opened 2022-02-21
'Bonus vs. Armored Units (50)' instead of 'Anti-Tank Rounds'

nobubblegums:
'TXT_KEY_PROMOTION_ANTI_TANK_HELP' doesn't exist in the database, without it Anti-Tank Rounds promotion isn't really descriptive. I tried inserting the help key, but couldn't do it for some reason, so I just changed the title to a format used by many promotions. I don't think this needs a help text anyway. Stackedpromotions also prevent duplicate texts from being shown in the unitpanel when title = help, so it doesn't look bad either.

https://github.com/LoneGazebo/Community-Patch-DLL/issues/8675

### DIFF
--- a/Community Balance Overhaul/Balance Changes/Text/en_US/PromotionText.sql
+++ b/Community Balance Overhaul/Balance Changes/Text/en_US/PromotionText.sql
@@ -238,7 +238,7 @@
 
 -- Adjust Anti-Tank
 	UPDATE Language_en_US
-	SET Text = 'Anti-Tank Rounds.'
+	SET Text = 'Bonus vs. Armored Units (50)'
 	WHERE Tag = 'TXT_KEY_PROMOTION_ANTI_TANK';
 
 	UPDATE Language_en_US


LessRekkless (changes requested):
instead of making this change, instead replace lines 244-246
```
UPDATE Language_en_US
	SET Text = '+50% [ICON_STRENGTH] Combat Strength VS [COLOR_POSITIVE_TEXT]Armored Units[ENDCOLOR].'
	WHERE Tag = 'TXT_KEY_PROMOTION_ANTI_TANK_HELP';
```
with
```
INSERT INTO Language_en_US (Tag, Text)
VALUES ('TXT_KEY_PROMOTION_ANTI_TANK_HELP', '+50% [ICON_STRENGTH] Combat Strength VS [COLOR_POSITIVE_TEXT]Armored Units[ENDCOLOR].')
```

nobubblegums:
Yeah I tried that first, for some reason couldn't get that working. It seems simple enough and I can't figure out what's going wrong, maybe something's off with my local files, so I kinda gave up. Could you do it, I can close this PR in that case

LessRekkless:
The entry in PromotionChanges.xml wasn't pointing to the help text (that didn't exist).  See #8680.

nobubblegums:
Nice, closing this now

Issue #8697, opened 2022-02-27
Rights of Man policy isn't giving food


azum4roll:
_1. Mod version, and DirectX version if reporting a graphical issue (X.Y.Z, e.g. 1.4.6, DX 10 & 11):_
1.4.6

_2. Mod list (if using Vox Populi only, leave blank):_
Full VP + IGE

_3. Error description:_ 
Rights of Man: +1 Food in Cities for every 2 non-Specialist Citizens. +2 Production and Gold from Villages.

It's not giving food as shown in the screenshot.
![image](https://user-images.githubusercontent.com/14953746/155881286-cfe924c5-169e-4105-a91f-52af8c5e8e52.png)

Serfdom, however, is giving culture for non-specialist citizens, despite using the same table as Rights of Man.
![image](https://user-images.githubusercontent.com/14953746/155881310-42040674-541e-4df4-933b-d571e9963b9c.png)

_4. Steps to reproduce (optional):_
Use IGE to give your city citizens and the Rights of Man policy and check the food UI.

---------------------------
Supporting information:
Please note that you can attach .zip files by dragging-and-dropping them. If possible, zip up all supporting data and post that way.

1. Log files (always attach your Logs folder, located at My Documents/My Games/Sid Meier's Civilization 5. For instructions, go to the repository's main page, under "To enable logging for bug reports"):

2. Save game (always attach a save that was made a turn before the error; located at My Documents/My Games/Sid Meier's Civilization 5/ModdedSaves; you can change autosave frequency in the game's Options menu):

3. CvMiniDump.dmp file (attach if experiencing a game crash. Located at Program Files/Steam/steamapps/common/Sid Meier's Civilization V):

4. Screenshots (optional):

maarten99:
For food bonuses the UI looks different than for culture, it gets added to the city tile instead of a "X from Policies".
The city has 11 food, which I'm assuming is 3 base + 1 from granary + 7 from the policy, making it correctly working?

PR #8703, opened 2022-02-28
fix Non-Specialist yields being 100x smaller

inkyant:
The code was copied from getJONSCulture and they left it as divided by 100 but the var is later divided by 100 again..


inkyant:
#8697

PR #8710, opened 2022-03-01
Minor Civ Unit Rebalance

LessRekkless:
Updates for Minor Civ Units:
Turtle Ship : +CS +Indomitable
Great Galleass: +RCS +CS +Targeting I +Move
Pracinha: More GAP on kill
Ballista: -Siege Inaccuracy -Limited Visibility +Field I
Battering Ram: +CS -Defense Penalty +City Assault +Upgrades to Tercio

If the Tercio upgrade is too weird (it goody huts into a Tercio too...), we can go back to Longswordsman, but Battering Ram is still much better at attacking even a Castle City with a Pikeman than a Longswordsman with Drill II and City Assault. It takes less damage too.


LessRekkless:
@nobubblegums Still needs the UNITCOMBAT_HUNNIC_BATTERING_RAM promotions changes to disable Shock.  If you can tell me what line you saw the City State Gift interaction with UNITCOMBAT, I can look into it.

PR #8770, opened 2022-03-30
only first attack gives xp

azum4roll:
Please help me make the changes only applicable to VP and not CP


RecursiveVision:
I think this should be run by the community first.

RecursiveVision:
#8792

PR #8820, opened 2022-04-18
Set workboat prereq resources to 0

nobubblegums:
Set workboat prereq resources to 0 instead of "false". Doesn't change anything, but makes the table look better


RecursiveVision:
The default is "true" in the code and this is not good style.

LessRekkless:
This is the difference between doing something in SQL and XML.  "true"/"false" in XML is automatically converted to 1/0 in the table.

image

LessRekkless:
The only widespread use of 'true'/'false' in SQL is the events table. Everything else is 1/0.

RecursiveVision:
I realized this later but forgot to merge this. Done.

Issue #6580, opened 2020-05-03
city production automation


ilteroi:
there is a mod which allows to automate production for non-puppet cities ([reference](https://forums.civfanatics.com/resources/auto-production-cities.27467/))

it simply sets the "ProductionAutomated" flag for the city in question on a button click in the city screen.

could we do this in VP as well? it's a nice comfort feature and also it allows easier testing of the production AI ...

RecursiveStar:
I second this.

pineappledan:
I wonder if this feature would address a bug i'm having with my Phoenicia custom civ...

ilteroi:
i've made a first attempt to add that feature but it does not work ... can somebody with more lua experience take a look? it's in the "automate" branch.

PR #8829, opened 2022-04-24
Automated production in cities

nobubblegums:
https://github.com/LoneGazebo/Community-Patch-DLL/issues/6580

Implements city production automation. The lua code is taken from the latest EUI, very few lines and apparently multiplayer compatible. Also included are DLL tweaks to get the most out of automation. Automated cities do not produce units, do not make emergency unit purchases, do not build world or national wonders or buildings that consume resources. They also do not build buildings with maintenance costs when running a deficit. As this only influences production, unlike puppets, city focus, citizen and specialist management can still be modified when automated, players can also add their own selections/investments without breaking automation. Automated cities can work projects and processes.
I used (isHuman() && isProductionAutomated()) (with excessive use of parentheses to be safe) to ensure this doesn't modify AI behavior, other implementations I could think about required new states and functions.
Tested with both EUI and nonEUI versions. I didn't include the necessary UI files for Community Patch as that would break other city UI mods. CP players can use Bare Necessities or City Production Automation and still get the benefits of improved automation behavior.


nobubblegums:
Btw the automation tooltip currently says "Automated production for this city." Should it be "in this city"?
I also realized I could've add a check for !isPuppet in addition to the other two, but iirc all related functions are already structured like isPuppet OR isHuman && isAutomated, so that's not needed.

PR #8830, opened 2022-04-25
Automate production in cities

nobubblegums:
https://github.com/LoneGazebo/Community-Patch-DLL/issues/6580

Implements city production automation. The lua code is taken from the latest EUI, has very few lines and apparently multiplayer compatible. Works by simply setting City:isProductionAutomated to true. 
Also included are DLL tweaks to get the most out of automation. Automated cities do not produce units, do not make emergency unit purchases, do not build world or national wonders or buildings that consume resources. They also do not build buildings with maintenance costs when running a deficit. As this only influences production, unlike puppets, city focus, citizen and specialist management can still be modified when automated, players can also add their own selections/investments without breaking automation. Automated cities can work projects and processes.
Works through isHumanAutomated() function which checks (isHuman() && isProductionAutomated() && !isPuppet()) to ensure this doesn't modify AI behavior. As far as I see isHuman isn't cached, while isProductionAutomated and isPuppet is cached, so I decided to not cache isHumanAutomated for simplicity. I can do that if this creates performance issues.

Tested with both EUI and nonEUI versions. I didn't include the necessary UI files for Community Patch as that would break other city UI mods. CP players can use Bare Necessities or City Production Automation and still get the benefits of improved automation behavior.


RecursiveVision:
❤️

Issue #8411, opened 2021-11-24
Promotion choice isn't mandatory (Promotion saving off)/ can't choose promotion after attacking


maarten99:
_1. Mod version (X.Y.Z, e.g. 1.2.1):_
1.2.1

_2. Mod list (if using Vox Populi only, leave blank):_

_3. Error description:_ 
When your unit can choose a promotion you usually have to still choose one after attacking, this is no longer the case.

_4. Steps to reproduce (optional):_
0. Promotion saving off
1. Level unit up
2. Attack before choosing promotion
3. You don't/can't choose a promotion until next turn (can be repeated)

---------------------------
Supporting information:
Please note that you can attach .zip files by dragging-and-dropping them. If possible, zip up all supporting data and post that way.

1. Log files (always attach your Logs folder, located at My Documents/My Games/Sid Meier's Civilization 5. For instructions, go to the repository's main page, under "To enable logging for bug reports"):

2. Save game (always attach a save that was made a turn before the error; located at My Documents/My Games/Sid Meier's Civilization 5/ModdedSaves; you can change autosave frequency in the game's Options menu):

3. CvMiniDump.dmp file (attach if experiencing a game crash. Located at Program Files/Steam/steamapps/common/Sid Meier's Civilization V):

4. Screenshots (optional):
34/30 xp with promotion saving off
![34/30 xp with promotion saving off](https://i.imgur.com/6V6m7cM.jpeg)

PR #8842, opened 2022-04-30
Function Optimizations

LessRekkless:
Spend less time calculating, more time playing!

balance changes:
- Improvements that scale GEMS expend abilities + Transnationalism Franchises no longer need to be within workable distance of an owned city (still needs to be in owned territory and built by the player).
- The Holy Land reformation belief now works for any combination of 2 Holy Sites and Landmarks (instead of requiring exactly 2 Holy Sites or 2 Landmarks).
- Combat Strength PER UnitCombat modifiers from Promotions should now work for all Ranged Units.


LessRekkless:
Should fix #8411

PR #8867, opened 2022-05-09
Start Biases

LessRekkless:
Reorganized start biases and gave justifications.
```
Civilization  Bias
Spain         ~~Any~~ Coast
Mongolia      Grassland/Plains
Shoshone      Plains/Grassland
Huns          Plains/Grassland
Poland        Avoid Jungle/Forest
Rome          Hills
```
Please comment here with any disagreements and why.

### DIFF
--- a/(2) Vox Populi/Balance Changes/Leaders/CivStartBiases.sql
+++ b/(2) Vox Populi/Balance Changes/Leaders/CivStartBiases.sql
@@ -9,11 +9,11 @@ INSERT INTO Civilization_Start_Along_Ocean
 	(CivilizationType, StartAlongOcean)
 VALUES
 	('CIVILIZATION_ENGLAND', '1'),     -- UA + Renaissance UU wants coast (historical)
-	('CIVILIZATION_POLYNESIA', '1'),   -- UA wants coast (historical)
+	('CIVILIZATION_POLYNESIA', '1'),   -- UA wants coast (historical) UI needs Coast
 	('CIVILIZATION_DENMARK', '1'),     -- UA + Classical UU want coast (historical)
 	('CIVILIZATION_CARTHAGE', '1'),    -- UA + Classical UU want coast (historical)
---	('CIVILIZATION_SPAIN', '1'),       -- UA wants coast but is also heavily rewarded for expanding so doesn't need it (NOT historical)
-	('CIVILIZATION_PORTUGAL', '1'),    -- UA + Medieval UU want coast (historical)
+	('CIVILIZATION_SPAIN', '1'),       -- UA wants coast but is also heavily rewarded for expanding so doesn't need it (NOT historical)
+	('CIVILIZATION_PORTUGAL', '1'),    -- UA + Medieval UU want coast, UI needs coast (historical)
 	('CIVILIZATION_VENICE', '1'),      -- UA wants coast (historical)
 	('CIVILIZATION_NETHERLANDS', '1'); -- Renaissance UU wants coast and not rewarded for expanding (historical)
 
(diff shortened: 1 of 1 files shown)


nobubblegums:
A bit hard to read, so Spain loses StartAlongOcean, Mongol gains Plains bias, Poland loses Grass and Plains&gets aversion to Forest and Jungle instead, Huns lose anti-Hills&gain Grass and Plains, Shoshone loses anti-Hills&gains Grass and Plains, Rome loses anti-Jungle&gains Hills.

I like Hill for Rome, why take away aversion to Jungle though, can't it be both? Huns&Mongols also makes sense but @azum4roll said horses spawn less on Plains than Grass, so maybe this should be accompanied by Horse boost to Plains.

LessRekkless:
Please note that this is "Region Selection".  Each Region is designated as one of: 
FOREST | JUNGLE | HILL | PLAINS | GRASSLAND | MOUNTAIN | TUNDRA | DESERT
based on the percentage of the type of . A region can only be one of these 

The tables sort civilizations based on an Order of precedence: 
Ocean -> River -> Priority -> Avoid.  
If a civilization appears in a higher priority, all appearances in lower priorities are ignored.

Therefore, Rome's Hills Priority makes any Avoid Priority pointless.

Unless he states specifically otherwise, I understand that azum4roll is referring to resources that require pastures to upgrade: Cattle and Horses appear Grassland, while only Horses appear in Plains.  This only matters to the Ger, which provides bonuses to Cattle, Sheep, and Horses in addition to the Granary's Wheat, Bananas, Deer, and Bison. That is for the Ger, Grassland (**Cattle**, Horses, Wheat) becomes _as good as_ Plains (**Bison**, Horses, Wheat).

LessRekkless:
So, I spent some time in AssignStartPlots, and azum4roll made some changes in 2020 to differentiate horse placement between Grassland and Plains. For the initial resource pass, Horses have a 1/20 chance of spawning on Grassland, and a 1/32 chance of spawning on Plains. There are other passes that are used to make up deficits as well, which also slightly biases grassland horses over plains horses.

azum4roll:
Rome was set to avoid jungle because it's the only terrain/feature that doesn't have iron. Iron spawns are pretty balanced between other terrains/features.

I don't get why Poland has to avoid trees if it's decided that horses isn't a problem for them.

If Spain can be not on coast I'd say civs with later naval UUs (England, Netherlands) don't have to be either.

LessRekkless:
Rome: that's fair, but I wanted to give it as good a chance as possible to have enough production in order to build its initial army.

Poland needs pastures to build its UB, and it won't find those resources in forest or jungle.

Spain needs only a single city in order to produce its entire Navy, and it's rewarded heavily for expanding aggressively. The same cannot be said for Netherlands or England.

Forest and Jungle really mess with Shoshone's ability to get timely UIs, and since it's not rewarded for conquering, it needs to find horses early or it will never build its UU.

jayeye2011:
> Please note that this is "Region Selection". Each Region is designated as one of: FOREST | JUNGLE | HILL (no feature) | PLAINS (no trees or hill) | GRASSLAND (no trees or hill) | MOUNTAIN | TUNDRA | DESERT based on the percentage of the denoted type of land. A region can only be designated one of these types.
> 
> The tables sort civilizations based on an Order of precedence: Ocean -> River -> Priority -> Avoid. If a civilization appears in a higher priority, all appearances in lower priorities are ignored.
> 
> Therefore, Rome's Hills Priority makes any Avoid Priority pointless.
> 
> > pasture resources appear more on grassland.
> 
> Unless he states specifically otherwise, I understand that azum4roll is referring to resources that require pastures to upgrade: Cattle and Horses appear in Grassland, while only Horses appear in Plains. This only matters to the Ger, which provides bonuses to Cattle, Sheep, and Horses in addition to the Granary's Wheat, Bananas, Deer, and Bison. That is, for the Ger, Grassland becomes _as good as_ Plains because all bonus resources are available. Therefore, the only thing that matters is that they get a higher chances for horses, which means being placed in flat Grassland or Plains.

@LessRekkless  Man thanks so much for this explanation. I like Rome but it more often than not ended up starting in tundra (very un-Rome like IMHO) and to a lesser extent jungle. By only setting avoid priority for Rome to both tundra and jungle and not setting any other priority this fixed this problem for me.  Question for you since you have figured out the code. Does the player go last for start location determination after all the AI civs or does the player civ get first dibs on start location? Also if start biases are set to disabled in advanced setup, does that mean all the civ starts will be evenly distributed across the map?

LessRekkless:
@jayeye2011
> Question for you since you have figured out the code. Does the player go last for start location determination after all the AI civs or does the player civ get first dibs on start location?

For each step in the priorities above, player order is randomized.

> Also if start biases are set to disabled in advanced setup, does that mean all the civ starts will be evenly distributed across the map?

No, it means that all civs are assigned a random region without preference to whether it's coastal/river/forest/etc. The regions themselves won't necessarily be evenly distributed.

jayeye2011:
@LessRekkless , thanks for answering my query.  I feel that a Civ's starting bias should not make low food starts a high probability such that you have to get an early worker just to make 2F farms so you can grow. I wonder if having only hills preference is going to result in this kind of situation for Rome. I would rather just have No Jungle as a starting bias in that case. With the way AssignStartingPlots.lua works now, it's very unlikely that you get iron near your first settle and then only a 2 iron resource usually out at the 3rd ring, even so I would not prioritize making non-resource hill mines, I would just re-roll the map.

LessRekkless:
> With the way AssignStartingPlots.lua works now, it's very unlikely that you get iron near your first settle and then only a 2 iron resource usually out at the 3rd ring

AssignStartingPlots requires that every starting location needs a specific amount of food before it's considered viable, and will add it if it needs to; Hills bias should work out just fine on that front.

jayeye2011:
Not sure start biases are all that important. I used the proposed CivStartBiases.sql and did 10 map starts with Rome, regular Pangea, standard size, default map settings.  Some starts were on a hill or in somewhat hilly terrain some were not. Some starts had what I consider adequate starting food, some did not. It all seems hit or miss to me. Attached is a zip file containing the 10 map rolls for Rome on standard Pangea map script
[10starts.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/8694990/10starts.zip)
.

jayeye2011:
The 10 starts I uploaded before were restarts from within the original start, meaning the same 7 AI Civs were used for each restart. I went back and did 10 scratch starts from the main menu so that each start should have a different set of AI Civs. This seemed marginally better with maybe 2 to 3 sub optimal food starts, hills not always there though
[10scratchstarts.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/8695071/10scratchstarts.zip)
.

LessRekkless:
With a single priority, if there are no regions labeled "HILLS", after signing all of the other priority civ spots, it will go back and found the region with the most of the particular terrain required.

LessRekkless:
You are also not showing the hidden bonus resources, which moderate yield discrepancies between starts.

azum4roll:
They have hills because their very early UU benefits from being on hills.

PR #8882, opened 2022-05-16
Remove default unit name when unit has custom name

1slayerpikmin1:
Currently, if the name of a unit is edited the normal name of the unit still appears in parentheses, such as:

SCREAMING EAGLES (PARATROOPER)

This change makes it appear as:

SCREAMING EAGLES

This is the same change I make every time I download a new version


LessRekkless:
Why, though?

1slayerpikmin1:
It only makes it harder to see the custom name and causes the entire name to become tiny. This also means making the change will allow the custom name to be longer before it starts overflowing off the space it has.
I don’t think anyone has the problem of forgetting what the unit is, so it currently doesn't really do anything

PR #8940, opened 2022-06-13
Germany Rework

pineappledan:
- Adds Influence per turn per gifted unit owned by the CS
 - removes WC votes per 3 CS allies
 - removes GAP per CS friend/ally
 - removes Landsknecht maintenance penalty (major anti-synergy with new UA)
 - moved Landsknecht back to Gunpowder
 - lowered production cost to 275 (was 280, base P&S is 300)
 - Increased faith cost (same as regular P&S, faith cost is determined by tech)
 - fixed all relevant text


(2) Vox Populi/Balance Changes/Units/UnitChanges.sql
```diff
@@ -1,15 +1,5 @@
 -- CIVILIANS AND GREAT PEOPLE
 
-	UPDATE Units SET PurchaseCooldown = 5  WHERE Type IN
```
LessRekkless, line 3: Was removing the purchase cooldown of these civilian units intended?

PR #8941, opened 2022-06-16
Linked & Group movement, DLL functions

nobubblegums:
I won't be able to mod for a while, but fundamentals are there and features are fully functional, only need some checks for edge cases & indicators on the UI. This is the most c++ intensive stuff I've ever done, so I'd appreciate if anyone could take a look at this, to see whether the code could be optimized better. I couldn't manage to use pointers or references for UnitIdContainer, maybe that's one area worth checking.

Linked Movement: Moves all units on the tile together, by calling DoLinkedMovement() on ALT+RCLICK. The function runs two loops, first a loop to get the movable units, their maximum and current movement points. The units are stored in a temporary vector for easier recall on the next loop. The second loop sets the units as IsLinked, and assigns the lowest max&current movement to all linked units. The ordering unit is designated as the LinkedLeader, and gets a standard move mission. The other unit IDs are stored on the Leader unit. The mission calls move() on the Leader, and every move() call made by the mission (i.e. one for each plot on the path) checks whether all the linked units can move. If so, everyone moves, if not, no-one. Upon finishing each move order, ContinueMission is called automatically, if the original move mission is completed, IsGrouped and IsLinkedLeader are reset, and units get back their maxMoves, and their current moves are also updated accordingly.

While plot by plot check is cumbersome, I couldn't think of another way. The problem lies with unit promotions affecting movement costs and the generated path. Ideally all movement promotions of the linked units would be temporarily disabled (unless the promotion is shared by all), but that seems like a costly computation.

Group Movement: Moves all units adjacent to the selected tile, in formation, by calling DoGroupMovement() on ALT+CTRL+RCLICK. The function is very similar to LinkedMovement, but it also loops through adjacent plots (only for the "get units&movement loop"). Another difference is that no unit IDs are stored here, as we just care about the end result, it doesn't matter if unit paths diverge from the formation during movement. Instead of only the leading unit getting a move mission, each unit gets one (non-leading units use approximate_target_ring1 flag).

I thought about creating a new class (i.e. CvUnitGroups), storing UnitIDs, base & group movement points etc. working similar to ArmyAI, but much more limited. That seemed like more work, and I guess would consume more memory. In the current version mostly a few booleans are stored, and units in linkedmovement also store a few integers. Since the latter gets reset after each order, majority of the time it would be empty.

### DIFF
--- a/(2) Vox Populi/Core Files/CoreLua/WorldView.lua
+++ b/(2) Vox Populi/Core Files/CoreLua/WorldView.lua
@@ -116,8 +116,7 @@ function ShowMovementRangeIndicator()
 
 	if bAlt then 
 		print ("displaying linked movement range")
---		Events.ShowMovementRange( iPlayerID, unit:GetSlowestUnitIDOnPlot() );
-		Events.ShowMovementRange( iPlayerID, unit:GetID() );
+		Events.ShowMovementRange( iPlayerID, unit:GetSlowestUnitIDOnPlot() );
 	else
 		Events.ShowMovementRange( iPlayerID, unit:GetID() );
 	end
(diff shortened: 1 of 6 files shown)


nobubblegums:
I'll be back soon,  @LessRekkless @ilteroi , should I go forward with this, any opinions?

nobubblegums:
@RecursiveVision I'm pretty much done with this, I don't know if it'll be savegame compatible though. Should I push it now or wait for 2.6?

Issue #8512, opened 2022-01-03
Triremes in lakes


Neborya:
_1. Mod version (X.Y.Z, e.g. 1.3.4):_
1.3.4

_2. Mod list (if using Vox Populi only, leave blank):_

_3. Error description:_ 
I thought in one of the earlier versions AI was disallowed to build ships in lakes? On the screenshot Genghis Khan has 5 triremes patrolling 8 water tiles (there is no access to the ocean). Is this a bug? Still I think that dromons would be of some use there...

![Civ5Screen0000](https://user-images.githubusercontent.com/27854329/147990136-4525de4f-9ff0-4aa6-84d2-598bdee9c85b.jpg)

ilteroi:
they do build ships here because there is an enemy city.

Neborya:
1 turn before the screenshot Turfan belonged to Mongolia...

ilteroi:
hmm, indeed. should not have happened then

axatin:
Maybe the same issue: It happens sometimes that a lake next to a city is misinterpreted as an ocean coast, and then triremes and lighthouses can be built.
![Screenshot (40)](https://user-images.githubusercontent.com/95587882/154571593-77363db3-cfad-4b7c-82ea-410b50bb39f9.png)

Issue #8935, opened 2022-06-12
Feature Request: Function to see if a city is connected to the ocean via canal


RecursiveVision:
@ilteroi This would solve the "ships spawning in lakes" issue. Disallow ships to spawn in lakes unless they're in the ocean or currently connected to the ocean. Naturally this would have to update in case of pillaging.

LessRekkless:
Here're my thoughts on it:
- If the city is next to a lake, then check the lake's adjacent areas; if any of the adjacent areas are next to water, then check to see if there is a path from the lake to that adjacent plot (by water).
- If yes, then that area is water and there's a canal connection.
- If that area is a lake, then repeat with that area.
- Else, if it's ocean, return true.
- Else, try another area

A simpler method that won't always be foolproof:
- check to see if two cities are connected by water
- Less expensive than the other method, but won't allow building ships on lakes connected to inner seas with no cities on it

This would affect, at minimum, AI production of ships. However, we could extend it to coastal buildings among other things. If we did this, I would say that coastal UIs can't function off of lakes, and coastal buildings require either actual adjacency to a coast tile, or access to the sea

PR #8968, opened 2022-06-25
Prevent Triremes and Lighthouses in Lakes

axatin:
Fix for issue #8512 .

Explanation: It should be "or" instead of "and" in the if-clause: A trireme/lighthouse is buildable if the city is next to water (isCoastal(-1)) **and** the body of water is large enough (isCoastal(pkBuildingInfo->GetMinAreaSize()). The negated condition must use **or** then.

Additionally, for coastal units the check of GetMinAreaSize (10 for all ships, -1 for work boats) was missing, so ships could be build in any city next to water.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvCity.cpp
+++ b/CvGameCoreDLL_Expansion2/CvCity.cpp
@@ -33206,7 +33206,7 @@ bool CvCity::isValidBuildingLocation(BuildingTypes eBuilding) const
 	if (pkBuildingInfo->IsWater())
 	{
 		//-1 is ocean (fast check)
-		if (!isCoastal(-1) && !isCoastal(pkBuildingInfo->GetMinAreaSize()))
+		if (!isCoastal(-1) || !isCoastal(pkBuildingInfo->GetMinAreaSize()))
 			return false;
 	}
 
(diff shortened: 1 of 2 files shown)


LessRekkless:
I don't think this was a bug, instead of a specific choice by previous devs.

Remember that lakes can be connected to seas via canals, and trade ships should be able to access these cities if they are connected. Now, these cities can't produce their own trade ships, but you can still transfer them over to the cities and it functions as before.

LessRekkless:
We also now have a disconnect between what coastal cities mean; namely, there are improvements that need to be next to coast (and this includes lakes).

axatin:
I agree with your comments, didn't think about the canals before. It would probably be better then to discuss the proposed changes in game mechanics in detail first.

When I looked through the code, it seemed to me that there are already different definitions of "coastal" in use: For example, in CvEconomicAI.cpp, line 4123, it says
" if(pLoopCity->isCoastal())  // don't run this for lakes or even small inland seas "
but as you said before, the function isCoastal() checks for **any** water access, not excluding lakes.

RecursiveVision:
I have a solution to this! :D 

https://github.com/LoneGazebo/Community-Patch-DLL/issues/8935

axatin:
In the forum, quite some people complained about this change. Maybe it's best to revert it for now. I will work on a proper implementation following the ideas in #8935 .

LessRekkless:
As we can see from the bug reports, it's exiting with false if it's next to a lake, even if it's also next to ocean. I don't know why `isCoastalLand()` is doing this, but it needs looking at.

PR #9006, opened 2022-07-13
EUI Top Panel Rework

nobubblegums:
Much to unpack here, in short, The top panel had some glaring issues of inconsistency, clutter and poor placement. This does the following:

 - Increased contrast for all text&numbers, and turn counters use a slightly different tone than their yields to help readability.
 - Trade Routes will be colored green when there are unused routes.
 
Consistency improvements:
 - Moved Per Turn yields of Faith and Gold to the left of their respective icons, same as Science, Culture and Tourism.
 - Moved progress bars to the right of the icon for Great Person and Golden Age, same as Science and Culture.
 - Removed parentheses around Military Supply.

 Readability & Clean-up:
Atlas icons vs Font icons: In simple terms, atlas icons are used by Great Person, Tech and Religious unit/buildings and are resized from 45x45 atlases, font icons are almost half the size and used by all other icons on the panel. While atlas icons take more space, they aren't easier to distinguish, and most of them are cropped and/or clip through bars.

 - Replaced atlas icons with font icons for Great Person and Religious unit/progress/building counters. Icon for "Luxury Resources" is replaced with the "Trade" font icon, as it was using the Great Merchant font icon before.
 - Removed the icon for currently researched tech. This information is provided by Science tooltip and the large tech icon just below the top panel, so this icon is redundant.
 
 - Reworked happiness indicators: Overall happiness is indicated by a dynamic happiness (emoji) icon instead of the generic green citizen icon. Icons for total happiness and total unhappiness are removed, instead they are color coded.
 
 Placement: 
 - Moved Gold section to the right of Faith to group main yields. Move Luxury Resources to the left of Trade Routes, on the right side.
 - Moved Great Person and Instant Yields to the right side.
 
I added the changes in a separate "ImprovedTopPanel" folder under 3a for now. In the unlikely case that this causes a huge uproar or major bugs, people can just remove the folder and use the old Top Panel, which I moved to the base EUI folder. I'll merge it properly with 2.6.

Most of the  changes can also bu customized in ImprovedTopPanel/TopPanel.lua . Those are simply booleans & if checks, I don't think they are expensive but I can remove customization too if that's the case.


nobubblegums:
Some screens
Old:
old
New:
new

PR #9038, opened 2022-07-24
Add support for building with clang

IanE9:


CvGameCoreDLL_Expansion2/Lua/CvLuaCity.cpp
```diff
@@ -1671,7 +1671,7 @@ int CvLuaCity::lGetGeneralProductionTurnsLeft(lua_State* L)
 //bool isFoodProduction();
 int CvLuaCity::lIsFoodProduction(lua_State* L)
 {
-	return BasicLuaMethod<bool>(L, &CvCity::isFoodProduction);
```
Hello71, line 1674: you can avoid patching all these calls by putting the new template argument last instead of first

CvGameCoreDLL_Expansion2/CvSpan.h
```diff
@@ -15,7 +15,11 @@
 
 enum DynamicExtentType
 {
+#ifdef __clang__
+	DYNAMIC_EXTENT = INT_MAX,
+#else
 	DYNAMIC_EXTENT = std::size_t(-1),
+#endif // __clang__
```
Hello71, line 22: this wasn't necessary for me, are you sure it is required?

CvLocalization/include/CvLocalization.h
```diff
@@ -288,6 +288,9 @@ namespace Localization
 		CvLocalizationAPI String(const char* szString);
 		CvLocalizationAPI String(const char* szString, unsigned int uiLen);
 		CvLocalizationAPI String(const String& other);
+#ifdef __clang__
+		inline explicit String(const CvString& string) : String(string.data(), string.length()) {}
+#endif // __clang__
```
Hello71, line 293: are you sure this is necessary?

Hello71 (commented):
I am working on a superset of this in my "linux" branch; most of my changes look pretty similar to yours, with some minor differences:

- I used uniform initialization syntax for Localization::String, but if there are no most-vexing-parse issues then parentheses should be fine
- I removed ErrorHandler and FLUA_COMPILE_TIME entirely
- I used meson instead of a custom script
- see above/below for other comments

your work seems a little better than mine though, so I would be happy to rebase my work on top of yours.

CvGameCoreDLL_Expansion2/CvSpan.h
IanE9: I was surprised by this as well. ``clang-cl`` completely refused the code when I tried without this. I'll try it again though to see if there's not some other way.

CvGameCoreDLL_Expansion2/Lua/CvLuaCity.cpp
IanE9: I'll go ahead and change em all 👍

CvLocalization/include/CvLocalization.h
IanE9: Not certain. This was just the easiest way for me to get it building right away. The ifdef is necessary because otherwise the VS2008 compiler errors because of the constructor delegation. If you've got something prettier I'd be happy to change it (it's ugly).

CvGameCoreDLL_Expansion2/CvSpan.h
IanE9: Turns out suppressing ``c++11-narrowing`` permits the original code to build. I'm not entirely sure how much I like that though.

IanE9:
> I am working on a superset of this in my "linux" branch; most of my changes look pretty similar to yours, with some minor differences:
> 
>     * I used uniform initialization syntax for Localization::String, but if there are no most-vexing-parse issues then parentheses should be fine
> 
>     * I removed ErrorHandler and FLUA_COMPILE_TIME entirely
> 
>     * I used meson instead of a custom script
> 
>     * see above/below for other comments
> 
> 
> your work seems a little better than mine though, so I would be happy to rebase my work on top of yours.

I was unaware of your work. I appreciate your feedback here though.

I think using a build tool like meson would be preferable here. I only made the build script because I wasn't entirely sure what the requirements to build would end up being when I started and I just wanted to make sure I'd be able to control the entire environment. Frankly I don't like the build script I've written much at all.

CvLocalization/include/CvLocalization.h
Hello71: I'm fairly sure that you can just delete this constructor. https://github.com/Hello71/Community-Patch-DLL/commit/6837bcb8d10a6718412e438c4ab7661f7a2e67ab is all that I needed. The way that this works on MSVC is that `Localization::String::String(const char *)` exists, so `CvString::operator const char *()` is inferred. That is an MSVC extension though, since it's inferring both the `CvString` -> `const char *` and `const char *` -> `Localization::String` conversions. This can be fixed by adding a single implicit conversion from `CvString` to `Localization::String` or by constructing the `Localization::String` directly with a `CvString` so that the second conversion doesn't need to be inferred.

CvGameCoreDLL_Expansion2/CvSpan.h
Hello71: Ah, I'm using `-w`, which I thought only disabled printing warnings, but now that you mention it I think it might also disable some default `-Werror`. Actually, now that I look closer at this change, I'm not sure that this makes sense. On 64-bit (which I know doesn't actually exist for Civ 5), DynamicExtentType might change size. Since it's not a real enum anyways, this should probably just be `const std::size_t DYNAMIC_EXTENT = -1;`.

Actually actually, https://gcc.godbolt.org/z/qbnY5bjz4 says that `enum { x = std::size_t(-1) }` should work? What error do you get?

CvLocalization/include/CvLocalization.h
IanE9: You're completely right about it being unnecessary. I'm not sure how I managed to miss that.

CvGameCoreDLL_Expansion2/CvSpan.h
IanE9:
```
CvGameCoreDLL_Expansion2/CvSpan.h(21,43) : error: non-type template argument evaluates to -1, which cannot be narrowed to type 'std::size_t' (aka 'unsigned int') [-Wc++11-narrowing]
template<typename T, std::size_t TExtent = DYNAMIC_EXTENT>
                                           ^
CvGameCoreDLL_Expansion2/CvSpan.h(118,16) : error: non-type template argument evaluates to -1, which cannot be narrowed to type 'std::size_t' (aka 'unsigned int') [-Wc++11-narrowing]
class CvSpan<T, DYNAMIC_EXTENT>
                ^
CvGameCoreDLL_Expansion2/CvSpan.h(316,61) : error: non-type template argument evaluates to -1, which cannot be narrowed to type 'std::size_t' (aka 'unsigned int') [-Wc++11-narrowing]
inline CvSpan<typename std::iterator_traits<It>::value_type, DYNAMIC_EXTENT> MakeSpan(It first, std::size_t count)
                                                             ^
CvGameCoreDLL_Expansion2/CvSpan.h(321,67) : error: non-type template argument evaluates to -1, which cannot be narrowed to type 'std::size_t' (aka 'unsigned int') [-Wc++11-narrowing]
inline const CvSpan<typename std::iterator_traits<It>::value_type, DYNAMIC_EXTENT> MakeConstSpan(It first, std::size_t count)
                                                                   ^
CvGameCoreDLL_Expansion2/CvSpan.h(327,61) : error: non-type template argument evaluates to -1, which cannot be narrowed to type 'std::size_t' (aka 'unsigned int') [-Wc++11-narrowing]
inline CvSpan<typename std::iterator_traits<It>::value_type, DYNAMIC_EXTENT> MakeSpan(It first, It last)
                                                             ^
CvGameCoreDLL_Expansion2/CvSpan.h(332,67) : error: non-type template argument evaluates to -1, which cannot be narrowed to type 'std::size_t' (aka 'unsigned int') [-Wc++11-narrowing]
inline const CvSpan<typename std::iterator_traits<It>::value_type, DYNAMIC_EXTENT> MakeConstSpan(It first, It last)
                                                                   ^
CvSpanSerialization.h(29,68) : error: non-type template argument evaluates to -1, which cannot be narrowed to type 'std::size_t' (aka 'unsigned int') [-Wc++11-narrowing]
inline FDataStream& operator<<(FDataStream& saveTo, const CvSpan<T, DYNAMIC_EXTENT>& readFrom)
                                                                    ^
CvSpanSerialization.h(38,70) : error: non-type template argument evaluates to -1, which cannot be narrowed to type 'std::size_t' (aka 'unsigned int') [-Wc++11-narrowing]
inline FDataStream& operator>>(FDataStream& loadFrom, const CvSpan<T, DYNAMIC_EXTENT>& writeTo)
                                                                      ^
```

The error is related to its usage as a template parameter. I also observed this working in other instances which is exactly why it surprised me so much. Maybe it's a quirk/bug related to clang for Windows specifically?

Godbolt accepts the following program under ``x86-64 clang`` no problem but attempting to build it with clang on windows leads to the same category of error being reported.

```cpp
#include <cstdlib>
enum DynamicExtentType {
	DYNAMIC_EXTENT = std::size_t(-1),
};

template<typename T, std::size_t TExtent = DYNAMIC_EXTENT>
struct CvSpan {
    std::size_t m;
    CvSpan() : m(TExtent) {}
};

int main(int argc, char** argv) {
    CvSpan<int> span;
}
```

Hello71: Hm, that *is* weird. I bet it's related to delayed template expansion and/or size_t being unsigned int (on Linux it's unsigned long). Does it work with `const std::size_t DYNAMIC_EXTENT = -1;`?

IanE9: ``const std::size_t DYNAMIC_EXTENT = -1;`` works like a charm. I'd expected that using a constant in a template argument position would have required `constexpr` so I never ended up trying it.

PR #9039, opened 2022-07-24
Update cppcheck.yml

JohnsterID:
run on master branch push only

_Originally posted by @RecursiveVision in https://github.com/LoneGazebo/Community-Patch-DLL/issues/8652#issuecomment-1111443921_: Any chance you could make it only scan pushes to master? It's doing it for side branches as well, which I suspect is eating into the limited number of scans.


RecursiveVision:
Thank you!

PR #9050, opened 2022-07-27
eliminate undefined behavior caused by ICvUnknown::operator delete

### LINKED ISSUE #9047: Usage of ICvUnknown's delete operator is undefined behavior
`ICvUnknown`'s `operator delete` is defined as follows:
```cpp
void DLLCALL operator delete(void* p)
{
	if (p)
	{
		ICvUnknown* inst = (ICvUnknown*)(p);
		inst->Destroy();
	}
}
```
As far as I can tell this was done as a "clever" trick to make `auto_ptr` work with `ICvUnknown` pointers since they are often reference counted. This however invokes undefined behavior in any scenario where `p` is not null. 

Casting `p` to `ICvUnknown*` and then accessing members of that pointer is undefined because there exists no instance of `ICvUnknown` at `p` by the time `operator delete` is entered. You can observe this in action here: https://cpp.godbolt.org/z/8Tc9Keao7.

I believe this is all that's necessary to eliminate the UB:
* Remove `ICvUnknown::operator delete` entirely.
* Expose `ICvUnknown::Destroy` as a public function.
* Replace all instances of ``delete p`` where `p` is of an  `ICvUnknown` derivative type with ``if (p) p->Destroy()``.
* Replace all instances of `auto_ptr<T>` where `T` is an `ICvUnknown` derivative type with some class that implements the same interface as `auto_ptr<T>` but invokes `ICvUnknown::Destroy` at destruction instead of `delete`.

IanE9:
fixes #9047

### DIFF
--- a/CvGameCoreDLLUtil/include/CvDllInterfaces.h
+++ b/CvGameCoreDLLUtil/include/CvDllInterfaces.h
@@ -305,15 +305,6 @@ class ICvUnknown
 public:
 	static GUID DLLCALL GetInterfaceId(){ return guidICvUnknown; }
 
-	void DLLCALL operator delete(void* p)
-	{
-		if (p)
-		{
-			ICvUnknown* inst = (ICvUnknown*)(p);
-			inst->Destroy();
-		}
-	}
-
 	virtual void* DLLCALL QueryInterface(GUID guidInterface) = 0;
 
  	template<typename T>
@@ -322,7 +313,10 @@ public:
  		return static_cast<T*>(QueryInterface(T::GetInterfaceId()));
  	}
 
-protected:
+	// Formerly protected.
+	// 
+	// This logically needs to be part of the public interface. Hopefully this
+	// change doesn't cause ABI compatibility issues.
 	virtual void DLLCALL Destroy() = 0;
 };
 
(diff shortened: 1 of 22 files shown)


RecursiveVision:
@IanE9 Are you sure this doesn't cause any crashes, etc ingame? I'm wary of modifying any DLLCALL functions.

IanE9:
> @IanE9 Are you sure this doesn't cause any crashes, etc ingame? I'm wary of modifying any DLLCALL functions.

In the limited testing I've done I've experienced no issues. Which function specifically are you worried about modifying though? If it's just `ICvUnknown::operator delete` you're worried about, it should be safe since it's neither imported nor exported.

These changes should also result in basically the same binary being generated. I'll go ahead and examine the generated code to ensure this is the case though and get back to you.

IanE9:
@RecursiveVision VC90 generates slightly different code in some instances. It looks like this may be due to the compiler missing the opportunity to inline `Wrap*Pointer` functions and this change having changed their signature slightly. Clang however doesn't miss the inlining opportunity and nearly identical code is generated in all instances I checked. The only differences I observed were address changes.

I also re-reviewed the usage of every `CvInterfacePtr` instance in the source code to try and make sure I hadn't missed anything. I couldn't find anything that looks like it would cause breakage, but I went ahead and tried to make sure the `operator bool` implementation was being used whenever possible for consistency.

I think the only thing here that _could_ be a problem is the lifting of `ICvUnknown::Destroy` from `protected` to `public`. Since C++ doesn't maintain a stable ABI, the compiler is free to change the layout of `ICvUnknown`'s vtable due to this change alone but in practice it looks like the only thing that influences the vtable layout under Microsoft's ABI is the order that the functions are declared in and the names of the functions in the case that a function is overloaded. In short I don't think the change will cause issues.

If you'd like a second opinion, maybe @Hello71 would like to review this since they seem pretty familiar with the C++ standard and might spot something I've missed.

Hello71:
> @IanE9 Are you sure this doesn't cause any crashes, etc ingame? I'm wary of modifying any DLLCALL functions.

`DLLCALL` is `__stdcall`, not `__declspec(dllexport)`. Virtual functions are "implicitly exported" via the virtual tables, but this `operator delete` is neither `__declspec(dllexport)` nor virtual and its address is clearly not taken because deleting it doesn't cause any errors.

> Since C++ doesn't maintain a stable ABI, the compiler is free to change the layout of ICvUnknown's vtable due to this change

Although C++ itself doesn't say anything about ABI, each platform does. Microsoft says very little about its ABI, except that it is stable within a given compiler version. Assuming you've checked that the vtable is the same after this change, then it must remain so, even if the rest of the code changes, or if the optimization level changes, etc.

In general, I think this change is *probably* fine? I'm not totally convinced that this is the *best* way to do it, but I don't see any massive issues. With C++11 I think the best way to do it would be `inline void deleteICvUnknown(ICvUnknown *ptr) { if (ptr) ptr->Destroy(); } template<class T> using CvInterfacePtr<T> = unique_ptr<T, decltype(&deleteICvUnknown)>;`.

Issue #9028, opened 2022-07-21
[OCC] Turn 67 CTD 


maarten99:
_1. Mod version, and DirectX version if reporting a graphical issue (X.Y.Z, e.g. 2.5.0, DX 10 & 11):_
2.5

_2. Mod list (if using Vox Populi only, leave blank):_
Alternative difficulty v13

_3. Error description:_ 
Crashes to desktop during city state's turn, right before my turn starts.

_4. Steps to reproduce (optional):_

Press next turn in save:  [crashKamehameha_0067 BC-1320.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/9162637/crashKamehameha_0067.BC-1320.zip)
(Accept Sweden's friend request, not sure if that matters)

Autosaves: [auto.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/9162649/auto.zip)

axatin:
This is caused by bug #8850 (here with the City State of Bogota)

Issue #8850, opened 2022-05-02
City State City Capture Quest bugged


Mileslewis:
Version 2.1.1

I keep getting quests to capture a city with no city specified. The quest then cancels next turn.

Game is OCC so might be to do with that?

maarten99:
I have had this same bug in the last 2 years, not always/ very randomly when it occurred. But when I did get it in a game, the quest never gave me a valid city target.
Clearing cache seems to fix this bug (before starting new game)

maarten99:
I started an OCC game while clearing cache first, but I still had the bug. That might have just been my superstition, I also believe it to be an OCC bug now. Here is how it looks in game:

![image](https://user-images.githubusercontent.com/24713571/173345628-b32ba0a4-e52d-42e8-97f1-f467af795730.png)

axatin:
It is an OCC bug. It was intended that this quest is not assigned to human players at all if they can't acquire any cities.

maarten99:
That would be an awful restriction to OCC, imo the quest should be changed to Capture instead of Acquire the city.

axatin:
The "Acquite City" quest is not fulfillable in OCC as captured cities are instantly destroyed, so it has to be excluded. When the bug is fixed, the city state will always give another quest to the player instead, so I don't see it as a big restriction.

The same bug can also happen with the quest "Liberate another City-State". I think this quest should be available in OCC since city liberation is possible.

axatin:
I don't have any experience with OCC games myself. I can imagine that when you have only one city it's very difficult to capture other cities, and maybe that's why the quest to liberate cities was also excluded? What do you think, should the quest be offered in OCC games?

maarten99:
The quest should definitely be offered to OCC players.

While it is true that capturing cities is more difficult when you're constrained by a low supply cap like in OCC, it is the most powerful quest reward and not offering it to OCC players would be an unreasonable nerf.

Even though you are disadvantaged in OCC, the instant raze is also very useful for civs like France, because they have bonuses on city capture. These bonuses together with instant raze can even be used to win the game more quickly than in non-OCC (I won a culture victory on turn ~240 with France OCC like this)

axatin:
okay, I'm working on it. I will create a new quest for OCC games only with the goal to destroy a city and with the same rewards as the other quest.

LessRekkless:
[...]

PR #9052, opened 2022-07-27
Add conquer city quest for OCC games

axatin:
Fix for #8850 and #9028 

- The bugged CS quest to acquire a city is turned off for OCC games
- A new quest to destroy a city is added for OCC games instead
- The liberation quest can now also be given in OCC games


IanE9:
Instead of adding a new quest type specifically for OOC games would it not be more appropriate to instead make `MINOR_CIV_QUEST_UNIT_GET_CITY` achievable by destroying the target city as you've done with the new quest type? Apologies if this has already been discussed somewhere and I've missed it.

CvGameCoreDLL_Expansion2/CvMinorCivAI.cpp
```diff
@@ -2824,6 +2854,33 @@ void CvMinorCivQuest::DoStartQuest(int iStartTurn, PlayerTypes pCallingPlayer)
...
+
+	const char* strTargetNameKey = pCity->getNameKey();
+
+	strMessage = Localization::Lookup("TXT_KEY_NOTIFICATION_QUEST_UNIT_GET_CITY_OCC");
+	strMessage << strTargetNameKey;
+	strSummary = Localization::Lookup("TXT_KEY_NOTIFICATION_SUMMARY_UNIT_GET_CITY_OCC");
+	strSummary << strTargetNameKey;
+
+	iNotificationX = pCity->plot()->getX();
+	iNotificationY = pCity->plot()->getY();
```
IanE9, line 2882: Loose indentation here. Empty line at the beginning of the else if body should also be removed.

CvGameCoreDLL_Expansion2/CvMinorCivAI.cpp
```diff
@@ -3424,6 +3481,11 @@ bool CvMinorCivQuest::DoFinishQuest()
 		strMessage = Localization::Lookup("TXT_KEY_NOTIFICATION_QUEST_GET_CITY_COMPLETE");
 		strSummary = Localization::Lookup("TXT_KEY_NOTIFICATION_SUMMARY_GET_CITY_COMPLETE");
 	}
+	else if (m_eType == MINOR_CIV_QUEST_UNIT_GET_CITY_OCC)
+	{
+	strMessage = Localization::Lookup("TXT_KEY_NOTIFICATION_QUEST_GET_CITY_OCC_COMPLETE");
+	strSummary = Localization::Lookup("TXT_KEY_NOTIFICATION_SUMMARY_GET_CITY_OCC_COMPLETE");
```
IanE9, line 3487: Loose indentation here.

IanE9 (changes requested):

RecursiveVision:
This would be more effectively accomplished by altering how the quest works for humans in OCC games. The text is fine, but adding a new quest is clunky.

PR #9062, opened 2022-07-30
Fix warnings reported by Clang

IanE9:
This fixes most of the warnings reported by Clang. Some of these changes may fix bugs but mostly they just make things far more compliant with the C++ standard.

`delete-abstract-non-virtual-dtor` warnings are fixed by #9050 so I haven't included them here.

### Blocking issues
* #9081
* #9082
* ~~#9086~~ - Too little info to fix right now

### Todo
* [x] Fix `microsoft-template-shadow` warnings.
* [x] Fix enum related switch `warnings`.  Fixing `tautological-constant-out-of-range-compare` warnings is not currently feasible.
* [x] Rename `Flua` directory to `FLua`.
* [x] Ensure changes haven't broken things.

### WontFix (at present)
* `invalid-offsetof` warnings - They're related to multiplayer functionality. Since multiplayer has multiple issues that make it nearly unplayable, they should be addressed if we ever get around to fixing it overall.


Hello71:
LGTM overall. I think I would prefer moving Flua to FLua instead of changing the includes though, as FLua seems to be the "proper name".

Hello71:
> - Can enum related warnings feasibly be fixed? Firaxis' usage of enums is bonkers.

This is probably a pre-C++ habit: in C, enum values can be used in certain contexts that const ints cannot, such as array dimensions, switch statements, and static variables: https://stackoverflow.com/questions/1674032/static-const-vs-define-vs-enum/1674459.

IanE9:
> LGTM overall. I think I would prefer moving Flua to FLua instead of changing the includes though, as FLua seems to be the "proper name".

I think you're right. I'll change this when I get a chance.

> > * Can enum related warnings feasibly be fixed? Firaxis' usage of enums is bonkers.
> 
> This is probably a pre-C++ habit: in C, enum values can be used in certain contexts that const ints cannot, such as array dimensions, switch statements, and static variables: https://stackoverflow.com/questions/1674032/static-const-vs-define-vs-enum/1674459.

This seems likely.

I'd like to at least be able to enable the `switch` warnings because they've at least helped me catch a bug where lua scripts were not receiving the correct modifier for puppet city golden age point yield penalties. Getting to the point where they'd be able to be enabled by default though is pretty tough because the enums are used so inconsistently throughout the codebase. For instance the `NO_*` discriminants are sometimes meant as an invalid state or lower bound while other times they're just a normal state that switches should always have a branch for.

I've also had to suppress `tautological-constant-out-of-range-compare` by default because although Clang documents an `enum_extensibility` attribute that should logically suppress these warnings where applicable (`PlayerTypes`, `TeamTypes`, etc), the attribute instead has no affect on the warnings.

Then there are enum definitions that are mixed in behavior. `FlowDirectionTypes` for instance mixes regular discriminant values with bit masking values and then those bit masks do not actually correlate with the values they're expected to mask such as `FLOWDIRECTION_NORTHEAST` being `1` but `FLOWDIRECTION_NORTHEAST_MASK` corresponding to the _sixth_ least significant bit. 

There are also several enums that also mix values that increment regularly with values that correspond to some string CRC hash which is incredibly bizarre.

JohnsterID:
Out of interest, after your hard work, would [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) work or would there just be too much to suppress noting the requirement to maintain vc90 compatibility?

IanE9:
> Out of interest, after your hard work, would [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) work or would there just be too much to suppress noting the requirement to maintain vc90 compatibility?

I don't think there would be much stopping us from adding this now, although I don't think we'll know what problems might present themselves until someone tries. Right now the only warning categories I think I'll end up suppressing by default are `invalid-offsetof`, `tautological-constant-out-of-range-compare`, and `comment` although `comment` could probably be fixed if we really cared enough.

If you'd like to try and add this I'd appreciate it, there are still at least 100 switch warnings I need to go through before I'll be comfortable with this pull request being reviewed and merged and even before then I'll need some time to make sure that I haven't critically broken anything. I'll probably be working on this for a little while longer because of that. If you need any help at all though I'd be happy to provide it. I'd even be supportive of adding clang-format just to get the style of the code consistent, but we'd probably want to wait until all outstanding work is submitted because it would touch a lot of the code and create a lot of manual merging work for anyone with incomplete changes.

clang.cpp
```diff
@@ -1,5 +1,37 @@
...
+        );
+    }
+    /// Convert an int64 with the low 32-bits in `ecx` and the high 32-bits
+    /// in `edx` to a double precision floating point output as `xmm0`.
+    __attribute__((naked)) void _ltod3() {
+        // Clang 14 generated SSE3 dependent implementation.
+        __asm__(".intel_syntax\n"
+            "sub     esp, 20\n"
+            "movd    xmm0, edx\n"
+            "movd    xmm1, ecx\n"
```
Hello71, line 27: are you sure _ltod3 is allowed to clobber xmm1? also, are you sure this can't be fixed by linking against msvcrt or somesuch?

CvGameCoreDLL_Expansion2/Lua/CvLuaPlayer.cpp
```diff
@@ -1473,6 +1473,7 @@ static DomainTypes LuaToTradeDomain(lua_State* L, int index)
 		return eDomain;
 	default:
 		luaL_error(L, "Invalid trade domain index %d", iDomain);
+		UNREACHABLE();
```
Hello71, line 1476: mark luaL_error noreturn?

clang.cpp
IanE9: These functions are normally provided by the VS2022 CRT. I'm not sure that linking against multiple CRTs is feasible. As for clobbering `xmm1`, Microsoft's implementation already does this along the branch where AVX instructions are unavailable so it's presumably safe to do so.

CvGameCoreDLL_Expansion2/Lua/CvLuaPlayer.cpp
IanE9: Looks like these functions are documented as never returning so I see no harm in doing so. I'll change the function definitions as appropriate.

JohnsterID:
> If you'd like to try and add this I'd appreciate it, there are still at least 100 switch warnings I need to go through before I'll be comfortable with this pull request being reviewed and merged and even before then I'll need some time to make sure that I haven't critically broken anything. I'll probably be working on this for a little while longer because of that. If you need any help at all though I'd be happy to provide it.

I've been away with work, if I had time I would love to attempt it. I'm only thinking out load, CMake for both clang and vc90. The first option would enable CI and the latter to test compliance. CI with CMake would enable clang-tidy and future clang-format on commits. Thanks for all your effort btw. Good luck going forward! Also, if I'm reading this [Stack Overflow](https://stackoverflow.com/questions/70095235/asan-suppressing-reports-in-external-libraries-llvm-13-windows) and linked MS blog correctly, looks like it might be possible can run asan on the dll ignoring false positives from the non-asan runtime.

JohnsterID:
Have you encountered the below, is it invalid?
1>CvTacticalAI.cpp(9283,9): error : no matching function for call to 'make_pair'
https://github.com/IanE9/Community-Patch-DLL/blob/f290a4b06557d04059000293e321b67ca5fb158b/CvGameCoreDLL_Expansion2/CvTacticalAI.cpp#L9283

JohnsterID:
Valid error?
CvTacticalAI.cpp(9283,9): error : no matching function for call to 'make_pair'
message : candidate function template not viable: expects an rvalue for 2nd argument
https://github.com/IanE9/Community-Patch-DLL/blob/clang-warnings/CvGameCoreDLL_Expansion2/CvTacticalAI.cpp#L9283

IanE9:
> Valid error? CvTacticalAI.cpp(9283,9): error : no matching function for call to 'make_pair' message : candidate function template not viable: expects an rvalue for 2nd argument https://github.com/IanE9/Community-Patch-DLL/blob/clang-warnings/CvGameCoreDLL_Expansion2/CvTacticalAI.cpp#L9283

What's the compiler setup? I've been building from the current branch without any errors.

JohnsterID:
> What's the compiler setup? I've been building from the current branch without any errors.

Just attempted clang-cl on Visual Studio 2019 and 2022. I turned off all warnings. v90 works fine.

IanE9:
> > What's the compiler setup? I've been building from the current branch without any errors.
> 
> Just attempted clang-cl on Visual Studio 2019 and 2022. I turned off all warnings. v90 works fine.

clang-cl probably needs to be configured to use VC90's libraries and headers. The build script I wrote does this by running `vsvars32.bat` from the directory specified by the `VS90COMNTOOLS` environment variable.

https://github.com/IanE9/Community-Patch-DLL/blob/clang-warnings/build_vp_clang.py#L14

JohnsterID:
> clang-cl probably needs to be configured to use VC90's libraries and headers. The build script I wrote does this by running `vsvars32.bat` from the directory specified by the `VS90COMNTOOLS` environment variable.

That would be it. Thanks.

PR #9067, opened 2022-07-31
New Farm Resources

nobubblegums:
Maize and Rice are added as alternative farm resources. Farm textures are generated according to the continent they are placed, and there's one specific texture for wheat farms. It's not possible to access how farm textures&models are handled and add in new resources that way, so old civfanatics modders worked around that by hijacking some generic continental farm textures and replacing them with new resource textures. I basically replicated that method in the DLL, so it isn't necessary to use custom events or save utils. Anyway, rice and maize textures replace Asian and American farms. This is done by overriding the farm_types.xml, and then changing the continent art just before the farm is placed in setImprovementType(). To prevent generic Asian&American farms from having rice or maize farm textures, generic Asian&American farms are turned into European&African farms, again in setImprovementType(). The difference is negligible and not a real loss.

I'm using whatever HungryForFood used for placement, and modified AssignStartPlots and Communitu accordingly. Placement seems good to me, rice spawns near rivers, and I think maize has more plainsy bias. Both are exactly the same with Wheat in any other regards (yields etc.), and since they are improved by the default farm improvement, there won't be any problems with beliefs, adjacency bonuses etc. If people think it'd be good to differentiate them to balance different starts, go for it.

The total farm resource number may be too high by now, although I didn't notice this in my tests. On a standard map each resource had single digits, and the map wasn't crowded with resources. I'm leaving the tuning&balance to @azum4roll @LessRekkless @pineappledan 

LUA code taken from Even More Resources, and art assets were taken from Even More Resources by HungryForFood  and Amer-Indian Resources by framedarchitect, modified by me where needed. They probably weren't the original artists, I'll try to reach those for permission, but it's pretty fair game by now.

Here's a test build https://drive.google.com/file/d/1QzSIT0nwS04wFeQbOyY1JzoMUuEXnWg9/view?usp=sharing


LessRekkless:
Note that Wheat is currently restricted to plains only.

RecursiveVision:
Rice in plains doesn't make a lot of sense.

nobubblegums:
@RecursiveVision yeah I guess that was a leftover from Amer-Indian resources, I'm removing that. So rice will spawn only in grassland rivers, marshes or floodplains. Harder to differentiate maize and wheat, I think maize could be biases for hot&wet climate.

azum4roll:
I don't think MaxLatitude is being used anywhere?

PR #9078, opened 2022-08-05
Check for null `pPlot` during `MINOR_CIV_QUEST_UNIT_GET_CITY` completion check

IanE9:
A user on discord was crashing because `pPlot` was null at the end of a turn. I don't know why but the plot coordinates were `(-1, -1)`. There's probably an oversight elsewhere that causes a bad plot to be used, but the null check was removed as part of https://github.com/LoneGazebo/Community-Patch-DLL/commit/4cd3032f501f5b594bc5c6273fbcce16e6866fbd so I see little harm in putting it back for now.


RecursiveVision:
Issue is already fixed!

PR #9085, opened 2022-08-07
Add new bonus resource files to mod project

IanE9:
They aren't packaged unless we add them to the project file. If they aren't included the map script crashes when generating a map.


RecursiveVision:
@nobubblegums Take note

nobubblegums:
Missed that one, thanks for pointing it out

PR #9088, opened 2022-08-07
Failsafe: Remove new farm resources if no agreement can be reached

nobubblegums:
It doesn't seem like people can agree on balance, if that doesn't happen until the release this should revert all the relevant changes.


LessRekkless:
There just needs to be a more complete pass in AssignStartingPlots to make it worth like the other bonus resources. @azum4roll ?

Balance concerns can play it after the release.

Otherwise, balance

nobubblegums:
I don't know if it helps or if it's possible but here are some things I noticed. This was with Communitu, will check with default maps too.
- Both maize and rice can spawn near desert areas. It's much harder for rice, but I've seen it happen when a grassland is adjacent to a oasis. For maize, it seems like when a river goes through a desert area, some tiles adjacent to river gets turned into plains, and maize can spawn on them. I think having a "no adjacent desert or oasis" condition both on maize and rice would make sense to differentiate placement. Actually, we can also add "no adjacent tundra" to be extra safe. Similarly, wheat can have a "no jungle adjacent" condition. 
- Bananas compete with both maize and rice, their spawn rate can be reduced a bit.

azum4roll:
see #9091

azum4roll:
Strategic View doesn't break... for now

azum4roll:
Adjacent plot checking won't work well, since some terrain adjustments happen after resource placement.

PR #9091, opened 2022-08-08
balance wheat, rice and maize spawns

azum4roll:
Communitu_79a v2.4.0
also remove rice's help text


nobubblegums:
Looks good, this what I've got with Continents on Huge
```
[863511.578] Map Script: Wheat list size:	434
[863511.578] Map Script: Rice list size:	101
[863511.578] Map Script: Maize list size:	693
[863517.656] Map Script: 6	Wheat...: 	24
[863517.656] Map Script: 55	Maize...: 	29
[863517.656] Map Script: 54	Rice....: 	9

```
I think making maize on plains freshwater only could be nice to differentiate from wheat, but not really that important.

azum4roll:
I just changed the tropical threshold to 0.15. The function is only used for determining which tree to place in AdjustTiles otherwise, and the threshold has always been inconsistent with the one in FeatureGenerator anyway.

nobubblegums:
Sure, we can just change the numbers if the function isn't used anywhere else.

PR #9096, opened 2022-08-09
Swordline change triggers for custom units&civs

nobubblegums:
https://github.com/LoneGazebo/Community-Patch-DLL/commit/576630d281e1ddad9d60957bd329a245a908c966

These triggers should bring any custom civs in line with the latest unit changes. That said, VP doesn't really use triggers much, exceptions are HappinessCompatibility.sql, LeftOverUnits.sql for Unit_BuildingClassPurchaseRequireds, and some triggers for adding custom resources to Hacienda, but they aren't "core" changes like Unit Combat Strength or Promotions. Then again, it's very late in VPs development, so burden of maintaining compatibility can be moved from modmods to VP in some cases. 
I don't know if triggers have any performance effect when they are not hit, I placed the file at the end so it shouldn't do anything for a pure VP installation.

### DIFF
--- a/(2) Vox Populi/Balance Changes/Units/UnitTriggers.sql
+++ b/(2) Vox Populi/Balance Changes/Units/UnitTriggers.sql
@@ -1,0 +1,23 @@
+CREATE TRIGGER VP_SwordsmanPromotion_Replace
+AFTER INSERT ON Units
+WHEN (SELECT Class FROM Units WHERE Type = NEW.UnitType) = 'UNITCLASS_SWORDSMAN'
+BEGIN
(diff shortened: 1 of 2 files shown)


RecursiveVision:
Ehhhh. I dunno about this one. Modmodders can make this change themselves.

Consider also that anyone trying to make it compatible by adding Professionalism manually rather than via trigger would create a list entry duplication error.

RecursiveVision:
Also, I don't like the automatic balance change with the Tercio line and removal of Cover, because it interferes with modder freedom.

Vetoing this one, this is the responsibility of modmodders.

nobubblegums:
shaq

PR #9103, opened 2022-08-12
Minor Songhai nerf

nobubblegums:
Simply removes War Canoes promotion from Songhai land units, which gave them embarked sight and defense. I&People also wanted to take out the "attack from sea without penalty", but that would require splitting the Amphibious promotion, which would create some other annoyances.


LessRekkless:
It wouldn't have been difficult to just make War Canoes have the `River` part of Amphibious, but without another bonus, a strictly worse version of Amphibious isn't going to fly.

PR #9115, opened 2022-08-16
Remove "Influence" from G.Merchant trade mission text

nobubblegums:
I don't know of any GMerchant variant that gives influence, didn't get any answers either when asked on discord. Diplo units (including GD) already display another text, so there should be no need for this.


IanE9:
If any DLL code or Lua code expects the string to be formatted with 2 values, we should change those locations to not send 2 format values as well. If it's possible for a mod to add influence to a great merchant we should just introduce a second string that doesn't include the useless influence bit as you've done here and use that one in the case that the influence gain is `0`.

RecursiveVision:
Yeah, this code won't work if the DLL expects 2 values to be passed to the string

nobubblegums:
It works with base VP but you're right, could cause some issues. Will work on it.

PR #9123, opened 2022-08-18
Cherry-pick various commits from Linux port

IanE9:
This brings the main repo into better sync with @Hello71's Linux port by bringing over non-breaking changes.

Review and feedback from @Hello71 would be appreciated before any merge takes place. Additionally, if it would be better/preferable for the changes to be independently brought over by them, I can close this and we can wait.


JohnsterID:
clang-tidy check, eventually should just add it as a auto review action on PRs. Plenty of options on Github Marketplace.
[ErrorList_2ef71da-html.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/9383191/ErrorList_2ef71da-html.zip)
[ErrorList_2ef71da.xlsx](https://github.com/LoneGazebo/Community-Patch-DLL/files/9383198/ErrorList_2ef71da.xlsx)

Hello71:
> The std:: thing is annoying - is there no other way to make it compatible?

I'm kinda waffling on that: the specific issue this is solving is that in C++17 (which is not required for the Linux port), std::byte conflicts with FFireTypes byte.

It might be possible to fix this by removing the FFireTypes byte in C++17+, but will require changing lots of instances of byte to unsigned char, because std::byte can't be used for arithmetic. In my opinion, this is a bad solution, because these types of conflicts will surely come up again with future C++ standards.

Another way to fix it would be to selectively `using std::min; using std::max; using std::auto_ptr;` etc. This is a far better solution than selectively fixing byte, but in my opinion is an inferior solution to removing these using declarations entirely, because:

1. Much of the code already uses `std::`. Always using `std::` removes the question of why sometimes it is used and sometimes it is not.
2. Some of the required `using`s could conflict with future methods; nobody will define a new `min` function, but someone might define a `hash` function which accidentally conflicts with `std::hash`, or a `map` variable which conflicts with `std::map`.

The first problem could be solved by consistently removing `std::` from the relevant usages. The second problem could be solved by restricting the set of `using`s to the bare minimum, and patching out the remainder. I can't remember exactly, but I think "bare minimum" would probably consist of min, max, auto_ptr, vector, and maybe pair and map.

Hello71:
Broadly speaking, I haven't submitted these commits yet because I want to go through and add explanations, as well as sort which ones are required for gcc, clang, and non-Windows. There are also a couple of dubious ones, like "Remove using namespace std", the various unity build commits, and the FLua ErrorHandler/COMPILE_TIME_CONDITION commits, as well as the "Aspyr ABI hack" commits which I think haven't been picked.

IanE9:
> Broadly speaking, I haven't submitted these commits yet because I want to go through and add explanations, as well as sort which ones are required for gcc, clang, and non-Windows. There are also a couple of dubious ones, like "Remove using namespace std", the various unity build commits, and the FLua ErrorHandler/COMPILE_TIME_CONDITION commits, as well as the "Aspyr ABI hack" commits which I think haven't been picked.

Since you've indicated you've got plans to submit these yourself at some point I'm comfortable closing this.

As for `using namespace std`, I'm of the opinion that we almost certainly shouldn't be doing that in the PCH or any other header. There are a great many opportunities for name conflicts that propagate to every compilation unit. I don't think it would be too much to ask that people just add `using namespace std` inside blocks that make heavy use of standard types or even just add it after all headers have been included in a CPP file.

JohnsterID:
> gcc, clang, and non-Windows

@Hello71 Any potential for mingw-w64 or no point considering the linking issues and release differences?

Hello71:
> Any potential for mingw-w64 or no point considering the linking issues and release differences?

This is definitely not possible, for several reasons:

1. The linker will almost certainly reject this combination.
2. It is not generally possible to use multiple C++ standard libraries in a single process. They will conflict in global variables, TCB/TLS, and probably others.
3. The ABI is different between libstdc++ and MSVC, meaning that C++ calls between the EXE and DLL will be mangled, and also C++ calls between the DLL and LIBs.

JohnsterID:
Thank you. I appreciate the detailed explanation. Watching the process and learning about the different parts is interesting.

PR #9188, opened 2022-09-12
Replace "Order" text with "Communism" to match the actual ideology name

Pathologic72:
Not sure if this is intentional, but it bugs me how two different names are used for the ideology throughout the mod.

This replaces all instances of "Order" in the UI and diplomacy text with "Communism" to match the name currently used in the game.


LessRekkless:
**Order** is the correct term for this ideology. Where is communism used?

PR #9228, opened 2022-09-28
clang-tidy

JohnsterID:
Some clang-tidy changes for review. Clang-tidy can be a rabbit hole (https://github.com/LoneGazebo/Community-Patch-DLL/pull/9215) but someone with more knowledge would have to use the tool.

```
error: no template named 'vector'; did you mean 'std::vector'? [clang-diagnostic-error]
error: must use 'class' tag to refer to type 'CvImprovementEntry' in this scope [clang-diagnostic-error]
```

### DIFF
--- a/CvGameCoreDLLUtil/include/CvWeightedVector.h
+++ b/CvGameCoreDLLUtil/include/CvWeightedVector.h
@@ -269,7 +269,7 @@ public:
 	};
 
 private:
-	vector<WeightedElement> m_items;
+	std::vector<WeightedElement> m_items;
 };
 
 //-------------------------------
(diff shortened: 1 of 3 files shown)


IanE9:
I'm not sure at all what would be causing clang-tidy to believe that `class` is necessary here. `class` in this position is supposed to tell the compiler that the name that follows it refers to a class type, but this should already be known since `CvImprovementEntry` has already been declared on line 47. The usage here is valid but it shouldn't change the meaning of the code nor should it be necessary (I think). Maybe the `_Ret_maybenull_` macro on the previous line is confusing it somehow?

PR #9263, opened 2022-10-06
make minimum island size for barbarian camp spawns configurable

artemigkh:
Defaults to 1, which is identical to old behavior, but can now be configured so that barbarians won't spawn on islands with less than n tiles


ilteroi:
yes @ landmass; but also wasn't the plan to allow pirate islands but not to count them?

PR #9292, opened 2022-10-15
Great Engineer / Wonder Balance

LessRekkless:
- Great Engineer production is based off best city, but for only 5 turns
- Owned World Wonders provide 5% increased cost if they are more than 2 eras old (currently 0%)


LessRekkless:
If this has too much bad mouth feel for conquerors, maybe this entire thing could be based on constructed wonders.  Would require another array, could be an unordered_map to save space.

Issue #9281, opened 2022-10-09
Urbanization Unhappiness in City Unhappiness Breakdown doesn't use "Red City" icon


RecursiveVision:
Assigning to you as requested @nobubblegums

PR #9301, opened 2022-10-15
Cityview unhappiness consistency & A new unhappy citizen icon for the banner

nobubblegums:
Fix for https://github.com/LoneGazebo/Community-Patch-DLL/issues/9281
- Cityview unhappiness tooltip now displays text consistent with the yield unhappiness (instead of the flavored text reserved for the city banner)
- Yield deficit uses red color instead of Cyan
- The citizen icon is replaced with the newly added "red citizen" on the city banner messages.

@RecursiveVision & @LessRekkless 
Since you are working on unhappiness, can you change outdated icons you come across for urbanization or "unhappy citizen" with the newly added ones? It's [ICON_URBANIZATION] for unhappiness caused by specialists, and [ICON_CITIZEN_RED] should be used in wherever a "green citizen" icon is used to signify unhappiness.


nobubblegums:
Hm fair point. I'm just deadset on green vs red color coding, but I can take that part out, or call a vote on that. (or not, we already have 60 proposals this session)

PR #9307, opened 2022-10-17
Update Civilopedia Happiness Section

axatin:
Rework of the Civilopedia Section on Happiness. The previous version was hard to understand, too much talk about functions and global medians. Also added missing information on Puppet Cities, Resisting Cities and Needs Modifiers. My sources were the previous civilopedia text, the wiki and the patch notes. Please check if everything is correct.

I didn't include the needs modifier from science here because I saw it will be removed in the next version, is that correct?


axatin:
Screenshot (138)
Screenshot (139)
Screenshot (140)
Screenshot (141)
Screenshot (142)
Screenshot (143)
Screenshot (144)
Screenshot (145)
Screenshot (146)

nobubblegums:
Great work!

LessRekkless:
Will continue with my edit pass throughout the day.

bsesar:
Thanks for the great work, @axatin! 👍 

If you don't mind, here are some suggestions for your text:
- "Happiness" page: puppet cities and occupied cities --> Puppet Cities and Occupied Cities (maybe also for consistency capitalize City and Cities throughout all pages)
- "Local Happiness in Cities" page: deck of cards style --> deck-of-cards style
- "What Causes Unhappiness" page:
         - from the different sources --> from different sources (not certain about removing "the", though)
         - in the other cities --> in other cities
         - I would suggest putting "Distress" before "Poverty" because the order of Needs still matters, right?
         - "  Poverty" has an extra space or two before the word itself
         - until to do something about it --> until you do something about it
         - "Urbanization" always sounded to me like workers being jealous of the intellectual elite 😀  
- "Calculation of Needs Thresholds" page: Thess --> These

axatin:
Thank you for your corrections, @bsesar! I fixed the typos (the capitalization is still a problem, hard to make it really consistent). Renaming Urbanization to Class Struggle is something I can't decide on my own.

Does the order of Needs matter? I didn't know this. This information should be included somewhere...

@LessRekkless feel free to make edits

Issue #9225, opened 2022-09-27
Strategic resource amounts in trade screen cannot be set, and may be secretly altered


Mandritti:
_1. Mod Version (X.Y.Z). Current Version: 2.6.0_

2.6.0, VP with promotion icons and tree, no EUI.

_3. List of Other Mods_

None.

_4. Describe the Issue_

It is impossible to set an amount for a strategic resource in trade with the AI which is not replaced with a different number, in this case the maximum I have.
It is possible, as here, to make the screen say 0, but if you accept the trade, it trades away 3 horses.

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[thesaves.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/9659399/thesaves.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

[thelogs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/9659394/thelogs.zip)

_8. Steps to reproduce the Issue (Optional)_

On turn 94, go to diplomacy with Wu Zetian by clicking her capital banner.  Add Marble from her, 4 gold per turn from you, then horses.  Edit the horse amount to anything at all.  Clicking somewhere else - another item, back to the gold amount, or minimizing a field such as Strategic Resources - will change the amount of horses to either 3, or 0 (but actually 3).

_9. Screenshots of the Issue (Optional)_
![20220927154444_1](https://user-images.githubusercontent.com/11051545/192627180-124cb369-a4b7-4a50-8787-8a8633c82166.jpg)

PR #9313, opened 2022-10-20
Fix for Issue 9225

axatin:
Fix for Issue #9225. The issue here was the following function with a misleading name:
```
int CvLuaDeal::lGetNumResource(lua_State* L)
{
	CvDeal* pkDeal = GetInstance(L);
	const PlayerTypes ePlayer = (PlayerTypes) lua_tointeger(L, 2);
	const ResourceTypes eResource = (ResourceTypes) lua_tointeger(L, 3);

	//the name of this method is highly misleading ... 
	//we actually want to report the amount of resources the player has available outside of this deal
	int iNumAvailablePreDeal = GET_PLAYER(ePlayer).getNumResourceAvailable(eResource, false);
	int iResult = iNumAvailablePreDeal - pkDeal->GetNumResourceInDeal(ePlayer, eResource);

	lua_pushinteger(L, iResult);
	return 1;
}
```
It was used in TradeLogic.lua incorrectly to get the number of tradeable resources, as the function name would suggest (why didn't they change the name of the function instead of making a comment about a misleading name?...)
In the Lua file there is actually no need for a function returning "the amount of resources the player has available outside of this deal", and the total number of tradeable resources is accessible using CvPlayer::getNumResourceAvailable, so I removed CvLuaDeal::lGetNumResource altogether and changed the Lua file accordingly.

Also fixed an issue that the deal valuation is not updated when the amount of strategic resources proposed is changed.

### DIFF
--- a/(2) Vox Populi/LUA/TradeLogic.lua
+++ b/(2) Vox Populi/LUA/TradeLogic.lua
@@ -730,8 +730,8 @@ end
 -- Update buttons at the bottom
 ---------------------------------------------------------
 function DoUpdateButtons()
-	
-	-- Dealing with a human in a MP game
+
+-- Dealing with a human in a MP game
     if (g_bPVPTrade) then
 		
         --print( "PVP Updating ProposeButton" );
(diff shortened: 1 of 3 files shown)


axatin:
@RecursiveVision wait, I just noted that there is also a TradeLogic.lua file for the EUI version that needs to be changed too. I will open a new PR soon

axatin:
okay, in the other Lua file the bug had already been fixed, but they chose to keep the original code as well:
```
print("This is the modded TradeLogic from Vox Populi")
----------------------------------------------------------------        
----------------------------------------------------------------     
-- modified by bc1 from 1.0.3.276 brave new world code
-- code is common using gk_mode and bnw_mode switches
```
```
if gk_mode then
	iAmount = math.min(g_Deal:GetNumResource(g_iUs, resourceId),iAmount);
else
	iAmount = math.min(Players[g_iUs]:GetNumResourceAvailable(resourceId, false),iAmount);
end
```
Is this how it should be done? Shall I re-include CvLuaDeal::lGetNumResource into the dll again? Or can I remove the function call also from the EUI Lua file?

Issue #9278, opened 2022-10-09
Show Growth Penalty and Settler Cost Increase in Top Panel Happiness Overview


RecursiveVision:
Do it, future me.

axatin:
I can also do this, I have to change the growth penalty formula anyway (congress proposal)

PR #9362, opened 2022-10-29
(Congress 10/22, Prop. 33) Smooth Unhappiness Growth Penalty

axatin:
Implementation of [Congress Proposal](https://forums.civfanatics.com/threads/33-proposal-smooth-grow-penalty-from-unhappiness.680005/) and of #9278


axatin:
Edit: I would like to use text colors for the approval rating also in the non-EUI version of the top panel, I think it is helpful. If the players don't like it, I can easily revert it back.

Issue #9350, opened 2022-10-27
2.7.3 - Turtle Ship keeps area in city with lake but no ocean access.


Stalker012345:
_1. Mod Version (X.Y.Z). Current Version: 2.7.3_

2.7.3

_2. DirectX Version_

_3. List of Other Mods_

[Other Mods.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/9874922/Other.Mods.zip)

_4. Describe the Issue_

A CS keeps gifting me a turtle ship in my capital, which is 100% useless due to the lack of ocean access.

![image](https://user-images.githubusercontent.com/20593976/198169349-a7b8bf1e-5532-4174-9c95-38e3a08cb3f7.png)

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[AutoSave_0231 AD-1605 - Copy.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/9874910/AutoSave_0231.AD-1605.-.Copy.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

[Uploading Logs - Copy.zip…]()

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

_8. Steps to reproduce the Issue (Optional)_

_9. Screenshots of the Issue (Optional)_

PR #9392, opened 2022-11-08
#9350 minor civ naval gifts

LessRekkless:
Militaristic City States should now gift naval units to the nearest city to which they are attached by the same (ocean) land mass.  If there is no city, then they to the nearest **coastal** city , or do not gift naval units.

In addition, only **coastal** City States will gift naval units.

* as always, coastal = adjacent to body of water made of 10+ plots

Fix for #9350


LessRekkless:
untested.

Issue #9436, opened 2022-11-24
3.0.1 - Bastion Fort Tooltip Text very slight issue


Stalker012345:
_1. Mod Version (X.Y.Z). Current Version: 3.0.1_

3.0.1

_2. DirectX Version_

_3. List of Other Mods_

_4. Describe the Issue_

![image](https://user-images.githubusercontent.com/20593976/203690081-05a2d194-888f-4412-b775-d78296945c5f.png)
A few extra apostraphes, barely worth mentioning.

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

_8. Steps to reproduce the Issue (Optional)_

_9. Screenshots of the Issue (Optional)_

PR #9447, opened 2022-11-24
Text fix

axatin:
Fix for #9436

Btw: Why is the TXT_KEY_BUILDING_ODEON_STRATEGY updated in the SQL file? Why not directly write it like that in the XML?


LessRekkless:
Because it's updating text, not adding a new text.

For text, this is generally the standard differentiation between sql and xml (with a few exceptions, related to adding a text where a related text already exists eg. promotions with names and not help text)

axatin:
I know that xml and sql are used for different things, that's not my problem. My question is: TXT_KEY_BUILDING_ODEON_STRATEGY doesn't exist in the base game, it's newly defined in the mod. Why is it immediately changed in another file of the same mod? Doesn't that mean that the original value that has been set in the XML is never used? (Same question also applies to PR #9454)

LessRekkless:
> I know that xml and sql are used for different things, that's not my problem. My question is: TXT_KEY_BUILDING_ODEON_STRATEGY doesn't exist in the base game, it's newly defined in the mod. Why is it immediately changed in another file of the same mod? Doesn't that mean that the original value that has been set in the XML is never used? (Same question also applies to PR #9454)

Ah, yes. I read ODEON and thought OPERA_HOUSE.  You're right in this case that there should only be one text line, in the XML.

The xmls for new buildings and units are more annoying; in situations with new units, I generally have a commented line stating that the actual value is in the other file. But in general, these files group all the units together to make sure the modder know what the entries contemporaries are like.

PR #9460, opened 2022-11-26
Human player handicap yields fix

legenciv:
The handicap yields for the human player were not being given due to the lack of the <HandicapInfo_DifficultyBonus> related parameters. Added and copied the parameters used for the AI to the human player.

Tested locally on emperor difficulty, both with BonusBase/A/B/C parameters at 0/0/0/0 and 10/350/230/145 and working as intended.


jayeye2011:
This should have been a  voted on in a VP Congress proposal.  Doesn't giving the same bonuses to the human player as the AI player on prince and above difficulty levels just makes it easier on higher difficulty levels? This would be the opposite of what Vox Populi is all about.

legenciv:
The bonuses aren't active to the human player by default, since all the ABC parameters for the human player are at 0. The player has to manually change the parameters in the file to gain any yields from handicap.

It can be used to study how the AI benefits from them in practice. Useful when testing hypothesis regarding balance or proposals, or when someone just wants to see what the AI is capable of in a given difficulty. I'm currently using it to study how AI Pedro benefits from the GAP coming from the handicap.

jayeye2011:
Okay, my bad for not looking closely enough at the parameter values.

PR #9529, opened 2022-12-08
FixCargoShips

### LINKED ISSUE #9518: Cargo Ship is stuck in a Fort and cannot be moved
_1. Mod Version (X.Y.Z). Current Version: 3.0.4_

_4. Describe the Issue_
I have built a Cargo Ship in Klagenfurt. The city is connected to sea tiles via a fort. When the Cargo Ship is created, it is placed in the fort and I cannot move it to a different city or create a Trade Route.

_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_
[cargo_ship.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/10147647/cargo_ship.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_
[Logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/10147648/Logs.zip)

Build a Cargo Ship in Klagenfurt and try to send it somewhere.

_9. Screenshots of the Issue (Optional)_
land_locked_cargo_ship

axatin:
Fix #9518 
- Cities that are not adjacent to water can't be connected by canals to the ocean
- Canals check is only applied in VP, no longer in CP

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlot.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlot.cpp
@@ -1058,64 +1058,64 @@ bool CvPlot::isCoastalLand(int iMinWaterSize, bool bUseCachedValue, bool bCheckC
 					return true;
 			}
 		}
-	}
 
-	// Not adjacent to water or not adjacent to enough water? If not checking for canals, abort!
-	// Also, water tiles are not land, so always return false
-	if (!bCheckCanals || !MOD_GLOBAL_PASSABLE_FORTS || isWater())
-		return false;
 
-	//check for areas of water connected by canals
-	//starting with the plot itself, we loop through all adjacent plots and add them to a list if they are water tiles
-	//or if there is an owned fort, citadel or city on them
-	//we do this repeatedly until we reach the required amount of plots or until every item of the list is checked
-	std::vector<const CvPlot*> vAccessibleWaterPlots(1, this);
-	TeamTypes eTeam = getTeam();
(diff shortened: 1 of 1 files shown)


axatin:
Then a canals check should be added in `isCoastalLand` here, I assume:
```
bool CvPlot::isCoastalCityOrPassableImprovement(PlayerTypes ePlayer, bool bCityMustBeFriendly, bool bImprovementMustBeFriendly) const
{
	bool bIsCity = isCity() && isCoastalLand();
	// Good enough
	if (bIsCity)
	{
		if (bCityMustBeFriendly)
			return IsFriendlyTerritory(ePlayer);
		else
			return true;
	}
```
But that would be quite a change. Would you mind if I opened a proposal for that to discuss it in the forum? It should be a community decision I think

axatin:
@LessRekkless pointed out in the forum discussion that the current implementation is that cities are passable for naval units only if they are next to a water tile (same as forts and citadels). So I think the fix here is correct, it makes the canals check consistent with the passability rules.

axatin:
Seems that is has no effect currently, so yes, I'd say so

axatin:
In the canals check, we can replace ` if (pAdjacentPlot->getTeam()==eTeam)` by ` if (pAdjacentPlot->getTeam()==eTeam || MOD_GLOBAL_PASSABLE_FORTS_ANY). I assume it's a similar change for the passability rules, but I don't know where they're implemented.

Issue #9523, opened 2022-12-06
3.0.4 My units can't get any XP :(


brsoft1981:
1.  3.0.4

2. DirectX Version

3. Only Community Patch

4. So, I'm playing with Japan. From some point of the game, my units stopped gaining any XP

______________________________________________________________________________________
5. Attached [Ivo.Civ5Save.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/10166749/Ivo.Civ5Save.zip)

6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

8. Steps to reproduce the Issue (Optional)

9. Screenshots of the Issue (Optional)_

JtG82:
I have the same problem. Against barbarians or city states the xp is coming, but not against other major civs. Therefore I did a little digging and testing. I noticed that it happens only when playing on "Marathon" game speed, other speeds are fine. I also tried the game on previous version (2.7.3.) and it worked fine on all speeds. 

So my wild guess is that it might have something to do with this new change in "cvunit.cpp" for it affects only normal combat and not barbarians or city states:

"
if (MOD_BALANCE_CORE_SCALING_XP)
	{
		iMaxValue *= GC.getGame().getGameSpeedInfo().getTrainPercent();
		iMaxValue /= 100;
	}
"
(Link to the change: https://github.com/LoneGazebo/Community-Patch-DLL/commit/391d8fcf0b329ed72a6372a0abac3eaf744fa510 )

Since the marathon gamespeed has ID "0" on civ5gamespeeds.xml, is it possible that the code on some point multiplies the xp value with the ID number and thus it ends up giving zero xp on marathon?

brsoft1981:
Yep, I'm playing on 'marathon' speed too, so maybe that's the issue.

alex-me642:
hi.  

> iMaxValue *= GC.getGame().getGameSpeedInfo().getTrainPercent();

the problem here is not getTrainPercent(), but the multiplication itself.   When an unit battles with barbarians or with civ states, iMaxValue above is defined as some const but with real civs it's equal to INT_MAX and the multiplication gives as integer overflow - most probably negative value.

it should be fixed ASAP, currently exp & promotion logic has been broken.

alex-me642:
hi, the logic is okay, and for barbs and civ states it works properly.

[...]

PR #9530, opened 2022-12-08
XP Cap fix

### LINKED ISSUE #9461: Barbarians give no exp
I am pretty sure [this commit](https://github.com/LoneGazebo/Community-Patch-DLL/commit/4970fc31f511fb9de14c8c5cfb748b20262aa9d5) breaks gaining exp from barbs whatsoever if player manually change BARBARIAN_MAX_XP_VALUE to -1 in CoreDefines.sql

I mean it broke in my game, but changing BARBARIAN_MAX_XP_VALUE back to positive value fixed everithing after reload.

LessRekkless:
- Now the default max XP is -1, which translates to INT_MAX in the change/setExperienceTimes100() functions.
 - ALL negative max XP is now considered no limit instead of just -1
 - fixes #9461 and #9523


LessRekkless:
untested!

PR #9546, opened 2022-12-12
Partial Fix 9504

### LINKED ISSUE #9504: Production tooltip incorrect
Version 3.0.4, non-EUI
There are several things wrong in the production tooltip of cities (non-EUI version). First, the Wltkd-modifier is listed twice. Second, the values don't add up at all. The base value is not the sum of the values listed above, and the total value is not 135% of the base value. 

Terrain, Buildings and extra Production from Population are correct. It should be 7 from specialists and not 3. What exactly is "extra production from population in empire"? That value looks wrong
Screenshot (190)

[Savegame](https://github.com/LoneGazebo/Community-Patch-DLL/files/10140835/Production.zip)

axatin:
Partial Fix #9504 

One part of the problem is an error in the lua file, but there are still some other issues here...

### DIFF
--- a/(2) Vox Populi/LUA/InfoTooltipInclude.lua
+++ b/(2) Vox Populi/LUA/InfoTooltipInclude.lua
@@ -1537,7 +1537,7 @@ function GetYieldTooltip(pCity, iYieldType, iBase, iTotal, strIconString, strMod
 	end
 
 	-- Base Yield from Pop in Empire
-	local iYieldPerPopInEmpire = pCity:GetYieldPerPopTimes100(iYieldType);
+	local iYieldPerPopInEmpire = pCity:GetYieldPerPopInEmpireTimes100(iYieldType);
 	if (iYieldPerPopInEmpire ~= 0) then
 		local iYieldFromPopInEmpire = iYieldPerPopInEmpire * Players[pCity:GetOwner()]:GetTotalPopulation();
 		iYieldFromPopInEmpire = iYieldFromPopInEmpire / 100;


axatin:
Explanation for the second fix: Yield changes for features were added to the base yield rate  in  CvCity::getBaseYieldRate. But those values don't represent the total yields from worked features in the city, but the addional yields of features in general in the city. And no values for features should be added separately in getBaseYieldRate at all, because yield changes from worked tiles with features are already processed in GetBaseYieldRateFromTerrain.

axatin:
Explanation for the specialists fix: The problem was in the function
```
void CvCity::updateExtraSpecialistYield(YieldTypes eYield, SpecialistTypes eSpecialist)
{
	VALIDATE_OBJECT
	CvAssertMsg(eYield >= 0, "eYield expected to be >= 0");
	CvAssertMsg(eYield < NUM_YIELD_TYPES, "eYield expected to be < NUM_YIELD_TYPES");

	int iOldYield = getExtraSpecialistYield(eYield);
	int iNewYield = 0;

	if (eSpecialist == NO_SPECIALIST)
	{
		for (int iI = 0; iI < GC.getNumSpecialistInfos(); iI++)
		{
			iNewYield += getExtraSpecialistYield(eYield, ((SpecialistTypes)iI));
		}
	}
	else
		iNewYield += getExtraSpecialistYield(eYield, eSpecialist);

	if (iOldYield != iNewYield)
	{
		m_aiExtraSpecialistYield[eYield] = iNewYield;
		CvAssert(getExtraSpecialistYield(eYield) >= 0);

		ChangeBaseYieldRateFromSpecialists(eYield, (iNewYield - iOldYield));
	}
}
```

The value in iOldYield is the total amount of extra yield of type `eYield` from specialists in the city. So, it's the sum of the extra yields from all specialists (of all types) that are currently assigned in the city. If `eSpecialist != NO_SPECIALIST`, then iNewYield contains the (newly calculated) sum of extra yields of type `eYield` only from Specialists of type `eSpecialist`. Comparing these two values doesn't make any sense. So, the function should never be called with a value of eSpecialist other than NO_SPECIALIST, and that's why I removed the second argument of the function altogether.

axatin:
And then a lot of things in the non-EUI LUA file were outdated

PR #9620, opened 2023-01-12
Fix Nuclear Plant gives Civil Servant more yields

n-core:
I'm waiting for the confirmation about this oversight. I don't think it's intendedfor Nuclear Power Plant to give Civil Servant more yields than other specialists.


n-core:
This one should be an oversight. You can merge it.

Issue #9605, opened 2023-01-08
Infiltrators promotion bug


Aqinu:
_1. Mod Version (X.Y.Z). Current Version: 3.1.1

_4. Describe the Issue_

_9. Screenshots of the Issue (Optional)_
Bonus "no adjacent friendly unit" doesn't work when you are attacked by melee unit (the bonus appears correctly in preview), it should work as the bonus "outside friendly territory" (which work correctly for both melee and range)

6 + (6 * 0.15) = 6.9 what is calculated
6 + (6 * (0.15 + 0.1)) = 7.5 what is should be

![Screenshot (29)](https://user-images.githubusercontent.com/49090452/211189256-86fa1be8-47cd-4152-be05-72d97bd58645.png)

For range is ok
9 + (9 * (0.15 + 0.1)) = 11.25

![Screenshot (30)](https://user-images.githubusercontent.com/49090452/211189430-1edbe8ac-8604-4490-a7a2-d09629ccdc21.png)

PR #9622, opened 2023-01-12
Infiltratorsfix

axatin:
not entirely sure about this one, please check thoroughly.

the problem here is in the function 
`CvUnit::GetGenericMeleeStrengthModifier(const CvUnit* pOtherUnit, const CvPlot* pBattlePlot, bool bIgnoreUnitAdjacencyBoni, const CvPlot* pFromPlot, bool bQuickAndDirty)` calculating modifiers for melee attacks.

This function is called both for the defending and for the attacking unit, and it checks for friendly units (and combat-modifying improvements, and great generals) nearby on pFromPlot: `if (!pBattlePlot->IsFriendlyUnitAdjacent(getTeam()) {...}`.

I think the logic here should be that when attacking, it should be pFromPlot (the plot of the attacker) that is checked, and when defending, it should be pBattlePlot (the plot of the defender). A similar distinction is made in `CvUnit::GetMaxRangedCombatStrength`.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvUnit.cpp
+++ b/CvGameCoreDLL_Expansion2/CvUnit.cpp
@@ -16342,7 +16342,7 @@ int CvUnit::GetDamageCombatModifier(bool bForDefenseAgainstRanged, int iAssumedD
 
 //	--------------------------------------------------------------------------------
 /// What are the generic strength modifiers for this Unit?
-int CvUnit::GetGenericMeleeStrengthModifier(const CvUnit* pOtherUnit, const CvPlot* pBattlePlot, 
+int CvUnit::GetGenericMeleeStrengthModifier(const CvUnit* pOtherUnit, const CvPlot* pBattlePlot, bool bAttacking,
 				bool bIgnoreUnitAdjacencyBoni, const CvPlot* pFromPlot, bool bQuickAndDirty) const
 {
 	VALIDATE_OBJECT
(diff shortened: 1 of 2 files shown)


LessRekkless (commented):
I think adding a bAttacking boolean to a function that is explicitly supposed to be unrelated to whether the unit is attacking or not is the wrong solution.

I think this can be solved without such a requirement, but will have to come back later.

What is important is whether the *unit with the promotion* is adjacent to allies, not if the battle space is adjacent to allies. That means you should only be looking at plots adjacent to the one where the unit with the ability is currently situated. It has nothing to do with pBattlePlot at all.

LessRekkless:
oof.  I have a lot to say here; I'll be back.

LessRekkless:
Okay, before I get into my big discussion about creating a convention for how auras and adjacent mods should work, I just want to state specifically _why_ the bug in #9605 exists.

Why did
```
(!bIgnoreUnitAdjacencyBoni && !pFromPlot->IsFriendlyUnitAdjacent(getTeam(), /*bCombatUnit*/ true))
```
not work initially?  `pFromPlot` is the the plot of the attacker's (player warrior) start location, whereas `getTeam()` is the defender's (city-state archer) team.  `IsFriendlyUnitAdjacent()` therefore checks plots adjacent to the _warrior_ to see if anyone allied to the _city-state_ exists.  In this case, that's the archer unit itself.

Clearly, we cannot use `pFromPlot` for this ability, and however we choose to fix it, the other adjacent bonuses should follow suit.

LessRekkless:
Continuing my Adjacency thoughts: Flanking bonuses are an adjacency bonus, but use the _enemy's_ plot (ie `pOtherUnit->plot()`). Half the time, this is `pBattlePlot` (during this unit's attack) and the other half of the time this is `pFromPlot` (during this unit's defense). As such, I don't think that we want adjacency bonuses to be based on specifically `pBattlePlot` either, since it would completely envelop **FlankAttackModifier** (when this unit attacks).

Therefore, I proffer that Adjacency bonuses be specifically the opposite of Flanking bonuses, ie. using this unit's `plot()` function. This is how adjacency works with ranged, and how axatin has set it up in this code (although it's setup as a flawed analog to the ranged function's `pTargetPlot`, which itself is actually an analog to the melee function's `pBattlePlot`).

LessRekkless:
@ilteroi do AI combat simulations only ever involve the active player moving units, or do inactive players have their units moved as well?

LessRekkless:
@axatin I submitted a PR to your branch with my recommended changes.

In the end, I agree that aura and adjacency bonuses are best targeted at the location where the unit is before combat takes place (after movement). This is how it works for ranged combat, and also how EUPanel displays its bonuses.

The easiest way to send this information to `CvUnit::GetGenericMeleeStrengthModifier()` does seem to be through an additional variable specifying which plot belongs to each unit, and a boolean works just fine for this. If anyone has any other ideas, noting that `pFromPlot` and `pBattlePlot` are not inherently linked to either unit, I'd love to hear them.

@RecursiveVision
 - a unit now benefits from all auras and adjacency bonuses based on its location before combat
 - this change/consolidation differentiates these bonuses from flanking, which is based off of the enemy's location before combat; all other location-based combat bonuses are based on the location where the _battle is taking place_.
 - great generals/admirals now benefit units defending against ranged attacks

LessRekkless (changes requested):
@axatin I submitted a PR to your branch with my recommended changes.

In the end, I agree that aura and adjacency bonuses are best targeted at the location where the unit is before combat takes place (after movement). This is how it works for ranged combat, and also how EUPanel displays its bonuses.

The easiest way to send this information to `CvUnit::GetGenericMeleeStrengthModifier()` does seem to be through an additional variable specifying which plot belongs to each unit, and a boolean works just fine for this. If anyone has any other ideas, noting that `pFromPlot` and `pBattlePlot` are not inherently linked to either unit, I'd love to hear them.

@RecursiveVision, in addition to fixing #9605, the following might be useful in the changelog
 - a unit now benefits from all auras and adjacency bonuses based on its location before combat
 - this change/consolidation differentiates these bonuses from flanking, which is based off of the enemy's location before combat; all other location-based combat bonuses are based on the location where the _battle is taking place_.
 - great generals/admirals now benefit units defending against ranged attacks

axatin:
@LessRekkless I have merged the changes you proposed.

PR #9632, opened 2023-01-24
Changes for proposal (3-09), except Covert Action and Wind Farm

legenciv:


legenciv:
Added last change for proposal 3-09. I was waiting for pineappledan's answer for N.Core's question, but it's been days without answer and the thread of the question is already closed to replies.

pineappledan:
Have you tested what adding yields onto mountain tiles will do?
If it makes mountain tiles workable for everyone then you shouldn't do it.

pineappledan:
Rather than implementing the water yields via Building_LakePlotYieldChanges and Building_SeaPlotYieldChanges, why not just add them with Building_TerrainYieldChanges using TERRAIN_COAST and TERRAIN_OCEAN ?

That's the same number of database entries and keeps all the boosts together.

legenciv:
It does make mountains workable. This was as stated in the tooltip, and prevents the Wind Farm from not working for the Inca.

The implementation via Building_LakePlotYieldChanges and Building_SeaPlotYieldChanges just keeps the entries consistent with how other buildings also apply their yields to such tiles.

Is it intended for an Inca player to not get yields on mountains from a Wind Farm? If not, I expect someone to eventually report it as a bug when they try it.

pineappledan:
Inca's yields on mountains already work by different rules. No one else gets to work mountains, and no one else has terrain yields that scale on era. If someone complains that wind farms don't further augment them then it's a simple matter of Inca being a special case. the wind farm adding yields to otherwise unworkable mountains for the other 42+ civs is far weirder.

legenciv:
Understood. I did wish you had answered earlier, since I wanted confirmation on that. The rest of the building's implementation had its own inconsistencies, so it was easy to think the mountains were another one.

n-core:
Just change the Wind Farm text to flat, hills, and water. Don't need to add yield to Mountain since it will become an awkward building where your city suddenly can work on mountain tiles.

pineappledan:
That’s why you ping, @legenciv so I can get a push notification. I am not lurking in people’s GitHub commits.

legenciv:
You were pinged in the forums, days ago. I held on committing on github to see your answer.

azum4roll:
What does Wind Farm look like on the EUI tooltip now?

legenciv:
@azum4roll 
wind farm

Detailed info on which tiles are affected, consistent with how the Solar Farm tooltip does. I can't say I like how it looks, but it is clearer and consistent with how this tech's buildings state their effects, making it easier to compare them.

solar farm

PR #9638, opened 2023-01-29
Proposals (2-02) and (2-31)

axatin:
- [Proposal (2-02)](https://forums.civfanatics.com/threads/2-02-city-state-coup-mechanic-proposals.681128/)
- [Proposal (2-31)](https://forums.civfanatics.com/threads/2-31-proposal-dont-reveal-information-about-counterspies-in-city-states.680661/)
- CP only: Coup chance calculation reverted to the BNW version (there were unintended VP changes there)

### DIFF
--- a/(2) Vox Populi/Balance Changes/Text/en_US/PolicyText.sql
+++ b/(2) Vox Populi/Balance Changes/Text/en_US/PolicyText.sql
@@ -166,7 +166,7 @@ WHERE Tag = 'TXT_KEY_POLICY_CIVIL_SOCIETY_HELP';
 
 -- Covert Action
 UPDATE Language_en_US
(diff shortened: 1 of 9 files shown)


axatin:
Second commit: Changes to AI logic.
- AI players can now place spies in city-states they're at war with (rigging elections and increasing coup chances still possible while at war)
- thresholds used to determine if the AI performs a coup are increased, and now depend on spy rank (if the success chances are low, it's better to wait until the next election; high-level spies are more valuable and should not try risky coups)
- question: why is a random roll done in CvEspionageClasses, Line 6885? It's not used to calculate if the coup is successful, that's done in `AttemptCoup`, but rather to determine if a coup should be tried by the AI. The random roll seems to be done every turn, to all this does is delaying coups with lower success chances by some turns...

PR #9648, opened 2023-02-05
Bugfix (2-57)

axatin:
I played a game with the new garrison proposal ([PR](https://github.com/LoneGazebo/Community-Patch-DLL/pull/9623)) implemented and I noticed a bug in the calculation of the ranged strike strength of cities. Naval units need to be excluded there as well. Also, the AI still used naval units as garrisons, this is now prevented.


azum4roll:
Good old negative city strike RCS

axatin:
exactly

Issue #9618, opened 2023-01-12
3.1.1 CTD


jdoug312:
_1. Mod Version (X.Y.Z). Current Version: 3.1.1_

3.1.1

_2. DirectX Version_

10

_3. List of Other Mods_

1-4B
Community Events v7
7C Israel
EAW
ENW
Ethnic Diversity
Great Prophet Historical Names
Nubia
Quick Turns
Sumer

_4. Describe the Issue_

CTD during AI's turn
______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[QuickSave.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/10397529/QuickSave.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

[logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/10397537/logs.zip)

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

[CvMiniDump.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/10397543/CvMiniDump.zip)

_8. Steps to reproduce the Issue (Optional)_

Hit next turn

_9. Screenshots of the Issue (Optional)_

axatin:
The problem here is that in the Phoenicia modmod, one of Phoenicia's cities is transferred to a city-state which is not alive at that point: (PhoenicianUI.lua)
```
local iShortestDistance = 99999
local pNewCityOwner = nil
for player in pairs(Players) do
	local MinorPlayer = Players[player]
	if (MinorPlayer and MinorPlayer:IsMinorCiv()) then 
	
		local iCSStartPlot = MinorPlayer:GetStartingPlot()
		if iCSStartPlot then
			
			local iDist = Map.PlotDistance(iCSStartPlot:GetX(), iCSStartPlot:GetY(), pNewCity:GetX(), pNewCity:GetY())
			if (iDist < iShortestDistance) then
				iShortestDistance = iDist
				pNewCityOwner = MinorPlayer
			end
		end
	end
end
pNewCityOwner:AcquireCity(pNewCity, false, false)
```

[...]

PR #9654, opened 2023-02-06
Fix9618

axatin:
Fix for #9618.

The CTD was caused by Indonesia's AI trying to get the (cached) distance between Indonesia and the City-State of Amia. The problem was that Amia had been founded in the very same turn (unique ability of Phoenicia). This modmod-related corner case wasn't covered and no distance had been calculated yet.

I've added that the distances are now updated if no cached distance exists yet. Question to the developers: Maybe this should be added before every `switch (GetPlayer()->GetProximityToPlayer(ePlayer))`, just in case? It should affect performance only minimally and make the AI better in case of any similar corner cases or errors that have gone unnoticed so far.


axatin:
@RecursiveVision you're right. I'll close the PR and continue the discussion in the comment section of #9618

PR #9657, opened 2023-02-11
Espionage related text fixes

n-core:
Yield clarity and fix for Police Station, also text fixes for NIA and White Tower.


n-core:
I dunno whether the text or the mechanic is wrong, but it's been there for a long time and left unnoticed.
https://github.com/LoneGazebo/Community-Patch-DLL/blob/62bc8866eb27bb41e9968fa7231ac36b60495e7a/(2)%20Vox%20Populi/Balance%20Changes/Buildings/BuildingChanges.sql#L895-L899

Or that `YIELD_GOLD` could be somehow interpreted as Science in-game?
But then White Tower gives Science and Gold when countering enemy spies.

PR #9686, opened 2023-03-01
Update PromotionChanges.sql

GeneralAmadeus:


GeneralAmadeus:
@RecursiveVision  What was wrong whit this?

PR #9687, opened 2023-03-01
Update PromotionChanges.xml

GeneralAmadeus:
proposal 3-11


GeneralAmadeus:
@RecursiveVision I have no idea how to do that.

PR #9688, opened 2023-03-01
Patch 10

GeneralAmadeus:
Going to bed version of proposal 3-11.


GeneralAmadeus:
@RecursiveVision Yes? At least as compete as i am able to make it before release. I have not changed the AI and have no idea how to do that at the moment.

PR #9691, opened 2023-03-02
Patch 15

GeneralAmadeus:


GeneralAmadeus:
A small name change because the previous name did not fit in the promotion box in the promotion tree.

PR #9702, opened 2023-03-12
Text fix (Tribute policy)

legenciv:
Text fix, the effect is working as intended. The Tribute policy states that it boosts heavy tribute, instead of giving culture and science on city founding/conquest, due to a later merge.


axatin:
The policy names should also be swapped: This one should be called "Imperium", and the other one "Tribute".

PR #9767, opened 2023-04-14
getActivePlayer() in getSmallFakeRandNum() causes desyncs

seroperson:
Hello! I did some MP debugging and looks like `getActivePlayer()` in `getSmallFakeRandNum()` causes a lot of desyncs across players because it differs on host and non-host player. With enabled logging it looks like (left - host; right - another one player): 
image

I'm not sure if `getStartYear()` is suitable here, point me if we should use something else.


ilteroi:
this is a very good find!

in fact, since getStartYear() is constant, it does not introduce any variation between successive calls, so it's not needed at all.

it's a bit unfortunate that getActivePlayer() does not work but we'll have to live with it.

PR #9782, opened 2023-04-19
Fix AI workers wrongly removing roads in some cases

### LINKED ISSUE #9741: AI doesn't build road connections properly
_1. Mod Version (X.Y.Z). Current Version: 3.2.2

_3. List of Other Mods_

_4. Describe the Issue_
The old issue still seems to be persisting. AI workers don't build proper road connection between the cities where it is 100% necessary.
In my example, there is no road on the tile where the Viking trebuchet is located. It's a crucial tile that connects their capital with another long established city. Also note the village built on the same tile and a river that interrupts the movement, but the road is missing. Turn 201.

_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[1 (2).zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/11069364/1.2.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

[Logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/11069368/Logs.zip)

### LINKED ISSUE #9757: AI workers keeps removing and readding roads
_1. Mod Version (X.Y.Z). Current Version: 3.3.1_

Vox Populi with EUI version 3.3.1 (seen on earlier versions as well).

_2. DirectX Version_

_3. List of Other Mods_

image

_4. Describe the Issue_

AI workers keep removing roads for seemingly no reason and then readding them in the same place directly afterwards.
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[road removal.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/11171351/road.removal.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

_9. Screenshots of the Issue (Optional)_

image

KungCheops:
Because of the early return in CvBuilderTaskingAI::ConnectCitiesToCapital, the AddRoutePlot function is never called for completed roads. Meaning that the m_routeNeededPlots list is never populated for existing road tiles on completed paths between cities.

Removing this early exit adds some extra computation load (of course). But as the behavior is currently broken with AI workers constantly adding and removing the same road tile over and over, the early exit is often useless anyway (any time a needed road is removed, we have to do all the computations anyway).

resolves #9757, should resolve #9741

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvBuilderTaskingAI.cpp
+++ b/CvGameCoreDLL_Expansion2/CvBuilderTaskingAI.cpp
@@ -265,18 +265,6 @@ void CvBuilderTaskingAI::ConnectCitiesToCapital(CvCity* pPlayerCapital, CvCity* 
 	// for quests we might be targeting a city state ...
 	bool bSamePlayer = (pTargetCity->getOwner() == pPlayerCapital->getOwner());
 
-	// if we already have a connection, bail out
-	bool bIndustrialRoute = (GC.getGame().GetIndustrialRoute() == eRoute);
-	if(bIndustrialRoute)
-	{
-		if (m_pPlayer->GetCityConnections()->AreCitiesDirectlyConnected(pPlayerCapital, pTargetCity, CvCityConnections::CONNECTION_RAILROAD))
-			return;
-	}
-	else if (m_pPlayer->GetCityConnections()->AreCitiesDirectlyConnected(pPlayerCapital, pTargetCity, CvCityConnections::CONNECTION_ROAD))
-	{
-		return;
-	}
-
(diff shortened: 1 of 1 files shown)


KungCheops:
I'm not sure but I think this behavior might have been caused by the recent addition of some additional tie-breakers, causing workers to prefer close tiles (and most preferring the tile they are currently on), causing them to want to remove roads instead of going to a further away tile to do something more valuable, but this seems unlikely to be the case, and honestly I'm unsure how this hasn't been an issue for years.

PR #9783, opened 2023-04-20
Don't try to move if MOVEFLAG_APPROX_TARGET_RING2 already satisfied

seroperson:
Hello! I have several MP saves where CS triremes are stucking because when they are trying to move to neighboring plot, the pathfinder says them smth like "you are already here (because of TARGET_RING2 flag)", `MoveToTargetButDontEndTurn` returns `false` (because unit's plot wasn't changed) and they are trying again and again until timeout. Looks like it's 'ok' for SP because timeout is `10` for local games, but for MP it's `1000` and sometimes we have to wait 2-3 mins each turn again and again. 
I'm not sure whether such fix is ok overall but at least it resolves such situations.


ilteroi (approved):
good catch; should also improve turn times ...

PR #9787, opened 2023-04-21
Fix AI logic for shortcut roads between non-capital cities

KungCheops:
AI workers should no longer remove and readd the same road tiles for shortcut roads.
AI workers will now build shortcut routes between cities which are not on the same landmass as the capital (and any other cities which do not have a direct road connection to the capital).


ilteroi (approved):
lgtm

PR #9790, opened 2023-04-22
Nullcheck when accessing plot->getUnitByIndex

seroperson:
Hello! Sometimes we are getting random crashes during MP session with callstacks like:
```
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvunit.cpp (25483): CvUnit::isDelayedDeath
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvtacticalai.cpp (8037): CvTacticalPlot::CvTacticalPlot
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvtacticalai.cpp (9764): CvTacticalPosition::addTacticalPlot
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvtacticalai.cpp (10156): TacticalAIHelpers::FindBestUnitAssignments
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvtacticalai.cpp (3633): CvTacticalAI::ExecuteAttackWithUnits
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvtacticalai.cpp (1136): CvTacticalAI::ExecuteDestroyUnitMoves
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvtacticalai.cpp (1259): CvTacticalAI::PlotBarbarianAttacks
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvtacticalai.cpp (821): CvTacticalAI::AssignBarbarianMoves
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvtacticalai.cpp (683): CvTacticalAI::ProcessDominanceZones
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvtacticalai.cpp (323): CvTacticalAI::Update
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvplayerai.cpp (285): CvPlayerAI::AI_unitUpdate
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvgame.cpp (9834): CvGame::updateMoves
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvgame.cpp (1586): CvGame::update
c:\users\seroperson\documents\github\community-patch-dll\cvgamecoredll_expansion2\cvdllgame.cpp (495): CvDllGame::Update
```
The middle of stacktrace can differ but last few calls are always the same. Crash reason is always a null pointer exception on CvUnit. 

I noticed that `plot->getUnitByIndex(i)` is usually followed by nullcheck but not everywhere. So, this PR adds nullcheck everywhere after `plot->getUnitByIndex(i)` to eliminate possible crashes. I'm not sure if it is required in all places but I think it won't be redundant.


seroperson:
Quite possible it is null because of some desync and the root cause should be fixed instead but as a simple additional failsafe I think such checks are fitting good there.

seroperson:
Yea, I agree that it sounds more like workaround and we should fix root cause, but in case of MP looks like it's a very long road when most desyncs will be eliminated. Moreover, if we are talking that such exceptions could help us with debugging 'something going wrong', I think it's not so clear: It would be actually good if whether I'm catching this exception during MP session, I can say what is the problem. But with such desyncs looks like it's a common situation when you crashed i.e. at some `CvUnit::isDelayedDeath` because of ... desynced food in some CS. And it's impossible to understand such relations just by dump, usually you need very deep analysis. So, in case of MP, that's why I think we should at least eliminate such crashes - because we already know that something going wrong: MP is desynced very much; and such crashes are just crashes which we can't use to gather any useful information.

I can suggest to also add logging to such places to be able to investigate that something is going wrong.

n-core:
You changed some EUI Lua files to use CP text here edb083d, don't do that. Add the file to the EUI compatibility instead if it doesn't exist there.
It will make EUI depend on CP even more, and I plan to detach EUI from CP. So any player who wants to play without CP can play without any UI problems.

seroperson:
Oh. I missed I still have opened PR and it now contains some of my dev changes. I'm closing it to not confuse anyone and I will open another one later.

PR #9806, opened 2023-04-25
Make AI less prone to reuse roads, should now be more willing to build quest routes as well.

KungCheops:


ilteroi:
14 commits and only a 6 lines of changes? can this be true?

KungCheops:
> 14 commits and only 4 lines of changes? can this be true?

Most of it was manually merged by Recursive, not sure how to remove the commits that are now superfluous.

KungCheops:
@RecursiveVision I just reverted the team logic to the way it was before. I guess this branch is messed up now so I'll just make a fresh one.

PR #9816, opened 2023-04-30
Replace hash algorithm

JohnsterID:
For review before merge. Added standard FNV-1a hash algorithm. The database order will need to checked that it is still maintained. Profiling required? For the FIXME, I'm still not sure about it, do we need to identify each container and source of the NULL entries then determine if it needs to be allowed or it is invalid?

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvGlobals.cpp
+++ b/CvGameCoreDLL_Expansion2/CvGlobals.cpp
@@ -3506,15 +3506,17 @@ void CvGlobals::GameDataPostCache()
 	calcGameDataHash();
 }
 
-template<typename T>
+template <typename T>
 static void HashGameDataCombine(CvGlobals::GameDataHash& seed, std::size_t& word, const T& container)
 {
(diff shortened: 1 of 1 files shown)


ilteroi:
i don't have concerns here but i don't understand which problem this is going to solve ... can you explain?

JohnsterID:
Original code comment stated that the hash algorithm could be better so I just replaced it with a better common standardised algorithm. It doesn't fix the FIXME though.

JohnsterID:
Added logging to original and new code to test. The NULL entries are still being hashed and output matches.

`logFile << "NULL entry in container of type " << typeid(container).name() << " at index " << std::distance(begin, it) << " in container " << &container << std::endl;`

[dbNULL.txt](https://github.com/LoneGazebo/Community-Patch-DLL/files/11402589/dbNULL.txt)
[dbNULL-FNV1a.txt](https://github.com/LoneGazebo/Community-Patch-DLL/files/11402588/dbNULL-FNV1a.txt)

PR #9825, opened 2023-05-06
Road AI improvements

KungCheops:
AI now much more likely to build road tiles that are needed for several routes first (make value calculation additive instead of just taking the max value).

AI should now be better at repairing important pillaged routes.

Modify IsBorderLand function to only return true for tiles that are bordering or are close to an unfriendly civ. This seems to be the intended behavior of the function.

Align strategic route plotting between CvCityConnections and CvBuilderTaskingAI.

AI can now repair routes in cities they are razing.

AI no longer wants to build strategic routes in cities they are razing.


KungCheops:
Only change that I'm not 100% sure on is the change to CvPlot::IsBorderLand, since that looks like it changes some AI strategic planning. But the new implementation looks more aligned with how it was intended to work.

PR #9839, opened 2023-05-09
'optimize' button, DoMinorBullyAnnex, DoMinorBuyout MP fix

seroperson:
Allows to use Rome's UA in MP (tested), allows to use 'optimize' button (tested), allows to use DoMinorBuyout in MP (wasn't tested). Also adds desynced variable name in `net_message_debug.log`.


seroperson:
Oh. I see Rome's UA was reworked. Well, functions which were used are mp-synchronized now 🌚

Issue #9864, opened 2023-05-21
Multiplayer Stability Test: 200TurnsGameTest


Cyclopia31:
_1. Mod Version (X.Y.Z). Current Version: 3.5.3_
Vox Populi 3.5.4

_2. DirectX Version_

DirectX 9

_3. List of Other Mods_

None, Base game modpack
Map: Communitu_79a v2.6.0, without Events and with Chill Barbarians.

_4. Describe the Issue_
Hello, I tested the multiplayer stability. I found a problem using the Inca. I also did a 200 turns game without The Inca and Desyncs are really rare, but there are some CTDs for the non-host player.

_A. Game A with The Inca_
1: I think I have found an easily repeatable Desync, when the Inca are chosen randomly or by a player the game will most of the time desync turn 0, the map is generated differently between both human players: after the desync the ressources change for the non-host player and its Capital become a "false capital" --> the city has a grayed star and the policies that specify "Capital" doesn't work for him for example.
Because this desync happens only when the Inca are here I am pretty sure it is link to the "new" Mountain Bias they have.

_B. Game B_200turns_
I have also played a game for 200 turns with 2 human players et 4 AI players and kept the logs/cache/saves/CvMiniDump.dmp for every CTDs/Desyncs for both player.

The good new: In 200 turns game I only had one Desync (I mean load screen) at turn 180, it was a harmless desync because I didn't see something broke because of it.

The bad new: Non-host player had a total of 6 CTDs, 0 for the Host player. All the CTDs were always during the AI turn. Reload the last Autosave worked well and both players could continue the game normally. 
List of CTDs (Non-host player only):
1- Turn 105-106
2-Turn 121-122
3- Turn 141-142
4- Turn 155-156
5- Turn 164-165
(Desync Turn 180)
6 Turn 194-195

I hope it can help. I think if the CTDs problem can be solve the multiplayer is totally playable, with 2 humans players + AI at least.

_C. Data_
Here a link of the logs/cache/saves/CvMiniDump.dmp for both player, for each CTD&Desync, because of the cache the .zip size is 574 mo, you can download it on this link: [https://drive.google.com/file/d/1Uoj8pQAN_lAOJoGgTOvaBXqdsHoFdPjQ/view?usp=share_link](url)

seroperson:
Yea, we were encountering turn 0 desync too. I'm not sure if Inca is the rootcause because we had it even without Incas. I think there is some randomness in map generator.

[...]

PR #9874, opened 2023-05-24
Fixing turn 0 desync / CTD if certain civs are ingame

seroperson:
See ([#9864](https://github.com/LoneGazebo/Community-Patch-DLL/issues/9864#issuecomment-1561509865))


azum4roll:
It's good. Individual start placement doesn't take fertility into account (valuates surrounding plot quality instead), so this won't affect anything.

PR #9876, opened 2023-05-25
fix "cannot attack in ocean" always showing up in civilopedia

azum4roll:


azum4roll:
todo: investigate why vanilla boolean columns can stay boolean in lua while new boolean columns become integers

PR #9958, opened 2023-06-25
Do not send network events while processing AI espionage

### LINKED ISSUE #9955: Multiplayer CTD after some turns
_1. Mod Version (X.Y.Z). Current Version: 3.7.0_
VP 3.7.0 EUI ModPack : https://drive.google.com/file/d/10OMjl8yVf675CPMN6Xv8s81sZGBLEba9/view?usp=sharing
Communitu_79a v2.6.2, without Events and with Chill Barbarians.

_2. DirectX Version_

DirectX11

_3. List of Other Mods_

None

_4. Describe the Issue_

I have tried to play a multiplayer game starting at Renaissance era with quick speed but after some turns I face a CTD from Non-Host player. When I Load the last autosave I can only play the current turn, finish the turn, AIs can play their turn. But when the new turn start,during the "automatic movement from the camera", the game CTDs for both player or for Non-host player only.

I had the same problem on my previous tentative with the same modpack.

_5. Save Game + Log + CvMiniDump

https://drive.google.com/file/d/1RQPsl54Io7jDNFXcdEWZcnToQCCzDvld/view?usp=sharing

Load the autosave with 2 players, play the first turn (press space key for every unit and select random buildings for the cities), then end the turn.

seroperson:
Fixes #9955. Similar to #9926 but for AI processing.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvCity.cpp
+++ b/CvGameCoreDLL_Expansion2/CvCity.cpp
@@ -6743,7 +6743,7 @@ void CvCity::DoEventChoice(CityEventChoiceTypes eEventChoice, CityEventTypes eCi
 						if (pkEventChoiceInfo->IsApplyEffectToSpyOwner())
 						{
 							if (GET_PLAYER(eSpyOwner).getCapitalCity() != NULL)
-								GET_PLAYER(eSpyOwner).getCapitalCity()->DoEventChoice(eEventChoice, NO_EVENT_CITY, true, -1, getOwner());
+								GET_PLAYER(eSpyOwner).getCapitalCity()->DoEventChoice(eEventChoice, NO_EVENT_CITY, false, -1, getOwner());
 
 							return;
 						}
--- a/CvGameCoreDLL_Expansion2/CvPlayerAI.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlayerAI.cpp
@@ -1038,7 +1038,7 @@ bool CvPlayerAI::AI_DoEspionageEventChoice(CityEventTypes eEvent, int uiSpyIndex
 
 				if (eBestEventChoice != NO_EVENT_CHOICE_CITY)
 				{
-					pCity->DoEventChoice(eBestEventChoice, NO_EVENT_CITY, true, uiSpyIndex, GetID());
+					pCity->DoEventChoice(eBestEventChoice, NO_EVENT_CITY, false, uiSpyIndex, GetID());
 					return true;
 				}
 			}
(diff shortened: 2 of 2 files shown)


CvGameCoreDLL_Expansion2/CvCity.cpp
```diff
@@ -6743,7 +6743,7 @@ void CvCity::DoEventChoice(CityEventChoiceTypes eEventChoice, CityEventTypes eCi
 						if (pkEventChoiceInfo->IsApplyEffectToSpyOwner())
 						{
 							if (GET_PLAYER(eSpyOwner).getCapitalCity() != NULL)
-								GET_PLAYER(eSpyOwner).getCapitalCity()->DoEventChoice(eEventChoice, NO_EVENT_CITY, true, -1, getOwner());
+								GET_PLAYER(eSpyOwner).getCapitalCity()->DoEventChoice(eEventChoice, NO_EVENT_CITY, false, -1, getOwner());
```
seroperson, line 6746: I also changed this one `bSendMsg` to `false` - it is not tied with issue but looks like we don't need to send network events here too. We should send network just when human is pressing some button but not inside some logic which already fires on every end.

seroperson:
Added one more commit with explicitely passing `bSendMsg` as `false` everywhere except of LUA calls. Possibly fixes some possible undiscovered desyncs/CTDs caused by double-event processing. Sometimes it is redundant due to `isHuman()` check [here](https://github.com/LoneGazebo/Community-Patch-DLL/blob/d01968be0d9f93da842e30bedf94aff0a1c1e00d/CvGameCoreDLL_Expansion2/CvCity.cpp#L6606) and [here](https://github.com/LoneGazebo/Community-Patch-DLL/blob/d01968be0d9f93da842e30bedf94aff0a1c1e00d/CvGameCoreDLL_Expansion2/CvPlayer.cpp#L8698) but anyway let it be double-check then.

CvGameCoreDLL_Expansion2/CvCity.cpp
```diff
@@ -6603,7 +6603,7 @@ CvString CvCity::GetDisabledTooltip(CityEventChoiceTypes eChosenEventChoice, int
 }
 void CvCity::DoEventChoice(CityEventChoiceTypes eEventChoice, CityEventTypes eCityEvent, bool bSendMsg, int iEspionageValue, PlayerTypes eSpyOwner)
 {
-	if (GC.getGame().isNetworkMultiPlayer() && bSendMsg && GET_PLAYER(getOwner()).isHuman()) {
+	if (GC.getGame().isNetworkMultiPlayer() && bSendMsg && (GET_PLAYER(getOwner()).isHuman() || (eSpyOwner != NO_PLAYER && GET_PLAYER(eSpyOwner).isHuman()))) {
```
seroperson, line 6606: Also added this missing check. Also not tied with the issue. Looks like when you are sending spy to some AI city such events are not broadcasted at all because initial check `GET_PLAYER(getOwner()).isHuman()` is false.

PR #9979, opened 2023-07-06
WIP Some ACTIVE_DIPLOMACY fixes

### LINKED ISSUE #9975: 3.7.5_Multiplayer test Modern Era
_1. Mod Version (X.Y.Z). Current Version: 3.7.5_

VP3.5.7

Modpack here: 
https://drive.google.com/file/d/1BnUJM0tUuZ9dShPv1mzWdYeipdifVzo5/view?usp=sharing

_2. DirectX Version_

DirectX11

_3. List of Other Mods_

None

_4. Describe the Issue_

This patch is mighty stable.
I played a game with 2 players and 6 AI, starting at modern era and quick speed.

In ~80 turns I got one CTD (memory related so nothing wrong here), and one force resync. Looking at the logs message it seems that a problem related to "m_aiLongestPotentialTradeRoute" happened a lot.

Another thing: Non-host player received a lot of deals from AI that are automatically withdraw. It never happened with Host player.

_5. Data

https://drive.google.com/file/d/1-hNzpF4hyx_IfLJSzHPjM702auW_ZEUT/view?usp=sharing

seroperson:
Should fix #9975. I'm not sure what exacly causes desync there because I was unable to reproduce it somewhy. But there is some flaws in MOD_ACTIVE_DIPLOMACY which causes several reproducable weird behaviours (and possibly they cause desyncs somehow):

Like example, host turn 227:
- Mongolia sends renew deal to the Netherlands
- Netherlands receives notification
- But when clicking on notification it disappears and deal withdrawn

It was caused by wrong logic in `CvDealAI::DoEqualizeDealWithHuman`. Maybe @ilteroi may correct me if I'm wrong in decisions: so, there were `pDeal` and `pCounterDeal` variables. Looks like `pCounterDeal` was intended to not modify initial deal and to do all the things on `pCounterDeal` and then make decision on it. BUT idk why they are used simultaneously: in one place we are using `pDeal`, then using `pCounterDeal`, then again `pDeal`. That's why there was situation when we modified `pCounterDeal` to be acceptable and then we were looking at initial unmodified deal `pDeal` again and returning wrong result. So, I fixed this behavior and now it runs ok.

Next example, host turn 228:
- Ethiopia sends new luxury trade to the Netherlands
- Netherlands receives notification
- But when clicking on notification it disappears and deal withdrawn

Here we have completely different rootcause: somehow `CvDealClasses::m_ProposedDeals` contains invalid and outdated deals. Maybe it is intended to be so, but that's why when we are inside `CvDiplomacyRequests::ActivateNext` (callback for diplomacy notification button), we are receiving some outdated deal instead of the last and valid one. I rewrote `CvGameDeals::RemoveProposedDeal` to ignore and remove invalid entries instead of just returning them. Also I changed `CvDiplomacyRequests::ActivateNext` to request latest proposed deal instead of the first one.

Should affect just only MOD_ACTIVE_DIPLOMACY.


seroperson:
Oh, I just realized that looks like we can clean `m_ProposedDeals` in the beginning of turn only in case of simultaneous turns.

seroperson:
Ok, so I revised the PR again and now there are the following changes:
- `CvDealAI::DoEqualizeDealWithHuman`, `CvDealAI::DoEqualizeDealWithAI` changes: get rid of `GetTempDeal`/`SetTempDeal` usage
- `CvDiplomacyRequests::CheckRemainingNotifications` changes: retrieve the last deal from `m_ProposedDeals` (the most up-to-date)
- `CvDiplomacyAI::DoTurn` changes: clear all proposed deals to/from AI in the beginning of their turn. It is a sanitize cleanup because somewhy `m_ProposedDeals` beign polluted with outdated / invalid deals. I think there is some troubles with proposed deals removal, but right now it is hard to find out the initial reason. That's how `m_ProposedDeals` looks like without cleanup (ids 1, 0 are humans):

image

PR #10000, opened 2023-07-20
 Fast-Tracking and Bugfixes

TechpriestEnginseer:
* Added Bare Minimum Support for Custom Resolution Scoring

* Added Belief-Tied Religion in Civilopedia (Ex. Celts)

* Fixed a bug in CanTrainCity where it was reporting false positives on Lua due to inconsistent return statements.

* Added Text for Civilopedia

* Added Civilization's Start Bias


CvGameCoreDLL_Expansion2/CvCity.cpp
LessRekkless, line 1: What lines changed in this file? From what to what?

PR #10005, opened 2023-07-23
Fix Vassal Overview display bug

mwhiter:
The Vassal Overview was displaying the number of turns left for capitulation when a human vassal visits the Vassalage Overview. This corrects this bug and makes it correctly display the number of turns left for voluntary vassalage (Which is 10 turns).

### DIFF
--- a/(2) Vox Populi/C4DF/LUA/VassalageOverview.lua
+++ b/(2) Vox Populi/C4DF/LUA/VassalageOverview.lua
@@ -299,13 +299,13 @@ function MasterSelected( ePlayer )
 	-- Can't request? Set up tooltip
 	if( not bCanRequestIndependence ) then
 		local iMinimumTurns = 0;
-		if ( g_pTeam:IsVoluntaryVassal(iVassalTeam) ) then
+		if ( g_pTeam:IsVoluntaryVassal(iMasterTeam) ) then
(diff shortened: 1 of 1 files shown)


(2) Vox Populi/C4DF/LUA/VassalageOverview.lua
```diff
@@ -299,13 +299,13 @@ function MasterSelected( ePlayer )
...
 		local iMinimumTurns = 0;
-		if ( g_pTeam:IsVoluntaryVassal(iVassalTeam) ) then
+		if ( g_pTeam:IsVoluntaryVassal(iMasterTeam) ) then
 			iMinimumTurns = Game.GetMinimumVoluntaryVassalTurns();
 		else 
 			iMinimumTurns = Game.GetMinimumVassalTurns();
 		end
 
-		local iNumTurnsIsVassal = g_pTeam:GetNumTurnsIsVassal( iVassalTeam );
+		local iNumTurnsIsVassal = g_pTeam:GetNumTurnsIsVassal();
```
mwhiter, line 308: This function in the DLL doesn't even accept a parameter.

mwhiter:
Pending additional DLL changes as one of the tooltips is built in source code.

PR #10009, opened 2023-07-24
Correct punctuation text error in Slinger Historical Info

1slayerpikmin1:
The en dash (–) and the right single quotation (’) characters do not appear in-game, so I replaced the with the hyphen (-) and the apostrophe ('), like what is used in other Civilopedia entries.


1slayerpikmin1:
I forgot to include the screenshot:

PR #10046, opened 2023-08-05
Sponsored proposals implementation

legenciv:


legenciv:
@RecursiveVision The one I sponsored there lost, axatin sponsored the one that won (his own counterproposal).

PR #10066, opened 2023-08-12
Addressing Issues and Proposal Fixes

TechpriestEnginseer:


reikken (commented):
Oh that's a lot more code than I was going to do. Is this better than just moving the multiplier to CvCity::ChangeJONSCultureStored?

LessRekkless (commented):
**Border Growth Multipliers Interact with Instant Culture**

This is changing something that isn't a bug.

PR #10068, opened 2023-08-12
Add table Trait_YieldFromMinorDemand

axatin:


LessRekkless:
Is this the first instance of a table/column that does a % converter?  The table reads like it should just give raw yields upon any demand from a Minor, as opposed to a percent converter on specifically Heavy Tribute.

LessRekkless:
Following the format for Agribusiness (`Building_YieldFromYieldPercent`) might be a better fit:

`Trait_YieldFromHeavyTributePercent`

LessRekkless:
Okay, this is following from Policy_YieldFromMinorDemand

PR #10072, opened 2023-08-13
(5-23) Heavy Tribute Rework

axatin:


(1) Community Patch/Core Files/Text/CoreText_en_US.xml
```diff
@@ -7198,106 +7204,106 @@
 
 		<!-- Quest Data -->
```
LessRekkless, line 7205: Are these the titles for the notifications? Doesn't keeping the name in the notification help when they all get squashed together?

PR #10076, opened 2023-08-14
ArtDefine_Landmarks SPECIAL fix

LessRekkless:
Feitoria, Moai, Pontoon Bridge all use a LandmarkHandler called "SPECIAL", however, this type doesn't exist in ArtDefine_LandmarkTypes, so errors are created in Database.log.

However, the previous fix (switch all the SPECIAL to SNAPSHOT) seems to have wrecked the proper creation of the Feitoria: the pentagon fort is replaced by a second port.  Instead of changing the type, I added SPECIAl to ArtDefine_LandmarkTypes. Both the Database.log seems to be good, and the Feitoria model seems to be created properly.

If you want to make sure that I am actually fixing something that matter, please create some Feitoria yourself before pulling this commit.  You should see that the pentagon fort appears when the Feitoria is placed on a Resource or away from water (will need to be placed with FireTuner/IGE), but is replaced by a second port model if the improvement is on the cost off of a Resource.


LessRekkless:
Weird that the database didn't seem to actually care about the typo.

LessRekkless:
From a completely clean Database.log to loading Civilization V:
Log BNW 2023-08-20 092017

Loading VP, new lines start on 78:
Log VP 2023-08-20 094518

This is what it looks like in BNW the database:
Database BNW 2023-08-20 091940
and in the updated VP database:
Database VP 2023-08-20 092107

PR #10079, opened 2023-08-14
Fix 25% yields from tribute being given to the wrong policy

axatin:


LessRekkless:
This is a fix for the implementation to proposal 5-23.

Issue #9946, opened 2023-06-21
Mausoleum grants 100gold per age on CP


Tesla101a:
_1. Mod Version (X.Y.Z). Current Version: 3.6.2_

Community patch only 3.6.2

_2. DirectX Version_

11

_3. List of Other Mods_

Ingame Editor v.39

_4. Describe the Issue_

Spending a great person when y ou have the mausoleum grants more gold than it should, seems like 100 gold per age or something like that

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

_8. Steps to reproduce the Issue (Optional)_

spend a GP

_9. Screenshots of the Issue (Optional)_
[masoleum logas.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/11811431/masoleum.logas.zip)
![screenshot mausoleum](https://github.com/LoneGazebo/Community-Patch-DLL/assets/15092772/f74e33e0-3992-47ee-b0cd-85cbc1d84
[Alexander_0055 BC-3175.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/11811435/Alexander_0055.BC-3175.zip)
c3d)

Tesla101a:
I'll do it right now :)

Tesla101a:
CP Only version 3.7.12 and IGE v 39.
![Ancient Era 100G](https://github.com/LoneGazebo/Community-Patch-DLL/assets/15092772/67cfd5f2-56f1-43ef-ace0-6e4a1ee7df69)
![Ancient Era GP](https://github.com/LoneGazebo/Community-Patch-DLL/assets/15092772/c38b27e0-f0d3-445f-8153-812811b1a10f)
![Classical Era GP](https://github.com/LoneGazebo/Community-Patch-DLL/assets/15092772/c2b04eb4-ba31-4c81-9590-9703ad56cd54)
![Classical Era 100G](https://github.com/LoneGazebo/Community-Patch-DLL/assets/15092772/c5049cd9-2416-40bf-a9d9-a6365c78c312)
![Medieval Era GP](https://github.com/LoneGazebo/Community-Patch-DLL/assets/15092772/d169bba3-a07e-467c-a87e-1ce0773b294f)
[...]

PR #10083, opened 2023-08-15
Great Person Expend Yield Fix

TechpriestEnginseer:
Mausoleum grants 100gold per age on Community Patch only fix.

Fixed a bug where Great People Expended way too many yields due to multiple era scaling on the total yield.

https://github.com/LoneGazebo/Community-Patch-DLL/issues/9946

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlayer.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlayer.cpp
@@ -27439,58 +27439,49 @@ void CvPlayer::doInstantYield(InstantYieldType iType, bool bCityFaith, GreatPers
 				}
 				case INSTANT_YIELD_TYPE_GP_USE:
 				{
-					if(pLoopCity->isCapital())
+					// Don't use this case if you aren't using a Great Person
+					ASSERT(eGreatPerson != NO_GREATPERSON);
+
+					if (pLoopCity->isCapital())
 					{
-						int iChange = getYieldGPExpend(eYield);
-						if(eYield == YIELD_GOLD)
-						{
-							iChange += GetGreatPersonExpendGold();
-						}
-						// Community Patch Behavior: Assume all GP Expend Yields scale with Era!
-						// Vanilla Behavior: Except the one that has the Mausoleum of Halicarnassus...
-						if (MOD_BALANCE_VP && bEraScale) 
+						iValue += getYieldGPExpend(eYield);
+						iValue += getGreatPersonExpendedYield(eGreatPerson, eYield);
+					}
+					if (bEraScale)
+					{
+						iValue *= iEra;
+					}
+
+					if (pReligion)
+					{
+						int iChange = 0;
+						// GetYieldFromGPUse (does scale with era + city limiter)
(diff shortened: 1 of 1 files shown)


LessRekkless (changes requested):
There's some wacky stuff going on in the `INSTANT_YIELD_TYPE_GP_USE` case:
 - GetGreatPersonExpendGold() and GetGreatPersonExpendedFaith() are the two Base Game functions in question (they are used by Mausoleum of Halicarnassus and the Reliquary Enhancer Belief, respectively). Neither of these should scale with era in CP.
 - surely the correct option to use is MOD_BALANCE_YIELD_SCALE_ERA? @RecursiveVision, are you planning on getting rid of this mod-option in favour of MOD_BALANCE_VP?
 - Why, in the religion section, is only YIELD_CULTURE scaling with era? According to bug reports, Ceremonial Burial's faith was scaling with era but the culture wasn't. This is because the faith scales with era separately in the YIELD_FAITH section. However, this means that the culture of To the Glory of God scales with era, but the Science and Gold do not.
 - Related to the previous point, why does each individual component scale with era separately?  I understand that the case itself is meant to scale on its own, so let's understand why that's the case:

We have:
  - a section that doesn't care what Great Person it is (but we always know that it's a great person and which one it is). yields do not scale with era explicitly
  - a section that cares about which great person it is. scales with era, and scales the previous section as well.
    - inside, we have a belief that both cares which great person it is, and a belief that doesn't (but is inaccessible if we don't know which great person). Only scales culture with era
  - a section that cares if the yield is faith (but not what great person it is). Scales faith with era, and will scale the previous section if the yield is faith.


general oddities:
- `CityScalerLimiter` is set up so that it returns the highest limiter from *all* of your beliefs. That is, if you have "To the Glory of God" whose max # cities is 20, you can raise that max to 25 by taking Council of Elders.
- The only difference between the `Belief_GreatPersonExpendedYield` table and the `Belief_YieldFromGPUse` is that the first table scales off a specific great person. They both scale with era according to the code. Why, then, is *To the Glory of God*, which triggers off of all great people, wasting space listing all of the great people in the first table instead of using the second table?
- Ceremonial Burial states that the yields scale with era, while To the Glory of God's does not.  (both do ... but only if the chosen yield is culture and/or faith)

Note:
- only one case can be true.  iValue starts at 0 before the switch expression, and is only  increased within its own case.


I propose cutting all of this out.  To the Glory of God should scale with era.  3* 20 (25) cities is a grand total of 60 science/culture/gold per great person, which is worthless at the point when you can actually convert that many cities. It's also bugged that the culture currently scales with era anyway. We can separate out the base game functions and scale them separately.

LessRekkless:
added a PR to the TechpriestEnginseer fork.

azum4roll:
> The only difference between the Belief_GreatPersonExpendedYield table and the Belief_YieldFromGPUse is that the first table scales off a specific great person. They both scale with era according to the code. Why, then, is To the Glory of God, which triggers off of all great people, wasting space listing all of the great people in the first table instead of using the second table?

And I thought it was because only one of them scales while the other doesn't...

PR #10089, opened 2023-08-17
(5-20) Withdraw from Melee

axatin:
Implementation of https://forums.civfanatics.com/threads/5-20-withdraw-from-melee-changes.684424/

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvUnit.cpp
+++ b/CvGameCoreDLL_Expansion2/CvUnit.cpp
@@ -1453,6 +1453,7 @@ void CvUnit::reset(int iID, UnitTypes eUnit, PlayerTypes eOwner, bool bConstruct
 	m_iCombatTimer = 0;
 	m_iCombatFirstStrikes = 0;
 	m_bMovedThisTurn = false;
+	m_bHasWithdrawnThisTurn = false;
 	m_bFortified = false;
 	m_iBlitzCount = 0;
 	m_iAmphibCount = 0;
@@ -15050,6 +15051,7 @@ CvUnit* CvUnit::DoUpgradeTo(UnitTypes eUnitType, bool bFree)
 			pNewUnit->m_iMadeInterceptionCount = m_iMadeInterceptionCount;
 			pNewUnit->m_eActivityType = m_eActivityType;
 			pNewUnit->m_bMovedThisTurn = m_bMovedThisTurn;
+			pNewUnit->m_bHasWithdrawnThisTurn = m_bHasWithdrawnThisTurn;
 			pNewUnit->m_bFortified = m_bFortified;
 		}
 		else 
@@ -17158,7 +17160,7 @@ int CvUnit::GetMaxAttackStrength(const CvPlot* pFromPlot, const CvPlot* pToPlot,
 			iModifier += attackBelow50HealthModifier();
 
 		//Heavy charge without escape
-		if (IsCanHeavyCharge() && !pDefender->CanFallBack(*this, false))
+		if (IsCanHeavyCharge() && pDefender->GetNumFallBackPlotsAvailable(*this)==0)
 		{
 			if (MOD_ATTRITION)
 				iModifier += 25;
@@ -17896,7 +17898,7 @@ int CvUnit::GetRangeCombatDamage(const CvUnit* pDefender, const CvCity* pCity, b
 		0 ) / 100;
 
 	//extra damage with special promotion
-	if (GetMoraleBreakChance() != 0 && pDefender && !pDefender->CanFallBack(*this,false))
+	if (GetMoraleBreakChance() != 0 && pDefender && pDefender->GetNumFallBackPlotsAvailable(*this) == 0)
 		iDamage = (iDamage * 150) / 100;
 
 	return iDamage;
@@ -20404,10 +20406,10 @@ void CvUnit::setXY(int iX, int iY, bool bGroup, bool bUpdate, bool bShow, bool b
 										bool bDoCapture = false;
 #if defined(MOD_BALANCE_CORE)
 										bool bDoEvade = false;
-										if (pLoopUnit->IsCivilianUnit() && pLoopUnit->getExtraWithdrawal() > 0 && pLoopUnit->CanFallBack(*this, true))
+										if (pLoopUnit->IsCivilianUnit() && pLoopUnit->CheckWithdrawal(*this))
 										{
 											bDoEvade = true;
-											pLoopUnit->DoFallBack(*this);
+											pLoopUnit->DoFallBack(*this, true);
 
 											CvNotifications* pNotification = GET_PLAYER(pLoopUnit->getOwner()).GetNotifications();
 											if (pNotification)
@@ -22226,6 +22228,18 @@ void CvUnit::SetFortified(bool bValue)
 		triggerFortifyAnimation(bValue);
 		m_bFortified = bValue;
 	}
+}
+
+bool CvUnit::getHasWithdrawnThisTurn() const
+{
+	VALIDATE_OBJECT
+		return m_bHasWithdrawnThisTurn;
+}
+
+void CvUnit::setHasWithdrawnThisTurn(bool bNewValue)
+{
+	VALIDATE_OBJECT
+		m_bHasWithdrawnThisTurn = bNewValue;
 }
 
 int CvUnit::DoAdjacentPlotDamage(CvPlot* pWhere, int iValue, const char* chTextKey)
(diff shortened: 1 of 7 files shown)


CvGameCoreDLL_Expansion2/CvUnit.cpp
```diff
@@ -31260,73 +31259,160 @@ int CvUnit::GetWithdrawChance(const CvUnit& attacker, const bool bCheckChances)
...
+int CvUnit::GetWithdrawChance(const CvUnit& attacker) const
+{
+	VALIDATE_OBJECT
+	int iWithdrawChance = getExtraWithdrawal();
+	if (iWithdrawChance == 0)
 		return 0;
-	
-	if (bCheckChances)
+
+	if (GetNumFallBackPlotsAvailable(attacker) == 0)
```
LessRekkless, line 31276: Given the work performed in `GetNumFallbackPlotsAvailable()`, this should be set to a variable, so it can be used without rerunning it again on line 31291.

CvGameCoreDLL_Expansion2/CvUnit.cpp
```diff
@@ -20404,10 +20406,10 @@ void CvUnit::setXY(int iX, int iY, bool bGroup, bool bUpdate, bool bShow, bool b
 										bool bDoCapture = false;
 #if defined(MOD_BALANCE_CORE)
 										bool bDoEvade = false;
-										if (pLoopUnit->IsCivilianUnit() && pLoopUnit->getExtraWithdrawal() > 0 && pLoopUnit->CanFallBack(*this, true))
+										if (pLoopUnit->IsCivilianUnit() && pLoopUnit->CheckWithdrawal(*this))
```
LessRekkless, line 20409: The situation relating to this line of code deals with a military unit trying to capture/kill an enemy civilian unit by entering its square.

If the will of VP is to always have units with withdraw chance work 100% the first time, then it should work the same way here.

CvGameCoreDLL_Expansion2/CvUnit.cpp
```diff
@@ -20404,10 +20406,10 @@ void CvUnit::setXY(int iX, int iY, bool bGroup, bool bUpdate, bool bShow, bool b
...
 										{
 											bDoEvade = true;
-											pLoopUnit->DoFallBack(*this);
+											pLoopUnit->DoFallBack(*this, true);
```
LessRekkless, line 20412: especially since you've added the withdraw flag here.

CvGameCoreDLL_Expansion2/CvUnit.cpp
axatin: It does work the same way. In MOD_BALANCE_VP, `CheckWithdrawal` always returns true on the first attack and false on subsequent ones.

CvGameCoreDLL_Expansion2/CvUnit.cpp
axatin: Done.

CvGameCoreDLL_Expansion2/CvUnit.cpp
LessRekkless: Right; missed the `GetWithdrawChance()` inside `CheckWithdrawal()`

CvGameCoreDLL_Expansion2/CvUnit.cpp
```diff
@@ -31260,73 +31259,160 @@ int CvUnit::GetWithdrawChance(const CvUnit& attacker, const bool bCheckChances)
...
+		}
 
-		if(pDestPlot && canMoveInto(*pDestPlot, MOVEFLAG_DESTINATION|MOVEFLAG_NO_EMBARK) && isNativeDomain(pDestPlot))
+		if (aValidPlotList.size() == 0)
+			return false;
+		else if (aValidPlotList.size() == 1)
+			pDestPlot = aValidPlotList[0];
+		else
 		{
+			// if there's more than one valid plot, select the plot with the fewest number of enemies
```
LessRekkless, line 31342: Possible optimization for lines 31337-31372
```c++
// if there's more than one valid plot, select the plot with the fewest number of enemies
multimap<int, CvPlot*> aNumEnemies;
for (vector<CvPlot*>::iterator it = aValidPlotList.begin(); it != aValidPlotList.end(); ++it)
{
	aNumEnemies.insert( make_pair( (*it)->GetNumEnemyUnitsAdjacent(getTeam(), getDomainType()), (*it) ) );
}
// remove plots that have more enemies than the minimum
multimap<int, CvPlot*>::iterator itRemove = aNumEnemies.lower_bound(aNumEnemies.begin()->first + 1);
aNumEnemies.erase(itRemove, aNumEnemies.end());
		
if (aNumEnemies.size() == 1)
	pDestPlot = aNumEnemies.begin()->second;
else
{
	// if there is still more than one possible plot, select the plot with the highest number of adjacent friendly units
	multimap<int, CvPlot*> aNumFriendlies;
	for (multimap<int, CvPlot*>::iterator it = aNumEnemies.begin(); it != aNumEnemies.end(); ++it)
	{
		aNumFriendlies.insert( make_pair( it->second->GetNumFriendlyUnitsAdjacent(getTeam(), getDomainType(), true, this), it->second ) );
	}
	// remove plots that have fewer friendlies than the maximum
	itRemove = aNumFriendlies.lower_bound(aNumFriendlies.rbegin()->first); // highest key is the first key in reverse order
	aNumFriendlies.erase(aNumFriendlies.begin(), itRemove);
			
	// make a random selection from the remaining plots
	multimap<int, CvPlot*>::iterator itChosenPlot = aNumFriendlies.begin();
	if (aNumFriendlies.size() > 1)
		advance(itChosenPlot, GC.getGame().getSmallFakeRandNum(aNumFriendlies.size(), plot()->GetPlotIndex() + GetID() + getDamage()) );
	pDestPlot = itChosenPlot->second;
}
```
Multimaps are powerful when you're selecting by key; especially in this instance, we only want the smallest or largest key, which are easily accessible. In this way we can minimize push_backs, for-loops and if statements.

I dunno, it may be too messy, but I had fun figuring it out.  Don't know if multimaps need to be std::multimap in this file.

tested using cpp.sh (with the C++98 compiler) and a dummy CvPlot class

axatin: nice! I wasn't really happy with my implementation either

CvGameCoreDLL_Expansion2/CvUnit.cpp
```diff
@@ -31260,73 +31259,160 @@ int CvUnit::GetWithdrawChance(const CvUnit& attacker, const bool bCheckChances)
...
 
-	int iRightOrLeftBias = (GC.getGame().getSmallFakeRandNum(10, plot()->GetPlotIndex()+GetID()+getDamage()) < 5) ? 1 : -1;
-	int iBiases[3] = {0,-1,1};
-	
-	for(int i = 0; i < 3; i++)
-	{
-		int iMovementDirection = (NUM_DIRECTION_TYPES + eAttackDirection + (iBiases[i] * iRightOrLeftBias)) % NUM_DIRECTION_TYPES;
-		CvPlot* pDestPlot = plotDirection(getX(), getY(), (DirectionTypes) iMovementDirection);
+	CvPlot* pDestPlot = NULL;
+	if (MOD_BALANCE_VP)
```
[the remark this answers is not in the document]
axatin: The CP players might be confused by the fact that units would no longer retreat to the tile opposite to the attacker whenever possible.

CvGameCoreDLL_Expansion2/CvUnit.cpp
```diff
@@ -31260,73 +31259,160 @@ int CvUnit::GetWithdrawChance(const CvUnit& attacker, const bool bCheckChances)
...
 		return 0;
-	
-	if (bCheckChances)
+
+	if (GetNumFallBackPlotsAvailable(attacker) == 0)
+		return 0;	
+
+	if (MOD_BALANCE_VP)
+	{
+		return getHasWithdrawnThisTurn() ? 0 : 100;
```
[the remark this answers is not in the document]
axatin: good idea, I changed it

CvGameCoreDLL_Expansion2/CvUnit.cpp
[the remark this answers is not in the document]
LessRekkless: They would already confused by the fact that units are never able to retreat to free tiles towards the attacker as they can in the base game.

I think this improvement in functionality is worth the potential confusion; especially since it's only people who know that this is the case are those who have looked into the code themselves.

CvGameCoreDLL_Expansion2/CvUnit.cpp
LessRekkless: You may have already done this, but please make sure that the code functions properly inside the civ dll itself.

axatin: I tested it before making the commit, the code works.

PR #10090, opened 2023-08-17
Rationalism

axatin:
- Fix City Governor not taking into account additional yields for non-Specialists from Policies
- Implementation of https://forums.civfanatics.com/threads/5-37-rationalism-reordering.684492/

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvCity.cpp
+++ b/CvGameCoreDLL_Expansion2/CvCity.cpp
@@ -17551,7 +17551,7 @@ int CvCity::foodConsumptionNonSpecialistTimes100() const
 
 	int iFoodPerPop = /*2*/ GD_INT_GET(FOOD_CONSUMPTION_PER_POPULATION) * 100;
 	iFoodPerPop += GetAdditionalFood() * 100;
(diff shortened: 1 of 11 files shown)


CvGameCoreDLL_Expansion2/CvCity.cpp
```diff
@@ -17551,7 +17551,7 @@ int CvCity::foodConsumptionNonSpecialistTimes100() const
 
 	int iFoodPerPop = /*2*/ GD_INT_GET(FOOD_CONSUMPTION_PER_POPULATION) * 100;
 	iFoodPerPop += GetAdditionalFood() * 100;
-	iFoodPerPop += GET_PLAYER(getOwner()).GetPlayerTraits()->GetNonSpecialistFoodChange();
+	iFoodPerPop += GET_PLAYER(getOwner()).GetNonSpecialistFoodChange() * 100 + GET_PLAYER(getOwner()).GetPlayerTraits()->GetNonSpecialistFoodChange() * 100;
```
LessRekkless, line 17554: Looks like you added a multiplier to the Player Trait as well; the trait no longer needs to be times 100? Do we know any modmods that use this trait?

axatin: You're right, it's used in a modmod with a decimal value. I reverted the change, and also adapted the minimum threshold

PR #10107, opened 2023-08-21
Some fixes

axatin:
- City governor assigns less specialists now
- City governor more likely to work multiple specialists of the same type
- Fix an error in war weariness calculation
- Fix military supply cap changing randomly


axatin:
I _think_ the bug in CalculateWarWeariness caused war weariness to be half of the intended value all the time. Because later in the code, there are these lines
```
// Signed peace last turn? Halve value for immediate relief.
	// If we eliminate another player, this won't apply!
	if (iLeastPeaceTurns <= 1)
	{
		iCurrentWarWeariness /= 2;
		bRecentPeace = true;
	}
```
and the number of Peace Turns is always 0 when comparing a player against himself (because no value has ever been actually calculated).

axatin:
and probably `ComputeWarWeariness` should also be called whenever a peace is made or a war is declared (for "immediate relief"), but I'm not familiar enough with the code to know where exactly to put it

PR #10114, opened 2023-08-23
finish off tile picker enhancement

azum4roll:
now factor in religion, policy, building yield changes (plus anything else that isn't already shown on the tile that I missed)


azum4roll:
Does NOT factor in any "X yields from working every Y tiles of this terrain" type of yield changes. Only yields that are directly on the tile.

PR #10148, opened 2023-09-03
DangerPlots and TacticalAI cache invalidation

### LINKED ISSUE #10147: 3.9.1_MultiplayerTest
_1. Mod Version (X.Y.Z). Current Version: 3.9.1_
VP3.9.1 with EUI
Modpack link: https://drive.google.com/file/d/1x8NPHXj91VhPlKgz2NIhoEHdjDOE0NJV/view?usp=sharing

_2. DirectX Version_
DirectX11

_3. List of Other Mods_
3/4 UC, EvenmoreRessources, SuperSettler, (visual) Unit Scaling and Formation for VP (v 12) (Just SQL stuffs).

_4. Describe the Issue_
With a friend we wanted to play a multiplayer game on the new patch.
We faced more issues than usually, expecially a "gamebreaking" desync: A greatprophet became unusuable at turn 227-228 after a force resync. It feels like a Djong transformed itself in a GreatProphet looking at the log ?

We faced two CTDs for the Non-Host player. The first one was repeatable (Turn 142-143) if we didn't deleted our cache. When the cache was deleted for both players we could continue the game normally. 
So I have kept both faulty cache folder to repeat the bug using a VM instead of he non-host computer but, sadly, I couldn't repeat the CTD: the game could continue normally even with the faulty cache :(...  I am not sure if this non-host only CTDs are desync related.

Thank you for your help :)

_5. Data_
Link of Autosaves and logs: https://drive.google.com/file/d/1K35S66kAZNZGXhdmadVb8fEFcH0mntSB/view?usp=sharing

seroperson:
Fixes #10147

Some details are [here](https://github.com/LoneGazebo/Community-Patch-DLL/issues/10147#issuecomment-1704318522) too.

This desync works like so:
- Turn begins, some core logic, ..., Player A starts the turn
- In this case all danger plots caches are invalidated because someone made a peace treaty (`CvPlayerManager::Refresh`, invalidates all caches)
- Player A moves some units
- Player A clicks some UI button, it recalculates danger plots cache for AI
image
- All players end the turn, next turn begins
- On the Player's A end, AI has the updated danger plots cache (because some UI button was clicked) and it works correctly
- Player B didn't click the mentioned UI button, so all danger plots are outdated initially. So, getting the different results for AI logic.

The situation is very similar to trade path cache: using the same cache for UI and for core logic. Possibly we should just invalidate trade path cache here too. Also possibly we should generalize and refactor all the caches to get rid of such desyncs forever, because looks like the logic which exists right now works wrong fundamentally.

P.S. Also made some changes to tactical AI cache, I'm not sure if it fixes something, but still let it be so.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlayer.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlayer.cpp
@@ -35267,8 +35267,13 @@ void CvPlayer::setTurnActive(bool bNewValue, bool bDoTurn) // R: bDoTurn default
 				GetTreasury()->DoUpdateCityConnectionGold();
 
 				//no tactical AI for human, only make sure we have current postures in case we want the AI to take over (debugging)
-				if (isHuman())
+				if (isHuman() || /* if MP, invalidate for AI too */ kGame.isNetworkMultiPlayer()) {
 					GetTacticalAI()->GetTacticalAnalysisMap()->Invalidate();
+				}
+
+				// update danger plots before the turn
+				// causes MP desyncs otherwise (see #10147), affects SP just a little
+				UpdateDangerPlots(false);
 
 				if(kGame.isFinalInitialized())
 				{
--- a/CvGameCoreDLL_Expansion2/CvTacticalAnalysisMap.cpp
+++ b/CvGameCoreDLL_Expansion2/CvTacticalAnalysisMap.cpp
@@ -561,12 +561,17 @@ bool CvTacticalAnalysisMap::IsUpToDate()
 	if (m_iLastUpdate == -1)
 		return false;
 
-	//otherwise consider it up to date if it's not our turn
-	if (m_ePlayer != GC.getGame().getActivePlayer())
-		return true;
+	if (GC.getGame().isNetworkMultiPlayer()) {
+		return (m_iLastUpdate == GC.getGame().getGameTurn());
+	}
+	else {
+		//otherwise consider it up to date if it's not our turn
+		if (m_ePlayer != GC.getGame().getActivePlayer())
+			return true;
 
-	//default check for age
-	return (m_iLastUpdate == GC.getGame().getGameTurn());
+		//default check for age
+		return (m_iLastUpdate == GC.getGame().getGameTurn());
+	}
 }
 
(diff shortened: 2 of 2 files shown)


ilteroi:
1. i don't see anything about danger plots here? is the name misleading or is something missing?
2. tactzones update seems ok, there might be a small performance impact but that should be acceptable
3. we actually have a different pathfinder instance for the UI to avoid conflicts (see CvGlobals::GetPathFinder) ... we could maybe put a similar check in GetTacticalAnalysisMap()?

ilteroi:
ah, i'm on mobile. need to check tomorrow

seroperson:
Yea, we definitely need separate UI/DLL caches for all the caches: TacticalAI, DangerPlots, TradeRoutePath and so on. But I think separate PR suits better for such overhaul changes.

ilteroi:
i'm not sure about this danger plot update - it's already done in DoTurnPostDiplomacy - so are we doing it twice now?

anyway this whole PR might be unnecessary if we have the "correct" fix (ie shadow dangerplots and tactmap instances for UI)

seroperson:
> i'm not sure about this danger plot update - it's already done in DoTurnPostDiplomacy - so are we doing it twice now?

`DoTurnPostDiplomacy` is called **after** player's turn. Therefore player (AI or not, does not matter) begins his turn with danger plot calculated according to previous turn. It is not tied with MP at all, just incorrect logic.

Desync happens when danger plots cache was updated from UI on one side but wasn't on the other.

Thus looks like we need to update it twice: 
- before the player's turn: to make the current player's logic behave correctly according to actual danger plots
- after the player's turn: to make the next players' logic behave correctly if they are depending on others danger plots

> anyway this whole PR might be unnecessary if we have the "correct" fix (ie shadow dangerplots and tactmap instances for UI)

We still need this one change to make current player's logic work correctly, otherwise we will make decisions based on danger plots from previous turn. 

However I think we still need to revise such logic again within UI/DLL cache splitting. The only trouble is that we don't know when it will be implemented :new_moon_with_face: I want to take this one activity, but for now I don't have much time.

Issue #9981, opened 2023-07-06
Irregular CTD


APLion-nsk:
_1. Mod Version (X.Y.Z). Current Version: 3.7.5_

3.7.4

_2. DirectX Version_

_3. List of Other Mods_

https://drive.google.com/file/d/1WVRzbTDeBfQAiXC7xzPuTrBixK6rtFbo/view?usp=sharing
map Tectonic v6 - VI

_4. Describe the Issue_

Every 10 or so turns game crashes during AI turn. Starting the game again fixes the issue for some time
______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[AutoSave_0584 AD-1378.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/11970480/AutoSave_0584.AD-1378.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

[Logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/11970483/Logs.zip)

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

Somehow it's empty
[CvMiniDump.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/11970489/CvMiniDump.zip)

_8. Steps to reproduce the Issue (Optional)_

Unable to reproduce consistently. It just happens pretty often

_9. Screenshots of the Issue (Optional)_

PR #10164, opened 2023-09-12
CvTacticalAI memory improvements

seroperson:


ilteroi:
i fixed the attack cache. don't think this is necessary

seroperson:
> i fixed the attack cache. don't think this is necessary

Can I test your changes with #9981? During profiling I saw that not attackCache is the problem (maybe at least not only), maybe I'm wrong, but still

Issue #10172, opened 2023-09-13
Really weird Building_BuildingClassYieldChanges Behavior


TechpriestEnginseer:
_1. Mod Version (X.Y.Z). Current Version: 3.9.1_
3.9.1

_3. List of Other Mods_
None, just VP needed (+ IGE to replicate asap)

_4. Describe the Issue_
If you have a Factory that adds +2 Production and you suddenly decide to remove it, the Production is still kept!

_8. Steps to reproduce the Issue (Optional)_
1. Build a Factory, sell or make it not appear. The Production is still kept.

_9. Screenshots of the Issue (Optional)_
![image](https://github.com/LoneGazebo/Community-Patch-DLL/assets/13937047/245088e5-ef9f-4316-a339-ccc8f0d79623)
+20 Production
![image](https://github.com/LoneGazebo/Community-Patch-DLL/assets/13937047/1e3d2d7a-a07b-4538-bf14-cf2c40152fe4)
Add Factory.
![image](https://github.com/LoneGazebo/Community-Patch-DLL/assets/13937047/b97bbc16-b2e0-4f28-b3a7-9461159324af)
Remove it again. +24 Production (there's an extra buff of it remaining)?
![image](https://github.com/LoneGazebo/Community-Patch-DLL/assets/13937047/6cbd0c14-ee37-4182-89bc-a20d11e3e5e7)
Readd Factory.
![image](https://github.com/LoneGazebo/Community-Patch-DLL/assets/13937047/b46cd6db-1859-4d86-b387-7772d901d615)
Remove Factory. +28 Production.
![image](https://github.com/LoneGazebo/Community-Patch-DLL/assets/13937047/f916a0e1-1a73-4560-92be-d4d66b31fa70)

PR #10181, opened 2023-09-15
fix for issue #10172

LessRekkless:
A simple way of dealing with trying to remove global yields from a city whose building has just been removed, while iterating through all cities, is to check if the current loop city is the city that has just removed the building.


LessRekkless:
should fix #10172

PR #10250, opened 2023-09-23
Split TradeRoutePathCache into UI and DLL instances

seroperson:
Savegame incompatible (!) because I moved `m_aiLongestPotentialTradeRoute` from `CvCity` to `CvGameTrade`.

Split TradeRoutePathCache into UI and DLL instances to improve MP stability. After all my PRs it still desyncs, but splitting should finally resolve such issues.

Also added some kind of QoL improvement to MP games: when desync happened, desynced player sends a message to the chat. If you do MP stability test, you can do your report immideately after such message appeared, because we already have a desync and we have to investigate it, no sense to play until CTD or something like that. @Cyclopia31
If you play the real MP session, it can be just a signal that something is going wrong now. You may re-create session to restore the host's state to get rid of desync consequences, but no guarantees that it would be not triggered again.
`m_bIsDesynced` variable resets after each turn and new message will appear with next turn if something is still desynced, so maybe spamming is possible in real MP sessions if you just continue the game after a detected desync. We can discuss, maybe it would be better to send such message just once and don't reset the `m_bIsDesynced` variable.
Looks like so (with fake desync at each turn just for an example):

@axatin @RecursiveVision @ilteroi please, check my CPP, because it can be error-prone a little bit 🌚

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvCity.cpp
+++ b/CvGameCoreDLL_Expansion2/CvCity.cpp
@@ -400,7 +400,6 @@ CvCity::CvCity() :
 	, m_aiYieldFromPillage()
 	, m_aiYieldFromPillageGlobal()
 	, m_aiNumTimesAttackedThisTurn()
-	, m_aiLongestPotentialTradeRoute()
 	, m_aiNumProjects()
 	, m_aiYieldFromKnownPantheons()
 	, m_aiGoldenAgeYieldMod()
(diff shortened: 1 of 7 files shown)


n-core:
> Session is running in desynced state, undefined behaviour may appear. Please, fill the issue on GitHub.

Maybe make it a text key? So then people who aren't playing in English can localize it.
`TXT_KEY_MP_WARNING_DESYNC` or something. 

Also, it's better to include the link where you can report it. Not all MP users get VP from the CFC forum or visit VP GitHub. Some of them get VP from their friend.
`Session is running in a desynced state, undefined behavior may appear. Please, fill in the issue on GitHub. (https://github.com/LoneGazebo/Community-Patch-DLL/issues)`

PR #10281, opened 2023-09-30
WIP MP Force Resync button

seroperson:
Still there are some things to do, but I will back for it a little bit later. Until merging I also want to:
- somehow set some appropriate icon for the button. Maybe there are some suggestions which icon can I re-use for such button?
- do some UI tweaks, such as disabling button if desync is already scheduled, some notification on click and so on.
- re-check whether I set all the `#if defined` things. Am I used them properly at all? Do we need to completely cut all the definitions or cutting only bodies is enough? Do we need to use `#if defined` or plain `if`?
- support DX9 binary (I'm unsure about Tablet, I think we can just skip it for now).

Confirmed working like a charm (at least during my brief test session).


CvGameCoreDLLUtil/include/CvEnums.h
```diff
@@ -15,6 +15,16 @@ enum CLOSED_ENUM eCiv5GameCoreMPoolType
 	c_eCiv5GameplayDLL = c_eMPoolTypeUserStart + 0x100,		// Adding 0x100 because we don't want to collide with the app's IDs
 };
 
+#if defined(MOD_EXE_HACKING)
```
Hello71, line 18: this will always be true, you need to use regular if statement if you want to do runtime check

(3a) EUI Compatibility Files/EUI/DC45_Generic.dds
seroperson, line 1: Copied from `UI_bc1`

seroperson:
Pushed new changes:
- Appropriate icon
- Dynamically hide/show, disable/enable button
- Chat notification when desync was scheduled
- Removed `if defined` things, replaced with plain `if`
- Rebase on current master

Seems ready to merge, BUT ... looks like something changed internally since I implemented it and now resync completely breaks the game for non-host players. I'm unsure whether only my force resync is broken or the usual is broken too in v4+, but I assume the second variant.

Here is how it looks like:
broken-resync

(maybe someone knows what's wrong here)

So, delaying the merge until this thing is resolved.

seroperson:
@RecursiveVision Fixed broken resync by removing this `gDLL->BeginSendBundle` part because there is already such logic after `setGameStarted` in `closeInactiveSlots()`. Also I moved `gDLL->sendPlayerInfo(eID)` outside of `slotStatus(eID) == SS_OPEN` block to send all the info to be sure. Maybe it is redundant, but seems like it works and let it be so at least for now, we will back for it in case of any issues.
Tested resync and loading from savefile + resync, should work ok.

Button appears just for EUI DX11 host. Any other configurations are unsupported for now.

PR #10375, opened 2023-10-20
Fix AI not being able to make peace offers

axatin:


axatin:
Second commmit: Fix AI overvaluing luxuries for WLTKD (yield bonuses from religion/policies were calculated as if they were given over the whole deal length, not just for the number of wltkd turns the cities would get)

PR #10426, opened 2023-11-02
QoL City View Enhancements for Building Construction

artemigkh:
This pull request introduces two related quality of life enhancements for EUI city view building construction:

1) Plots with yields that would be boosted by a building are now highlighted when that building is hovered over in the city view building selection list.

Ex. Hovering over arena would highlight tiles with olives, hovering over forge highlights tiles with mines, hydroelectric plant highlights river tiles, etc.

This should make it easier to evaluate how good a building is in a particular city in regards to its plot improvements

2) Buildings in the selection list can now be hidden with the "H" key when hovering over them.

* If at least one building is hidden, a "Unhide Buildings" button will become available to restore them to the selection list

This is intended to clean up production lists for cities where I know I'm never going to want to see a building again, like hotels in a small satellite for example.


artemigkh:
Here's a screenshot with an example of the highlighting of boosted tiles, hovering over Forge here:
image

It's not a very bright or flashy indicator so I don't anticipate much hate for it but I could probably add an option to disable it in ingame settings if you think it's necessary

adaneslavo:
Damn, did you take into account all currently available tables? How did you manage to cover all of that?
Just wow.

artemigkh:
@adaneslavo 

I'm not sure if it uses all the tables for plot yields, but it checks info from the following:
- `Building_ResourceYieldChanges`
- `Building_TerrainYieldChanges`
- `Building_RiverPlotYieldChanges`
- `Building_SeaPlotYieldChanges`
- `Building_SeaResourceYieldChanges`
- `Building_FeatureYieldChanges`
- `Building_ImprovementYieldChangesGlobal`

Which I think covers most cases. If you're interested in implementation details check `CvCity::GetPlotsBoostedByBuilding` in the changed files.

I don't think this will make improved city view incompatible, it's only adding new features. I have no plans to update improved city view but most of the work was done in DLL and exposed via Lua methods so it shouldn't be too hard for any other city view UI to get the same functionality using these same Lua methods

n-core:
Wow...
You actually did implement that fast-tracked proposal (hide a building on city production list).

Apparently you add a custom Lua call for it which is good for keeping your hidden buildings into the memory.

And also highlighting tiles that are affected by a building is a neat feature.

PR #10443, opened 2023-11-07
cleanup

azum4roll:
don't merge before next version is ready


adaneslavo:
Just curious. What making Helper temp table changes? It looks like interesting method I wasn't aware.

PR #10444, opened 2023-11-07
congress 6 proposals

azum4roll:


adaneslavo:
Oho, sound like I will need to update my MW version of asp.lua :)

azum4roll:
Most of it is removing unused variables and clearing trailing whitespace characters.

azum4roll:
I fell down a rabbit hole and now I can't stop cleaning up Communitu_79a.
All of this depends on the Ivory change too, so I can't split the commit.

PR #10471, opened 2023-11-21
additional embarkation sight fixes

azum4roll:


azum4roll:
Even when it's the status quo? I edited the description to say +2 in the merge, when it was +1 in pre-4.0 VP.

Issue #9285, opened 2022-10-12
Feature Request - Include VP compatible version of Planet Simulator (see zip file attachment) with VP install


jayeye2011:
_1. Mod Version (X.Y.Z). Current Version: 2.7.3_

_2. DirectX Version_

_3. List of Other Mods_

_4. Describe the Issue_

As noted in prior issue reports, the changes made to AssignStartingPlots.lua have made some map scripts incompatible with Vox Populli. One popular map script is PlanetSimulator, which for the most part does seem to work except it over-rides the Natural Wonders Placement functions. as noted by azum4roll https://forums.civfanatics.com/threads/communitu_79a-mapscript-vp-version.668265/post-16348995 
It seems like a simple matter to remove the over-ride local functions used in PlanetSimulator.lua and just have it call the ones from AssignStartingPlots.lua. I have attached a modified version of LamilLerran's updated version of Bobert13's PlanetSimulator mapscript that does just that. As far as I can tell looking at the lua.log file, it does place natural wonders using the VP AssignStartingPlot.lua functions, including sea natural wonders.

[Planet Simulator vLL3.7 VP.lua.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/9788581/Planet.Simulator.vLL3.7.VP.lua.zip)

[Lua log.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/9766816/Lua.log.zip)

Edit: Added psparky's random option continents or pangea map preset selection  setting
[Planet Simulator vLL3.7.1 VP.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/9795530/Planet.Simulator.vLL3.7.1.VP.zip)

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

_8. Steps to reproduce the Issue (Optional)_

_9. Screenshots of the Issue (Optional)_

jayeye2011:
Never mind. Realized this is not an appropriate request since I am not the original map script author.

PR #10479, opened 2023-11-25
map stuff

azum4roll:


jayeye2011:
Milae Pangea v6, Milae's Map v14, and Planet Simulator mapscripts all  use "FEATURE_ATOLL".  It is very unlikely that the original Planet Simulator author is going to update the mapscript for this change since he has never done so in the past. Milae doesn't seem to be doing much Vox Populi stuff nowadays either but who knows? At least, could the Community Patch project team add a Vox Populi compatible version of Planet Simulator to Vox Populi as it is one of only a few custom map scripts that can reliably generate a pangea map. While Pangaea, Frontier, Sandstorm, and Milae Pangea can provide a pangea map, I believe Planet Simulator offers its own unique flavor of pangea and is worth retaining as a Vox Populi map option for fans of the mod.

azum4roll:
> Milae Pangea v6, Milae's Map v14, and Planet Simulator mapscripts all use "FEATURE_ATOLL". It is very unlikely that the original Planet Simulator author is going to update the mapscript for this change since he has never done so in the past. Milae doesn't seem to be doing much Vox Populi stuff nowadays either but who knows? At least, could the Community Patch project team add a Vox Populi compatible version of Planet Simulator to Vox Populi as it is one of only a few custom map scripts that can reliably generate a pangea map. While Pangaea, Frontier, Sandstorm, and Milae Pangea can provide a pangea map, I believe Planet Simulator offers its own unique flavor of pangea and is worth retaining as a Vox Populi map option for fans of the mod. Here is a version that works with the prior ASP lua that could be used as a basis for the update. [Planet Simulator vLL3.7.1 VP.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/13535996/Planet.Simulator.vLL3.7.1.VP.zip)

Who updated Planet Simulator last?

jayeye2011:
I did the one in the zip file I provided, just for personal use with Vox Populi. It is a modified version of LamilLerran's from the forum which does not override the ASP functions for resource allocation and city state placement. 
https://forums.civfanatics.com/threads/planet-simulator.526882/post-13240524
As far as I can tell from reading the thread, neither Bobert13 or LamilLerran has ever provided a VP specific version, in fact they seem to be inactive wrt Civ V.

jayeye2011:
It is the same one from https://github.com/LoneGazebo/Community-Patch-DLL/issues/9285

jayeye2011:
Okay, thanks.

PR #10504, opened 2023-12-17
Add AI visibility handling logic and fix some bugs

KungCheops:
Add mechanism for AI to keep track of what plots other AI players can see. Now instead of cheating by checking what another player can see directly, it will now base its knowledge on known enemy territory and unit placements. Currently does not handle the case where an enemy unit moves during your turn (retreating); shouldn't have a big impact overall.

Fix AI sometimes not clearing their route caches if they only have one city.

Fix AI never building improvements near their own capital if they are at war with a civ with a spy there (part of new visibility mechanism).

Possibly fixes AI mistakenly cancelling sneak operations based on bad information.

Fix AI never building lower level roads even when it's supposed to.

Fix vassals removing their master's roads.

Masters now build routes through vassals' lands when needed.

AI now more prone to remove roads when losing gold, rather than the other way round.

Minor cleanup regarding road building and removing. Should fix potential Iroquois and Songhai issues.

Fix incorrect planned road maintenance cost calculation.

Remove unused variables.


KungCheops:
Gonna do some additional testing in the morning but the core seems to be working correctly. Only think I need to verify is Iroquois and Songhai railroad building.

Issue #10521, opened 2023-12-26
The game freezes on CityStates turn (after finding The Barringer Crater) 


brsoft1981:
_1. Mod Version (X.Y.Z). Current Version: 4.3.0_
Vox Populi (with EUI)

_2. DirectX Version_

_3. List of Other Mods_
None

_4. Describe the Issue_
As soon as my Scout discovers The Barringer Crater, the game freezes on the next turn (during CS turn)

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_
I'm attaching autosave and savegame (before and after)
[Saves.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/13771045/Saves.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_
[Logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/13771046/Logs.zip)

_9. Screenshots of the Issue (Optional)_
![Turn9_before_discovering](https://github.com/LoneGazebo/Community-Patch-DLL/assets/96129291/3ebe5036-fcab-4488-9de2-6b770a6384ba)

![Turn9_after_discovering](https://github.com/LoneGazebo/Community-Patch-DLL/assets/96129291/c53343dc-d330-484f-9a2d-b929aeb9c2c2)

AstronomAlpaca:
I've tried a few games now and they all seem to crash at around turn 10, or when meeting a city state (on the CS turn)

SuperNoobCamper:
I came To report the same Issue, using different setups/no mods other than VP does not help either

PR #10523, opened 2023-12-26
Fix CityEventChanges.sql being loaded twice

axatin:


axatin:
Second commit: Fix for https://github.com/LoneGazebo/Community-Patch-DLL/issues/10521, a city-state was trying to give France 170000000 juggernauts because m_iJuggernauts wasn't initialized

PR #10566, opened 2024-01-05
Fix bug related to city state road building quest

KungCheops:
AI will now only build routes from capital to city state.


azum4roll:
Does it take existing roads into account?

KungCheops:
All road building takes existing roads into account, it's part of the pathfinding.

PR #10567, opened 2024-01-05
Bug fixes + small visibility rework

KungCheops:
Fix bug related to city state road building quest

AI will now only build city state quest routes from capital.

Fix Human automated workers never removing obsolete routes.

Update known visibility computation to align with actual visibility calculations. Uses same functionality as is used to calculate line of sight normally. This fixes visibility not being considered for some unit plots. Once again uses a bit more memory but not much.

Move some functions from CvTacticalAI into CvPlot.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvAIOperation.cpp
+++ b/CvGameCoreDLL_Expansion2/CvAIOperation.cpp
@@ -1615,7 +1615,7 @@ bool CvAIOperationMilitary::CheckTransitionToNextStage()
 					int nVisible = 0;
 					for (CvUnit* pUnit = pThisArmy->GetFirstUnit(); pUnit; pUnit = pThisArmy->GetNextUnit(pUnit))
 					{
-						if (GET_PLAYER(m_eOwner).GetTacticalAI()->IsVisibleToPlayer(pUnit->plot(), GET_PLAYER(m_eEnemy).getTeam()))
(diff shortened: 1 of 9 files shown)


ryanmusante:
Human automated workers operate differently than AI automated in most situations?

KungCheops:
> Human automated workers operate differently than AI automated in most situations?

No, that was a rare exception.

Vordikk:
Does this mean that autoworkers will remove my double-triple roads toward enemy lines just because they are "obsolete"?

KungCheops:
Unfortunately, yes. Maybe I could add a flag for human created roads (do not touch)..

Cyclopia31:
> Unfortunately, yes. Maybe I could add a flag for human created roads (do not touch)..

I think there is already an option for that on the settings :)

Vordikk:
There is option to not replace any improvements. I'd like to separate these if possible. Like, there is lumbermill, on which spawns Coal. This one surely needs an upgrade. But since i'm playing without replacing improvements because it was granting many problems in past, this doesn't happen.
Is this option related to *removing* improvements (roads only) too?

KungCheops:
> There is option to not replace any improvements. I'd like to separate these if possible. Like, there is lumbermill, on which spawns Coal. This one surely needs an upgrade. But since i'm playing without replacing improvements because it was granting many problems in past, this doesn't happen. Is this option related to _removing_ improvements (roads only) too?

I could add a check like that for road removal if it's a good compromise.

PR #10608, opened 2024-01-17
4.4.2 fixes

azum4roll:
Map generation testing not complete, don't merge yet


CvGameCoreDLL_Expansion2/CvMap.cpp
```diff
@@ -1152,6 +1152,30 @@ CvArea* CvMap::findBiggestArea(bool bWater)
 }
 
 
+//	--------------------------------------------------------------------------------
+CvLandmass* CvMap::findBiggestLandmass(bool bWater)
```
vaderkos, line 1156: I'm just wondering if return can be cached? 
I'm not sure how this process works but if map is already generated is there a reason to rerun following code when it is called in lua? Maybe this should rerun only if new generation process start?

PR #10611, opened 2024-01-18
Builder AI logic rework

KungCheops:
Change AI worker logic into more of a "top-down" framework. The AI will start each turn by precomputing the value of each improvement on each plot. It will then assign the closest worker for each of the highest priority improvements.

Should have better performance than previous implementation and increases value of high priority plots at the expense of valuing local tiles, although local tiles are still preferred if the evaluation is the same. This should also greatly increase coordination between workers.

Work boats are not included in the rework.

Non-automated human workers and great people will get an assigned job together with the automated workers, which is shown as a recommendation when the unit is selected.

Update route evaluation code to be simpler. Give extra value to routes that are needed for connection to capital. Give extra value to shorter roads (capital roads only).

Merge m_routeNeededPlots and m_routeWantedPlots into one.

All workers will now reevaluate their current best job on each turn. Shouldn't cause any significant performance hit since the heavy computation is done on a "per-civ" basis rather than "per-worker". May cause workers to act seemingly more erratically, but that is because there is a better improvement to build than the one that is currently being built.

Remove "can any enemy see this plot" from the "should worker consider this plot" worker logic. Plot danger check is still done.

Fix poor road pathfinding performance by simplifying computation a lot.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvAStar.cpp
+++ b/CvGameCoreDLL_Expansion2/CvAStar.cpp
@@ -2186,8 +2186,8 @@ int CityConnectionWaterValid(const CvAStarNode* parent, const CvAStarNode* node,
 /// Prefer building routes that can have villages.
 static int BuildRouteVillageBonus(CvPlayer* pPlayer, CvPlot* pPlot, RouteTypes eRouteType, CvBuilderTaskingAI* eBuilderTaskingAi)
 {
-	// If we're not the owner of this plot, bail out
-	if (!pPlot->isOwned() || pPlot->getOwner() != pPlayer->GetID())
+	// If someone else owns this plot, bail out
+	if (pPlot->isOwned() && pPlot->getOwner() != pPlayer->GetID())
 		return 0;
 
 	// Building a village here removes the route in this case
@@ -2198,8 +2198,11 @@ static int BuildRouteVillageBonus(CvPlayer* pPlayer, CvPlot* pPlot, RouteTypes e
 	if (pPlot->getResourceType() != NO_RESOURCE)
 		return 0;
 
-	int iBonus = 0;
+	// No villages for China near cities, no villages for Brazil on jungle/forest, no villages for Netherlands on marshes.
+	if (!eBuilderTaskingAi->MayWantVillageOnPlot(pPlot))
+		return 0;
 
+	// If we have a town or village here, give a big bonus
 	ImprovementTypes eImprovement = pPlot->getImprovementType();
 	if (eImprovement != NO_IMPROVEMENT)
 	{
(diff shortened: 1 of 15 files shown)


KungCheops:
"Merge m_routeNeededPlots and m_routeWantedPlots into one." probably shouldn't be included in any change notes, more of an internal refactoring.

KungCheops:
> "Merge m_routeNeededPlots and m_routeWantedPlots into one." probably shouldn't be included in any change notes, more of an internal refactoring.

Also, this one is save game breaking.

CvGameCoreDLL_Expansion2/CvUnit.cpp
```diff
@@ -3425,16 +3425,32 @@ bool CvUnit::isActionRecommended(int iAction)
...
+
+			if (plot() != pDirectivePlot)
+				continue;
+
+			bool bCanBuild = GET_PLAYER(getOwner()).GetBuilderTaskingAI()->EvaluateBuilder(this, eDirective);
+
+			if (!bCanBuild)
+				continue;
+
+			if (eDirective.m_eBuild == eBuild)
```
vaderkos, line 3446: Is there any reason? Why not? `return eDirective.m_eBuild == eBuild`

KungCheops: Good catch! Previously there was only the "return true" statement :)

KungCheops: Will include simplification is next commit, but waiting to see if there are any minor things that turn up first.

CvGameCoreDLL_Expansion2/CvGame.cpp
```diff
@@ -9668,15 +9668,16 @@ void CvGame::updateMoves()
 
 		if(player.isAlive())
 		{
-			bool needsAIUpdate = player.hasUnitsThatNeedAIUpdate();
-			if(player.isTurnActive() || needsAIUpdate)
+			bool bAutomatedUnitNeedsUpdate = player.hasUnitsThatNeedAIUpdate();
+			bool bHomelandAINeedsUpdate = player.GetHomelandAI()->NeedsUpdate();
+			if(player.isTurnActive() || bAutomatedUnitNeedsUpdate || bHomelandAINeedsUpdate)
```
vaderkos, line 9673: This condition is repeated at least 3 times `|| bAutomatedUnitNeedsUpdate || bHomelandAINeedsUpdate`
Is there some `else` magic below so each if should have it?

KungCheops: This code is a big old mess, it looks like it's pretty much unaltered from the original source code. Looks like there's a bunch of stuff that goes on in the different layers of if/else. From my testing it seems like this function is called once for each AI turn and continuously polled during the player's turn.

KungCheops: For this change I needed to do a little addition to ensure that the homeland AI is updated even when the player has no automated units. (Otherwise worker recommendations won't show up).

Nu-Illasera:
This looks like a big change, personal request : not to commit this pull request in the upcoming version 4.4.3 or 4.5

Reason : There are tons of issues that need to be resolved , some of them include CTD, some include ground-breaking and soft-locking bugs.

I am asking for now, not to introduce any re-working algorithm or new features.

After the release of the upcoming version, i would ask to zone in a bit more on stabilizing things and just bug-solving (Freeze development mode) before the continuation of introducing new features and re-work code infrastructures.

(I know it is volunteer work, but we need a solid infrastructure before we tear things apart to re-work them).

Requesting a delay on this.

KungCheops:
I see where you're coming from. I think this change is pretty much completely isolated from all other changes, and the only thing I haven't tested is multiplayer compatibility, but I believe all of my code should be completely deterministic (assuming that stable_sort is indeed stable :) ).

KungCheops:
I'll leave it up to Recursive if he wants to merge or not. Otherwise I'll keep all these juicy changes to myself (for now) ;)

Nu-Illasera:
> I see where you're coming from. I think this change is pretty much completely isolated from all other changes, and the only thing I haven't tested is multiplayer compatibility, but I believe all of my code should be completely deterministic (assuming that stable_sort is indeed stable :) ).

I am not arguing about the efficiency , i am arguing vs the change as it is a big change.

The general rule is , 1 line of code, 1 bug, is it isolated to just being worker logic re-work, if you say so, i haven't read the code.

Will it not affect other things? YOU CAN'T KNOW THAT, you need unit testing, you need to consider a billion things in such a complicated game, that is a speculation that it won't affect other things, therefor i recommend waiting, I will be the first to support and help you test further, I just wish to wait for a version or two, that's all, nobody is rejecting anything, i welcome this change.

azum4roll:
Current version is already a beta, so there's no harm in packing in more changes. Unless we want a stable version right now.

Every developer works in their own pace so it's hard to tell everyone to "stop working on big changes" or "keep all your big changes to a separate branch".

CvGameCoreDLL_Expansion2/CvAStar.cpp
```diff
@@ -2198,62 +2198,38 @@ static int BuildRouteVillageBonus(CvPlayer* pPlayer, CvPlot* pPlot, RouteTypes e
...
 
+	// If we have a town or village here, give a big bonus
 	ImprovementTypes eImprovement = pPlot->getImprovementType();
 	if (eImprovement != NO_IMPROVEMENT)
 	{
 		CvImprovementEntry* pkImprovementInfo = GC.getImprovementInfo(eImprovement);
+
 		if (pkImprovementInfo != NULL)
 		{
 			for (int iI = 0; iI < NUM_YIELD_TYPES; iI++)
```
ilteroi, line 2217: this still seems expensive - typically it's enough to iterate up to YIELD_TOURISM

azum4roll: Not YIELD_GOLDEN_AGE_POINTS?

KungCheops: Can change, really it was the canBuild call that was causing issues. Changing the loop cutoff might shave off a millisecond or two per turn I suppose.

PR #10627, opened 2024-02-01
UI improvements and bugfixes

axatin:
Bugfixes:
- Reduce number of Barbarians spawning during the CS horde quest (was unintentionally increased in the RNG rework)
- Fix influence decay not triggering an update of CS alliances, which could result in the ally having less influence than another player
- When WLTKD ends in a city and the newly demanded resource is already connected to the trade network, the next WLTKD is entered immediately and not one turn later
- Fix some instant yield notifications appearing one turn too late (instant yields from adopting a policy, removing heresy, buying something with faith)

UI improvements:
- Condense notifications when war is declared or peace is made: Notifications are shown only for the major players, and the notifications contain a list of the vassals and CS allies involved
- It is now mentioned in the notification if a declaration of war is caused by a Defensive Pact
- Add a notification when a foreign city is conquered
- Inquisitors now have a preview of the number of followers after removing heresy in a city
- EUI: Notifications for cultural influence no longer stack (it wasn't possible to cycle through them as clicking on them leads to the tourism screen)
- Fix duplicate notifications when antiquity sites are revealed
- Add a "Close" button to the popup to select a conquered city's fate


Nu-Illasera:
Probably not the best place to write it but perhaps we should "Group" notifications when too many are present, at the cost of the player pressing an extra button.

Example : Having an espionage group with spy related notifications
War related notifications for war related events.
barbarian events.
city states events.

and so on, and pressing on a group of such events will expand the list to enumerate all the related events for the group of notifications. 

*I do not know if other mods are doing as such, i only play CP.

PR #10633, opened 2024-02-03
Some route evaluation fixes

KungCheops:
Ignore route length when evaluating road value for now. Will need a better way to take road length into account.

More intelligent road reuse for capital connection route planning.

Fix some bugs regarding railroad evaluation.


KungCheops:
Hmm seems that they are not building railroads now, gotta continue tomorrow..

PR #10652, opened 2024-02-11
Rework of AI Pantheon Selection

axatin:
AI pantheon selection completely rewritten. AI selects pantheon beliefs based on terrain, existing and potential cities, policy choices, player traits and neighbors. Selection of other beliefs is unchanged.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPolicyAI.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPolicyAI.cpp
@@ -1034,7 +1034,7 @@ Firaxis::Array< int, NUM_YIELD_TYPES > CvPolicyAI::WeightPolicyAttributes(CvPlay
 			iNumCities++;
 		else if (pPlayerTraits->IsSmaller())
 			iNumCities--;
-		if (GC.getGame().getGameTurn() <= 125)
+		if (GC.getGame().getGameTurn() <= GC.getGame().getEstimateEndTurn() / 4)
 		{
 			//Estimation, because early game we assume we'll expand.
 			iNumCities = (3 + iNumCities);
@@ -4729,7 +4729,7 @@ int CvPolicyAI::WeighPolicy(CvPlayer* pPlayer, PolicyTypes ePolicy)
 				}
 				else if (GC.getFlavorTypes((FlavorTypes)iFlavor) == "FLAVOR_EXPANSION" )
 				{
-					iWeight += iFlavorValue;
+						iWeight += iFlavorValue;
 				}
 				else if (GC.getFlavorTypes((FlavorTypes)iFlavor) == "FLAVOR_PRODUCTION")
 				{
--- a/CvGameCoreDLL_Expansion2/CvReligionClasses.cpp
+++ b/CvGameCoreDLL_Expansion2/CvReligionClasses.cpp
@@ -6259,6 +6259,7 @@ BeliefTypes CvReligionAI::ChoosePantheonBelief()
 	std::vector<BeliefTypes> availableBeliefs = pGameReligions->GetAvailablePantheonBeliefs();
 #endif
 
+	CvWeightedVector<int> viPlotWeights = CalculatePlotWeightsForBeliefSelection(/*bConsiderExpansion*/ true);
 	for(std::vector<BeliefTypes>::iterator it = availableBeliefs.begin();
 	        it!= availableBeliefs.end(); ++it)
 	{
@@ -6266,7 +6267,7 @@ BeliefTypes CvReligionAI::ChoosePantheonBelief()
 		CvBeliefEntry* pEntry = m_pBeliefs->GetEntry(eBelief);
 		if(pEntry)
 		{
-			const int iScore = ScoreBelief(pEntry);
+			const int iScore = ScoreBelief(pEntry, viPlotWeights);
 			beliefChoices.push_back(eBelief, iScore);
 		}
 	}
@@ -6299,6 +6300,8 @@ BeliefTypes CvReligionAI::ChooseFounderBelief()
 	std::vector<BeliefTypes> availableBeliefs = pGameReligions->GetAvailableFounderBeliefs();
 #endif
 
+	CvWeightedVector<int> viPlotWeights = CalculatePlotWeightsForBeliefSelection(/*bConsiderExpansion*/ false);
+
 	for(std::vector<BeliefTypes>::iterator it = availableBeliefs.begin();
 	        it!= availableBeliefs.end(); ++it)
 	{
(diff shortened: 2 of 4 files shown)


Nu-Illasera:
Recommending reviewing and testing before committing and to commit next version and not in 4.5, too big of a change.

axatin:
It is tested of course, do you really think I make PRs with larger reworks without testing them?

Explanation of how it works: The AI calculates an estimate of the yields per turn it will get when adopting the pantheon. Each yield type is assigned a value based on AI flavors and general considerations (Faith is more valuable than Gold, etc). In addition, there is an availiability modifier, which is 10 for a yield that‘s granted immediately upon adoption of the pantheon, and lower for yields that will be granted later (or if it‘s uncertain if the yield will ever be granted).

Example: You have a city with 2 owned and worked jungle tiles and 3 jungle tiles adjacent to the city borders, but still within working range. A pantheon gives 1 culture per worked jungle, then the score for this pantheon is (2 * 10 * culture_value) + (3 * 6 * culture_value). The values of 10 and 6 reflect that 2 culture per turn are given immediately, and 3 more potentially later. culture_value depends on the civilization and is higher for culture-focused civs.

Some pantheon effects are not directly translatable into yields per turn, for those other formulas are used. For example, the value of the pantheon giving additional city strength depends mostly on the number of warmongering neighbors.

I don‘t see a problem with adding it into 4.5, it affects only a very specific part of the game and leaves everything else unchanged. Also, a rework of AI belief selection is necessary because it doesn’t work well right now. Check it out, play a game of CP in 4.4.2 (or any version before) and you‘ll notice that civs in desert usually don‘t take the desert pantheon, civs with a jungle start often don‘t take the jungle pantheon, etc.

Nu-Illasera:
Thank you sir.

Some notes : 
1A.)I assume you have tested it; but it doesn't absolve us from errors, if it would, there would be no issues on this github project assuming everyone test their code and even use tools and analyzers to help them. the best of testing is releasing to the mass.

1B.)Currently we are in a state in which i am not sure if @RecursiveVision is away, volunteer work it is , but the accumulation of more and more pull requests and commits without releasing smaller versions more frequent is gonna bite us in the ass on the day of release, He needs to appoint another maintainer that can roll releases (NOT UNOFFICIAL .dll files, but full releases).

2.)From your description, It seems like you are introducing a bias here "Faith is more valuable than gold" (Says who?)

3.)From your description, it seems like short-sighted choice (Its better than before, sure) but what if the city does indeed have some jungle tiles, but everywhere else, there are desert tiles, desert folklore would pay off more (In the long run, but that's how a player would think).
*Maybe it will pay off more if the AI will play into his favor and forget about everything else, Even if you have jungle tiles, if you are a warmonger, go for strength boosting bonuses (Otherwise, you are just reviewing the current game state you are in and make a choice that is good for 1 turn).

4.)does the AI know ahead of time who is a warmonger? are there many warmongers early game on turn 20 or so?
or does it read from some database, as for the leader's traits? does it take into account randomized personality?

I am grateful for the good work and would like to test it further when committed.

axatin:
1a. Errors can happen, true. As to 'releasing to the mass': Yes, that‘s the reason why I want this to be released, so I can get community feedback.

1b. Other devs, including myself, would be able to make releases, there are instructions how to do it. Also, me and others play games based on the current master branch and also AI games are run, so we are doing some testing. 

2. Why should an equal valuation of all yield types be considered the default or  'unbiased'? I don‘t think such a valuation is sensible in Civ 5.

3. As I said, the AI takes future yields into account but values them lower. So, a pantheon giving a high amount of yields in the future may be preferred over one giving a low amount right now. Remember that if it takes too long to get a pantheon online, there is a higher risk that you won‘t be able to found and lose the pantheon bonuses completely eventually.

I would disagree with the second statement. While strategy-based decisions are an important factor, if a terrain pantheon gives a lot of yields, it may be the better choice as those yields help getting a general early advantage that can carry over to the middle/late game. A one-dimensional approach wouldn‘t always be good, and I tried to include different aspects in the total valuation. 

4. It‘s based on which civs have combat-related bonuses (which are also the ones who play aggressively unless random personalities are turned on). If you start right next to one of those, you‘ll know by turn 20. It would be an option to also include policy choices. 

Yes, improving the AI and their strategic and long-term-planning abilities is desirable, unfortunately it‘s easier said than done.

Nu-Illasera:
Thank you for the reply and i do agree with everything you wrote, AS A PLAYER WITH A PLAYER MINDSET.

A player, has a grand scheme, a player has a plan to achieve a grand scheme
in simple words, if a player wants world domination and gets a start with a massive culture boost
the player will know how to translate Culture -> into army units.

An AI will end up diverging from that path (Again, that is unless we will change the AI algorithm to be something like GOAP (that monolith used for FEAR, but it will work here PERFECTLY in a 4X game as well if adjusted for it) , a goal / plan based AI will work your changes.

Without it, i fear that the AI thanks to those boosts, even if it got a personality and preference, will diverge from its original goal, as mentioned above, i do not know if it will be able to translate culture into army if its going for world domination.

A personal opinion that its better to have dumb one-dimensional approach , in a game with 12 AIs, let 11 AIs do stupid crap as long as the 12th AI is giving the player a good challenge, the player will enjoy the game session.

I Will test this and give better feedback (I tried compiling with visual studio 2010 and had some issues, will resolve them soon and give them a go, in the meantime, let's go full speed with releasing this).

Nu-Illasera:
@RecursiveVision Being offensive was not my intention but i don't stop anyone who wishes to be offended.

I did thank the author for taking his time to reply and his hard work and promised to help testing.

There is no "Thank you" because i had no request here, there is no feature request only hypothetical talk.
At times, We need to ask if we are heading towards the right direction and if we are satisfying with it before we pile up more code (That may lead to sunk cost fallacy and endless patching jobs, chasing endless issues to close). There is nothing wrong with holding a discussion with acknowledging the limited amount of resources in our possession (mostly time and will).

azum4roll:
If you care about privacy, you shouldn't put any personal info on a social network, regardless of any privacy policy.

> as this game is symmetrical (Meaning, what you can do , the AI can do

Not yet. We're close to achieving this (way closer than vanilla), but there are still things humans can do that AI can't, and vice versa.

JohnsterID:
@axatin clang warnings:
```
==== C:\Users\user\Documents\GitHub\Community-Patch-DLL\CvGameCoreDLL_Expansion2\CvReligionClasses.cpp ====
C:\Users\user\Documents\GitHub\Community-Patch-DLL\CvGameCoreDLL_Expansion2\CvReligionClasses.cpp(8146,25) : warning: comparison of different enumeration types ('CivilizationTypes' and 'PlayerTypes') [-Wenum-compare]
 8146 |                                                                 if (eRequiredCiv != m_pPlayer->GetID())
      |                                                                     ~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~
C:\Users\user\Documents\GitHub\Community-Patch-DLL\CvGameCoreDLL_Expansion2\CvReligionClasses.cpp(8326,24) : warning: comparison of different enumeration types ('CivilizationTypes' and 'PlayerTypes') [-Wenum-compare]
 8326 |                                                         if (eRequiredCiv != m_pPlayer->GetID())
      |                                                             ~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~
2 warnings generated.
```

PR #10716, opened 2024-03-08
Use static_cast

JohnsterID:
Ran Resharper C++ to check issues, replaced C-style casts and fixed dynamic_cast issue https://github.com/Hello71/Community-Patch-DLL/commit/310e9626620866948a0c32f5782349ad4db58aed. VS and clang build works.


azum4roll:
Should we (all devs) always use static_cast from now on?

JohnsterID:
I would say that the project should move away from C-style casts and use static_cast but devs can jump in with more information. Guidance on programming standards seem to match that. [LLVM Coding Standards - Prefer C++-style casts](https://llvm.org/docs/CodingStandards.html#prefer-c-style-casts)
We shouldn't use dynamic_cast due to https://github.com/Hello71/Community-Patch-DLL/commit/310e9626620866948a0c32f5782349ad4db58aed. [LLVM Coding Standards - Do not use RTTI or Exceptions](https://llvm.org/docs/CodingStandards.html#do-not-use-rtti-or-exceptions)
By slowly using clang-tidy and other tools, we can move towards a set standard.

JohnsterID:
We can put a pin in it and work on other changes. I'll revert the dynamic_cast though.

PR #10739, opened 2024-03-19
Increase compiler heap size

axatin:
Fixes this error that occurs when trying to compile from the current master
compileerror


axatin:
nevermind, duplicate of #10738

PR #10778, opened 2024-03-31
Added LuaCATS declaration files for Locale.

vaderkos:
Added readme how to use LuaCATS.


n-core:
Oh, this is great. This is something that I had in mind while modding Civ 5 Lua files. Scouring back and forth between Modiki and the CPP files is a pain.
Civ 6 has something like this, but as a plugin. I think we should do the same once we add most of the API functions.

The Civ VI Environment Emulation: https://github.com/Wild-W/vscode-civVI-environment-master.git

LuaCATS/Locale.d.lua
```diff
@@ -0,0 +1,275 @@
...
+--- @param textKey string
+--- @param ... string | number
+--- @return string
+function Locale.LookupLanguage(langType, textKey, ...) end
+
+--- ```lua
+--- -- Example
+--- Locale.Compare('a', 'a') -- 0
+--- Locale.Compare('a', 'abc') -- -1
+--- Localge.Compare('abc', 'a') -- 1
```
n-core, line 32: Typo here.

n-core (changes requested):

vaderkos:
> Oh, this is great. This is something that I had in mind while modding Civ 5 Lua files. Scouring back and forth between Modiki and the CPP files is a pain. The Civ 6 community has something like this but as a plugin. I think we should do the same once we add most of the API functions.
> 
> The Civ VI Environment Emulation: https://github.com/Wild-W/vscode-civVI-environment-master.git
> 
> I will contribute some to this because I use the Lua Language Server plugin as well.

Enums are problematic, when they're dynamic XD

LuaCATS/Locale.d.lua
vaderkos: Fixed

azum4roll:
I don't think there are dynamic enums.

vaderkos:
> I don't think there are dynamic enums.

I meant enums that are changes using mods, like extra EndTurnBlockingType for example, I'm unsure how I should document this.

```lua
-- Like this
--- @type table<string, integer | nil>
EndTurnBlockingTypes = {}

-- Or like this
--- @enum EndTurnBlockingType
EndTurnBlockingTypes = {
 -- Here comes default game ones
  --- @type nil | integer
  ENDTURN_BLOCKING_PENDING_DEAL = 999,
  --- @type nil | integer
  ENDTURN_BLOCKING_EVENT_CHOICE  = 777,
}
```

Or just assume they're always not nil

vaderkos:
> @vaderkos Could you explain in simple terms what this does? I'm looking at it and I'm confused. Lua isn't my strong suit.
In short.
It adds intellisense to Locale global table, and include function in lua. 
So you have completion and a little warning when using mismatch types. Doesn't influence in any way how the game or code works. It's like declaration file in TypeScript.

PR #10821, opened 2024-04-13
Draft: Add Squads modmod to be shipped with main mod

artemigkh:
Cleaned up some code, mod rename and new UUID, and added usage instructions in description

image

image

Still requires integration with installer files but I believe this should be everything squads needs to run


(4b) Squads for VP/ThirdParty/PlotIterators.lua
```diff
@@ -0,0 +1,238 @@
...
+  --print(string.format("PlotRingIterator((%i, %i), r=%i, s=%i, d=%s)", pPlot:GetX(), pPlot:GetY(), r, (sector or SECTOR_NORTH), (anticlock and "rev" or "fwd")))
+  -- The important thing to remember with hex-coordinates is that x+y+z = 0
+  -- so we never actually need to store z as we can always calculate it as -(x+y)
+  -- See http://keekerdc.com/2011/03/hexagon-grids-coordinate-systems-and-distance-calculations/
+
+  if (pPlot ~= nil and r > 0) then
+    local hex = ToHexFromGrid({x=pPlot:GetX(), y=pPlot:GetY()})
+    local x, y = hex.x, hex.y
+
+    -- Along the North edge of the hex (x-r, y+r, z) to (x, y+r, z-r)
```
vaderkos, line 54: Can't these function be declared as also side in outer scope, looks like they're pur, so no need to reallocate them each time?

(4b) Squads for VP/UI/SquadsUtil.lua
```diff
@@ -0,0 +1,50 @@
+print("loading SquadsUtil.lua")
+
+function ValidateText(text)
```
vaderkos, line 3: Is this function used, I can't find where

(4b) Squads for VP/ThirdParty/PlotIterators.lua
```diff
@@ -0,0 +1,238 @@
...
+CENTRE_INCLUDE = true
+CENTRE_EXCLUDE = false
+
+function PlotRingIterator(pPlot, r, sector, anticlock)
+  --print(string.format("PlotRingIterator((%i, %i), r=%i, s=%i, d=%s)", pPlot:GetX(), pPlot:GetY(), r, (sector or SECTOR_NORTH), (anticlock and "rev" or "fwd")))
+  -- The important thing to remember with hex-coordinates is that x+y+z = 0
+  -- so we never actually need to store z as we can always calculate it as -(x+y)
+  -- See http://keekerdc.com/2011/03/hexagon-grids-coordinate-systems-and-distance-calculations/
+
+  if (pPlot ~= nil and r > 0) then
```
vaderkos, line 50: Maybe reverse condition and return function that returns nil to reduce nesting, and else won't be needed

(4b) Squads for VP/UI/SquadBases.lua
```diff
@@ -0,0 +1,102 @@
...
+
+function SetSquadBasePlot(squadNumber, newBasePlot)
+    squadBasesLut[squadNumber] = newBasePlot;
+end
+
+function ResetSquadBasePlot(squadNumber)
+    squadBasesLut[squadNumber] = nil
+end
+
+for i=1,10 do
```
vaderkos, line 20: So there is a limit in 10 squads? What if there are more squads?

(4b) Squads for VP/DB/Squads.sql
```diff
@@ -0,0 +1,11 @@
+UPDATE CustomModOptions SET Value = 1 WHERE Name = 'SQUADS';
+
+INSERT INTO InterfaceModes("Type", "Description", "CursorType")
+VALUES ("INTERFACEMODE_SQUAD_UNIT_MANAGEMENT", "TXT_KEY_INTERFACEMODE_SQUAD_UNIT_MANAGEMENT", "CURSOR_DEFAULT");
```
vaderkos, line 4: This can be one insert with values 3 times

(4b) Squads for VP/UI/SquadBases.lua
artemigkh: There's a limit of 10 squads, yes, though this can be extended if people are finding this to be too small a number (I have not heard anyone say this in 2 years).
There should be an equivalent `LoadTable()` call after the local map is initialized though I think perhaps the whole squad bases thing should be removed, don't think anyone uses it  (if this indeed is broken and nobody has spoken up about it that's probably pretty indicative of noone even trying to use it lol)

(4b) Squads for VP/ThirdParty/PlotIterators.lua
artemigkh: everything in `ThirdParty/` is code I didn't write, works exactly as expected, and I have no plans of rewriting any of it :)

vaderkos: Ah, ok :)

PR #10864, opened 2024-05-01
(7-58) Update BuildingTextChanges.sql

jarcast:
(7-58) Rename East India Company to Chartered Company

https://forums.civfanatics.com/threads/7-58-rename-east-india-company-to-chartered-company.689044/

Shall we change also the civilopedia text?
The current text is the following:

> The various East India Companies were joint-stock ventures chartered by the crown to pursue trade in Asia and the islands around the Indian Ocean. Reaching their height during the 17th and 18th centuries, the various companies maintained their own military forces, established settlements under their administration, undertook explorations, fought wars, collected taxes, and built plantations, mines and roads - all in the pursuit of profit. The most successful of these corporations were the "Honourable" East India Company of Britain (chartered by Queen Elizabeth in 1600 AD), the Dutch East Indies Company (1602) and the French Company of the Indies (1664).


azum4roll:
Minor nitpick: we don't need to comment the building name if it's changed on the same file :)

PR #10875, opened 2024-05-03
Icons in descriptions

Hokath:
Completed adding icons for the policy trees that were missing them after Tradition update (Progress, Authority, Imperialism).
Descriptions not otherwise changed.

ICON_VICTORY_X is something I didn't know we had.


azum4roll:
I disagree with adding icons to words like "infrastructure", "warlike", or anything that isn't an actual game component.

Hokath:
I would argue "infrastructure" merits the ICON_WORKER because in other places it means ICON_PRODUCTION (which it doesn't mean here +2 hammers gang)

Putting ICON_WAR on warlike I think is completely justified. Authority benefits from being in a war, no two ways about it. Note not the same as Imperialism which is better at war, but doesn't intrinsically benefit from being at war like Authority +yields do.

PR #10884, opened 2024-05-07
Proposals 7

axatin:


azum4roll (commented):
I thought the spy level change was supposed to be revertible?

axatin:
They are, the database columns for England/Bletchley Park still exist, you can just mod it back in.

azum4roll:
> Also, in (7-20), the Siamese Wat isn't being changed - it should be.

Changing the Constabulary class does it for both buildings.

azum4roll:
> They are, the database columns for England/Bletchley Park still exist, you can just mod it back in.

Ok, I misread and thought some necessary columns/defines were removed.

The UI could be an issue. If you're going to remove the rank icon, what's going to tell the players their spies are stronger?

axatin:
> Changing the Constabulary class does it for both buildings.

Exactly. But I forgot to change the text for the Wat, fixed now.

Having it shown in the UI is difficult, I don't want to have a useless spy level icon in the espionage screen for unmodded VP. Maybe just override EspionageOverview.lua in your mod?

PR #10885, opened 2024-05-08
Add AI support for 'IsNewOwner' improvement property

KungCheops:
Will find the city that will be the owner of the tile and calculate yield gains for that city.

Give a small flat bonus to evaluation for claiming an unowned tile.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvBuilderTaskingAI.cpp
+++ b/CvGameCoreDLL_Expansion2/CvBuilderTaskingAI.cpp
@@ -2639,7 +2639,8 @@ int CvBuilderTaskingAI::ScorePlotBuild(CvPlot* pPlot, ImprovementTypes eImprovem
 
 	// If we are performing a culture bomb, find which city will be owning the plot
 	bool bIsCultureBomb = pkImprovementInfo ? pkImprovementInfo->GetCultureBombRadius() > 0 : false;
-	if (bIsCultureBomb && !pOwningCity)
+	bool bIsTileClaim = pkImprovementInfo ? pkImprovementInfo->IsNewOwner() : false;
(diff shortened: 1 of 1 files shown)


CvGameCoreDLL_Expansion2/CvBuilderTaskingAI.cpp
```diff
@@ -2639,7 +2639,8 @@ int CvBuilderTaskingAI::ScorePlotBuild(CvPlot* pPlot, ImprovementTypes eImprovem
 
 	// If we are performing a culture bomb, find which city will be owning the plot
 	bool bIsCultureBomb = pkImprovementInfo ? pkImprovementInfo->GetCultureBombRadius() > 0 : false;
-	if (bIsCultureBomb && !pOwningCity)
+	bool bIsTileClaim = pkImprovementInfo ? pkImprovementInfo->IsNewOwner() : false;
+	if ((bIsCultureBomb || bIsTileClaim) && !pOwningCity)
```
vaderkos, line 2643: Is `pkImprovementInfo` double check really required?
```suggestion
  bool bIsCultureBomb = false;
  bool bIsTileClaim = false;
  
  if (pkImprovementInfo)
  {
    bIsCultureBomb = pkImprovementInfo->GetCultureBombRadius() > 0;
    bIsTileClaim = pkImprovementInfo->IsNewOwner()
  }
	if (!pOwningCity && (bIsCultureBomb || bIsTileClaim))
```

KungCheops: pkImprovementInfo null check is done quite a few times in this function, could be optimized but don't think it's a very heavy computation :)

PR #10886, opened 2024-05-08
(7-41) Increase Dojo's Yield Conversion from Combat XP

RecursiveVision:
https://forums.civfanatics.com/threads/7-41-increase-dojos-yield-conversion-from-combat-xp.689003


azum4roll:
This makes me really wish for decimal instant yields...

PR #10896, opened 2024-05-11
Worker AI tweaks and fixes

KungCheops:
Make worker AI more bold. Will no longer flee in a stupid manner.

Fixes and tweaks to road planning and evaluation logic.

Consolidate sea and land worker logic.

AI should now use combat units to build improvements when applicable.

Remove Polynesia buff to work boat build speed.


KungCheops:
> @KungCheops But they do still have some sense of self-preservation, right?

Yes, workers used to run away randomly whenever any danger was nearby. Now they will instead just try to avoid dangerous tiles when planning their work.

I've hopefully accounted for all different kinds of civilian units, but I'm fairly sure none of them will try to path through danger. And if they are having trouble they should try to move to safety anyway.

LessRekkless:
Maybe @ilteroi can shed some light if there was anything useful with the function that should have been salvaged.

PR #10914, opened 2024-05-18
Passive Espionage Game Option

axatin:


azum4roll:
NewGameOptionText.xml exists...

But don't worry I'll fix in cleanup.

axatin:
> NewGameOptionText.xml exists...
> 
> But don't worry I'll fix in cleanup.

thanks

Issue #10848, opened 2024-04-27
Memory issues and CTD link and display problems possibily due to memory 


cidk2000:
_1. Mod Version (X.Y.Z). Current Version: 4.7.1_

Vox Populi 4.7.1

_2. DirectX Version_

DX11 

_3. List of Other Mods_

None

_4. Describe the Issue_

extra CTDS including the usual science screen now a lot from city screen and not enough memory always late game 

small 6 players 12 city states communitas map (mostly set to random on settings)
______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[AutoSave_0367 AD-1947.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/15138221/AutoSave_0367.AD-1947.zip)

selecting a city to add International space station show Out of Memory on other Buildings 

_6. Logs (ALWAYS ATTACH THESE IF PO

[Logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/15138225/Logs.zip)

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

[CvMiniDump.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/15138227/CvMiniDump.zip)

_8. Steps to reproduce the Issue (Optional)_

CTD can occur from science screen/city or attempting to load a menu screen to either save 

_9. Screenshots of the Issue (Optional)_
![error](https://github.com/LoneGazebo/Community-Patch-DLL/assets/106168061/4c05be87-1f97-4283-8230-bd604ea23e95) - from older game without logging
![new city](https://github.com/LoneGazebo/Community-Patch-DLL/assets/106168061/2d67a44b-b292-4a33-b03d-c827bea6d922) from older game but without extra items in the build queue
![wierd issue](https://github.com/LoneGazebo/Community-Patch-DLL/assets/106168061/5572ee56-1a90-4c11-9eb1-10817e609ed5) this is from the uploaded latest game before a ctd the recycling centre has been invested and tried to add the international space station to the queue but nothing displays - display error

cidk2000:
additionally highlighting too many tooltips either too quickly or by reading too many can cause the Out of Memory Message on the city screen i didn't hang around on the city screen selecting and now Ive moved turn this has now happened at exactly the same place on both games prev game was on 4.6

cidk2000:
[...]

Issue #10574, opened 2024-01-07
Early CTD after settling with Spain


subversiveasset:
_1. Mod Version (X.Y.Z). Current Version: 4.4.2_

Current Version: 4.4.2_

_2. DirectX Version_

DX 11

_3. List of Other Mods_

(1) Community Patch
(2) Vox Populi
(3a) VP - EUI Compatibility Files
(4a) UI - Promotion Tree for VP
(over) More Unique Components for VP (v 87)
JFD's Cultural Diversity - (1) (Core) Utilities (v 21)
JFD's Cultural Diversity - (2) (Sound) Unit Dialogue (v 1)
JFD's Cultural Diversity - (2) (Art) Splash Screens (v 1)
JFD's Cultural Diversity - (2) (Core) Mod Support (v 1)
JFD's Cultural Diversity - (2) (Gameplay) Starting Bonuses (v 1)
JFD's Cultural Diversity - (2) (Art) Art Styles (v 1)
JFD's Cultural Diversity (Soundtrack)

_4. Describe the Issue_

Sometime after settling the second city as Spain and going into the city screen, the game crashes to desktop. It might be from selecting something to build. but I'm not sure.

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[IsabellaSETTLE_CTD_0034 BC-2640.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/13852299/IsabellaSETTLE_CTD_0034.BC-2640.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

[logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/13852310/logs.zip)

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

[CvMiniDump.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/13852313/CvMiniDump.zip)

_8. Steps to reproduce the Issue (Optional)_

Settle at the place located, browse over build screen options

_9. Screenshots of the Issue (Optional)_

ryanmusante:
Are those JFD mods compatible with VP?

subversiveasset:
They haven't had any problems in the past. I know that's probably not a super rigorous answer.

[...]

Issue #10534, opened 2023-12-27
Game crashing on later eras when hovering the cursor over avaliable buildings in city view


MrMemis:
- Mod Version (4.4.1). 

- Issue description.
The game crashes to desktop on later eras exactly at the moment of hovering the cursor on a buildable structure in the lower left side of the city view. It starts nice in ancient, and as it progresses it hits this problem and becomes unplayable. Asking at the forums i was told, and quote, "It's probably an issue with @L. Vern tile highlighting / building hiding, as those trigger when hovering over a building."

-Logs
[Crash logs.zip](https://github.com/LoneGazebo/Community-Patch-DLL/files/13781976/Crash.logs.zip)

-CvMiniDump.dmp File 
[CvMiniDump.dmp](https://github.com/LoneGazebo/Community-Patch-DLL/files/13781987/CvMiniDump.dmp)

-Steps to reproduce the Issue (Optional)
Just start any game at lets say the Renaissance era and beyond, make a city and start hovering the cursor over the buildings on the left to construct, it will crash at some point. This does not seem to happen early in the game around the ancient era for some reason.
This doesnt seem to crash with units, services, wonders or else, just buildings.

-Screenshots of the Issue (Optional)
<img width="949" alt="civ crash 1" src="https://github.com/LoneGazebo/Community-Patch-DLL/assets/155024713/75639f77-67fe-45c5-b1aa-f7863b558fdc">
<img width="956" alt="civ crash 2" src="https://github.com/LoneGazebo/Community-Patch-DLL/assets/155024713/7bee80b5-fef6-45f8-8f55-742e3c628fe9">
<img width="951" alt="civ crash 3" src="https://github.com/LoneGazebo/Community-Patch-DLL/assets/155024713/31f998e1-39bf-4b88-9ac4-3a8c6e4f5b49">

MrMemis:
Update: so after messing around a bit i loaded IGE and checked on a city, and after that it would seem that the games no longer crash (?), even with the IGE disabled. Is incredibly odd but is how is going so far, will keep testing later.

artemigkh:
I haven't been able to reproduce this, neither with nor without IGE.
The minidump file shows the crash is coming from the exe and not the DLL so it might be a strange manifestation of an out of memory crash, the same way opening the tech tree can kill the game when out of memory

[...]

PR #10923, opened 2024-05-22
Add interface option for disabling city view highlighting of plots buffed by hovered building

artemigkh:
Adds an option for disabling highlighting of plots that would be buffed by a building when hovered over in the construction menu
image

This change should provide a workaround for people experiencing some city view CTDs as it appears that `SerialEventHexHighlight` may cause memory problems

* https://github.com/LoneGazebo/Community-Patch-DLL/issues/10534
* https://github.com/LoneGazebo/Community-Patch-DLL/issues/10574
* https://github.com/LoneGazebo/Community-Patch-DLL/issues/10848
* https://forums.civfanatics.com/threads/new-version-4-7-3-april-28-2024.689478/page-2#post-16598238

### DIFF
--- a/(3a) EUI Compatibility Files/LUA/CityView.lua
+++ b/(3a) EUI Compatibility Files/LUA/CityView.lua
@@ -128,6 +128,8 @@ local YieldTypes = YieldTypes
 local g_options = Modding.OpenUserData( "Enhanced User Interface Options", 1)
 local g_isAdvisor = true
 
+local g_isDisableBuildingBonusHighlights = g_options.GetValue( "DisableBuildingBonusHighlights" ) == 1
+
 local g_isSeparateCityProductionEUI = g_options.GetValue( "SeparateCityProduction" ) == 1
 --print("Separate City Production: "..tostring(g_isSeparateCityProductionEUI))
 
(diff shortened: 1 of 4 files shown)


azum4roll:
It's unlikely to be SerialEventHexHighlight since it's used by worked plots as well and has never caused issues.

artemigkh:
I think it's because of the way it interacts with the existing city hex highlights, as Squads also uses highlights extensively and I've never heard of any problems there

azum4roll:
So technically on EUI if you move your mouse on and off cities quickly it'll also crash.

axatin:
> If it's causing memory crashes, shouldn't we just remove it entirely?

or at least make it an option that’s turned off by default

artemigkh:
I tested it and `SerialEventHexHighlight` also reliably causes crashes outside of city view. Turning off `SerialEventHexHighlight` by default, even though helps for some people (I will note even before the addition of building bonus highlights, people had memory CTDs with EUI that were resolved by removing it, probably in part from the very same plot highlighting functionality) is a solution that is plainly throwing away the baby with the bathwater. Plot highlights provide invaluable information in many facets of the game, and turning them off because a small minority of people have problems is a solution I'm not a fan of.

I have no plans for further changes to this MR; as it stands it still seems like a useful UI option for those that want more control over their display. If it is determined to be too much of a workaround that has no place in the community patch by its maintainers I totally understand, feel free to close it.

artemigkh:
After seeing several more reports of CTDs on building hover, I've changed my mind. Clearly a substantial number of people are consistently crashing on building selection list hover after the functionality I added, I'm closing this ticket and will shortly put up a new MR with it disabled by default and making it opt in instead.

PR #10961, opened 2024-06-09
Treat all civs as having been met if the player is observer

KungCheops:


KungCheops:
Makes it so you can see the scoreboard, who's allied to city-states, etc. when in observer mode.

PR #10999, opened 2024-06-16
clang-tidy

JohnsterID:
Ran clang-tidy. No issues on build.


CvGameCoreDLL_Expansion2/CvTeam.cpp
azum4roll, line 1: But this is missing the variable assignment.

PR #11017, opened 2024-06-21
latest and greatest

ilteroi:
most of the commits below have already been cherry picked ...


ilteroi:
could have squashed it but yeah, works for me

PR #11021, opened 2024-06-22
cppcheck

JohnsterID:
Fix some cppcheck issues. Builds work and mod runs.


azum4roll (commented):
Should we really be modifying libraries?

JohnsterID:
Looking at the commit history of the libs, they have been previously modified. For Fireworks, initial compiler setup, debugging, clang compilation and compiler warnings.

PR #11024, opened 2024-06-22
readd cppcheck runner

JohnsterID:
First runs will be slow to build the cache. Place the artifact cppcheck.xml in local source directory to view in cppcheck.

Could be improved, e.g. second job doesn't run if analysis was completed in first job.

### DIFF
--- a/.github/workflows/cppcheck.yml
+++ b/.github/workflows/cppcheck.yml
@@ -1,0 +1,123 @@
+name: cppcheck
+on:
+  push:
+    branches: [ master ]
+  workflow_dispatch:
+
+concurrency:
+  group: cppcheck-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  analysis:
+    runs-on: ubuntu-22.04
+    env:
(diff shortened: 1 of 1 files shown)


JohnsterID:
@RecursiveVision Enables cppcheck CI so people can easily load and check cppcheck static analysis. It was previously added on https://github.com/LoneGazebo/Community-Patch-DLL/pull/8652 but removed to solve cppcheck not completing.

https://www.thegeekdiary.com/cppcheck-a-static-analysis-tool-for-c-c-code/

JohnsterID:
@RecursiveVision The workflow was using concurrency option to stop current run if a newer run is starting so process never completes if there are heaps of commits. We need at least one run to complete before trying the concurrency option. Fixed https://github.com/LoneGazebo/Community-Patch-DLL/pull/11074.

azum4roll:
The build log is currently extremely large due to the same 36 unused function warnings being repeated for every single file. It's hard to find the actually useful warnings. Is there a setting to ignore those included files?

Issue #9641, opened 2023-01-30
crash when leaving strategic view in observer mode


ilteroi:
i use a lot of autoplay, but there has always been a crash when trying to leave the strategic view during autoplay. but i'd like to use strategic view because it uses less GPU resources. note that there is an option to throttle the game engine when the window loses focus, but that is useless for autoplay, because the game does not progress as fast then - it's not just the renderer that is throttled but also the game core (genius!)

anyway turns out the sequence is as follows:

MiniMapPanel.lua:

```
function OnStrategicView()
	local bIsObserver = PreGame.GetSlotStatus( Game.GetActivePlayer() ) == SlotStatus.SS_OBSERVER;
	if (bIsObserver) then
		-- Observer gets to toggle the world view completely off.
		local eViewType = GetGameViewRenderType();
		if (eViewType == GameViewTypes.GAMEVIEW_NONE) then
			SetGameViewRenderType(GameViewTypes.GAMEVIEW_STANDARD);			
		else
			if (eViewType == GameViewTypes.GAMEVIEW_STANDARD) then
				SetGameViewRenderType(GameViewTypes.GAMEVIEW_STRATEGIC);
			else
				SetGameViewRenderType(GameViewTypes.GAMEVIEW_NONE);
			end
		end
	else
		ToggleStrategicView();
	end		
end
```

InGame.lua

```
-- Check to see if the world view controls should be hidden
function OnGameViewTypeChanged(eNewType)
	local bWorldViewHide = (GetGameViewRenderType() == GameViewTypes.GAMEVIEW_NONE);
	Controls.WorldViewControls:SetHide( bWorldViewHide );
	Controls.StagingRoom:SetHide( not bWorldViewHide );
end
```

finally CvDllGame.cpp

```
void CvDllGame::Uninit()
{
	// There exists a bug that allows the game core to be shutdown early if mods are loaded.
	// Executing `SetGameViewRenderType(GameViewTypes.GAMEVIEW_NONE)` in Lua will trigger this.
	// 
	// Because the first step of the shutdown process deinitializes the game instance, this
	// should trap before the program has a chance to enter a broken state which hopefully
	// protects users from any exploits that could be leveraged while the state is broken.
	if (CvPreGame::gameStarted())
		BUILTIN_TRAP();

	m_pGame->uninit();
}
```

so it's more or less a forced crash, but no idea how we get from GAMEVIEW_NONE to CvDllGame::Uninit()

questions for the experts:
[...]

PR #11030, opened 2024-06-23
Observer mode

axatin:
I added some features to the observer mode with the intention of making it easier for developers to observe and improve AI behavior. Non-developers might also be interested in them. Also fixed a crash in observer mode (#9641).

----

Press Ctrl-Shift-O to enter observer mode:
- In observer mode, the AI will take control over your civ, and the turns of all players are processed automatically.
- The map is completely revealed and the current production of all cities is shown.
- If quick movement is turned off, the camera will switch to the capital of each player at the start of their turn. If quick movement is turned on, the camera will remain at one place.
- If quick combat is turned off, the combat animations for all units on the map are shown.
- The top panel is shown for the player that's currently active (major civs only). Their tech tree and policy panel can be accessed from the top panel
- Pressing Ctrl-Shift-O while in observer mode will give the human player control of the AI civilization whose turn is currently being processed

Press Ctrl-Shift-T to have the AI take control over the civ you're playing:
- The AI will make all decisions for the civ, all turns will be processed automatically
- World map, city banners and top panel continue to be shown from the perspective of the civ you originally controlled
- Note that the AI will play on AI difficulty level
- Press Ctrl-Shift-T to gain back control over your civ (after the other AI player's turns have been processed)

Press Ctrl-Shift-P to switch to the next AI player
- Can only be used when not in observer mode and when it's your turn
- The AI will take control over your civ. You will take control over the AI civ next in the turn order

Of course, all of these are disabled in multiplayer.

Known issues / limitations:
- The minimap is sometimes not shown correctly. Can't be fixed
- Some UI screens, like espionage screen and military overview, don't display anything
- Unrevealed resources are always shown in observer mode
- A game saved in Ctrl-Shift-T mode (does anyone have a good name for this?) will be loaded as a game in Ctrl-Shift-O mode


azum4roll:
> The top panel is shown for the player that's currently active (major civs only). Their tech tree and policy panel can be accessed from the top panel

What happens when active player changes while you're in the menus?

JohnsterID:
Possibility for Ctrl-Shift-T mode for MP testing/debugging? Announcement in the chat if someone turns it on?

axatin:
> What happens when active player changes while you're in the menus? Which is ALL the time since the UI is locked while the AI is processing and there's only 3-5 seconds the UI works for every player's turn (if quick movement/combat are on)

The panels don't update, they keep showing the information for the player that was active when the panel was opened.

> Is it the one that happens when you use debug mode to "Reveal Map" while observer mode is already on (the map is already fully revealed)?

Don't know about that one, but the issue I observed is that the minimap always switches to the lowest zoom level (as if all plots were revealed) and there are visual bugs at the borders of the map. It happens also when using IGE to switch players. I couldn't find a fix for it, minimap generation is not done in the dll.



> Possibility for Ctrl-Shift-T mode for MP testing/debugging? Announcement in the chat if someone turns it on?

Switching from player to observer mode in MP is difficult, it would probably mess with the synchronization. In MP, games can be loaded in observer (Ctrl-Shift-O) mode from the main menu, I'll try adding a hotkey to switch from observer mode to Ctrl-Shift-T mode.

Maybe global observer mode and local observer mode are good names.

PR #11081, opened 2024-07-05
more cleanup stuff

azum4roll:


JohnsterID:
@azum4roll msvc build warning only, not on clang.
https://github.com/LoneGazebo/Community-Patch-DLL/actions/runs/9807219785/job/27080561123
```
1>CvUnit.cpp(20919): warning C4100: 'fAdditionalTextDelay' : unreferenced formal parameter
```

azum4roll:
I know, but that requires a refactor that I couldn't do yet.

JohnsterID:
Understood, I wasn't even looking at the clang runner with W1 enabled as there are too many unused. The normal non-sdk clang script doesn't have W1 so it is clean.

PR #11082, opened 2024-07-05
clang-tidy

JohnsterID:
clang-tidy replacements. Both builds work.


CvGameCoreDLL_Expansion2/CvCity.cpp
```diff
@@ -8487,10 +8487,7 @@ bool CvCity::hasBuildingPrerequisites(BuildingTypes eBuilding) const
 	}
 
 	BuildingTypes ePrereqBuilding = static_cast<BuildingTypes>(pkBuildingInfo->GetNeedBuildingThisCity());
-	if (ePrereqBuilding != NO_BUILDING && !HasBuilding(ePrereqBuilding))
-		return false;
-
-	return true;
+	return !(ePrereqBuilding != NO_BUILDING && !HasBuilding(ePrereqBuilding));
```
azum4roll, line 8490: I don't like this. Really hard to read.

`return (ePrereqBuilding == NO_BUILDING || HasBuilding(ePrereqBuilding));` is better, but for readability and ease of future editing I'd still split these.

PR #11124, opened 2024-07-11
Aloe vera bonus resource placing compatibility

jarcast:
Edit of Communitu_79a.lua and AssignStartingPlots.lua to support my Aloe Vera resource modmod. Changes are similar to Even More Resource modmod. I tested the changes beforehand and they work.


azum4roll:
Done in #11126

PR #11139, opened 2024-07-13
Fix bug with improvement AI and a rare crash

KungCheops:


CvGameCoreDLL_Expansion2/CvMinorCivAI.cpp
```diff
@@ -5696,9 +5696,18 @@ int CvMinorCivAI::GetNumThreateningBarbarians()
 	int iCount = 0;
 
 	int iLoop = 0;
-	for(CvUnit* pLoopUnit = GET_PLAYER(BARBARIAN_PLAYER).firstUnit(&iLoop); NULL != pLoopUnit; pLoopUnit = GET_PLAYER(BARBARIAN_PLAYER).nextUnit(&iLoop))
-		if(pLoopUnit->plot()->isAdjacentTeam(GetPlayer()->getTeam()))
+	for (CvUnit* pLoopUnit = GET_PLAYER(BARBARIAN_PLAYER).firstUnit(&iLoop); NULL != pLoopUnit; pLoopUnit = GET_PLAYER(BARBARIAN_PLAYER).nextUnit(&iLoop))
+	{
+		if (!pLoopUnit)
```
azum4roll, line 5701: This already can't be NULL.

KungCheops: yeah I suppose one of the null checks is unnecessary, but also there shouldn't be any real performance loss I think

KungCheops: the crash was because one barbarian unit was in a non-existent tile

azum4roll: But how? We probably should find the root cause.

PR #11170, opened 2024-07-22
Add improvement golden age extra yields table

KungCheops:
Add improvement golden age extra yields table for 4UC Persia.

Fix some bugs with improvement planning yield calculations.

Normalize golden age yield calculations for improvement planning.

Fix two bugs with vanilla Celts UA improvement planning.


KungCheops:
> Hmm, I _think_ this is savegame compatible.

I can double check during lunch

KungCheops:
I was able to load my save from without the addition without any crash

PR #11177, opened 2024-07-28
Last fixes

axatin:
Last fixes before 4uc


azum4roll:
What's that file and what does it impact?

axatin:
> What's that file and what does it impact?

It‘s used for the forced resync button in multiplayer, it existed already in EUI and I moved it to non-EUI

Issue #11183, opened 2024-07-30
Game becomes unresponsive during AI turn


saamohod:
_1. Mod Version (X.Y.Z). Current Version: 4.14.1_

_3. List of Other Mods_
![20240729152112_1](https://github.com/user-attachments/assets/69438d76-2780-4a97-abe7-67d35bfe2804)

_4. Describe the Issue_
Can't remember this happening before. Not a crash but a freeze. Had to close the game via Ctrl-Shift-Esc.
Tried again from the turn before, pressed End Turn - same thing during AI turn processing.

[1.zip](https://github.com/user-attachments/files/16425884/1.zip)

axatin:
Seems to be some issue on your end, I can continue past the next turn normally.

saamohod:
@axatin 
Any chance of having some sort of a hotfix? I'm afraid the fix will come bundled with a new incompatible version and I will have to abandon this game.
Thank you for your work guys!

axatin:
@saamohod 
Sure, here you go
[CvGameCore_Expansion2.zip](https://github.com/user-attachments/files/16475897/CvGameCore_Expansion2.zip)

PR #11192, opened 2024-08-01
Change function to be non-recursive

KungCheops:
Should fix possible infinite recursion issue/any other possible issue with the function.


KungCheops:
Fixes #11183

PR #11231, opened 2024-08-17
New adjacency with builder AI support

KungCheops:


azum4roll:
What's the point of making any of those functions static (rather than private)? You aren't calling them from outside the class.

KungCheops:
> What's the point of making any of those functions static (rather than private)? You aren't calling them from outside the class.

The functions are static because they don't need to access any of the member variables of a CvBuilderTaskingAI object. Static doesn't mean you can call it outside of the class.

KungCheops:
gonna move this to drafts, since rek is going to be changing the commits this is including

KungCheops:
> @KungCheops Finally merged Rek's commits, you can finish this.

nice, will do tomorrow evening

PR #11232, opened 2024-08-17
NoAdjacentCity column

LessRekkless:
New Improvement Restriction - NoAdjacentCity:
 - forbids an improvement from being built next to a city


KungCheops:
> @KungCheops Is any AI support needed?

No, it should work no problem

PR #11249, opened 2024-08-28
Fix bug Gallery not unlocked by "Faith of the Masses" reformation belief

jarcast:
Gallery was added as culture building but was left out from being purchasable with Faith for mysterious and lazy reasons. This commit intends to fix it although it should have been fixed already for the last 2UC VP release. Better late than never I guess.


azum4roll:
It's intended that both Faith of the Masses and Divine Teachings boost the same number of buildings.

PR #11266, opened 2024-09-05
Fix Villa resources spawning rules

KungCheops:
Resources can now spawn in enemy territory.

Resources now have no preference for owned or unowned tiles.

Resource can no longer spawn on oasis.

Consolidate resource spawning validity logic.


azum4roll:
Isn't it supposed to check canHaveResource?

KungCheops:
> Isn't it supposed to check canHaveResource?

no

azum4roll:
Why though? I'd like something to control the spawns so you don't spawn a fig on snow for example.

PR #11271, opened 2024-09-08
Add table Improvement_YieldPerXAdjacentTerrain

KungCheops:
Includes worker AI support.

Includes civilopedia support.

Update Inca terrace farm as per (8-20).


azum4roll:
Shouldn't have touched any sql... I'll have to rewrite everything anyway.

Btw, has anyone figured out how to use plural forms of text keys directly in other text keys (as opposed to in Lua/C++)?

PR #11309, opened 2024-09-29
Fix CTD

axatin:


azum4roll:
If we had the ID remapper in DLL we would've have to add null checks everywhere...

axatin:
Yes, it only happens if a modmod has deleted rows in the CityEvents table

And yes, having to add null checks everywhere is annoying. It's also bad for performance if the loops go through null entries.

PR #11316, opened 2024-10-07
clang-tidy

JohnsterID:
Ran the below, builds work.

[readability-avoid-return-with-void-value](https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-return-with-void-value.html)
[readability-math-missing-parentheses](https://clang.llvm.org/extra/clang-tidy/checks/readability/math-missing-parentheses.html)
[cppcoreguidelines-pro-type-cstyle-cast](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-cstyle-cast.html)


azum4roll (commented):
I don't like forcing this kind of parentheses. Most of the times it's already clear enough especially when the multiplication/division is already at the start. Having extra layers sometimes just makes it harder to read.

JohnsterID:
Reverted to get this merged. A good IDE should highlight the brackets to make it clear but it would be personal preference and then the project adopting a coding style.

PR #11328, opened 2024-10-17
Fix AI not building route for city state quests

KungCheops:
Plus some minor code cleanup.


Mwallx:
is there still build route quest in vp? seems like I've never experienced it, might be a 4uc issue?

azum4roll:
The CS needs to be near one of your cities and be non-coastal.

KungCheops:
Hmm, this PR doesn't actually fix what I thought it fixed. Must've been some coincidence that caused the AI to start building routes to city states.

PR #11340, opened 2024-10-24
add new tables and columns to the database

azum4roll:
Build check done
Start game check done
Now fixing inevitable Lua errors with all those BOOLEAN columns changing to boolean.


azum4roll:
Test run as observer done, should be safe to merge.

CvGameCoreDLL_Expansion2/Lua/CvLuaUnit.cpp
```diff
@@ -1811,23 +1811,6 @@ int CvLuaUnit::lIsRangeAttackOnlyInDomain(lua_State* L)
 	return 1;
 }
 
-//------------------------------------------------------------------------------
-int CvLuaUnit::lIsCoastalAttackOnly(lua_State* L)
```
KungCheops, line 1815: It's not possible to build the dll after you removed this (forgot to update line 338). Looks like there's a reference to this in Bombardment.lua as well.

azum4roll: It shouldn't affect build, only having error when Bombardment.lua is run (when a human selects a city to attack, which I didn't test).

azum4roll: fixed in latest commit

PR #11344, opened 2024-10-30
fixed issue with self determinism policy

mt-empty:
where, if a city state was liberated before adopting the policy, its effects were not granted to the player.


mt-empty:
not ready

PR #11365, opened 2024-11-16
Use python clang build in Visual Studio

JohnsterID:
Add visual studio config to call the python script. I'm sure we can eventually figure out how to add the clang solution completely to vcxproj to not use the python script.

### DIFF
--- a/CvGameCoreDLL_Expansion2/VoxPopuli.vcxproj
+++ b/CvGameCoreDLL_Expansion2/VoxPopuli.vcxproj
@@ -9,6 +9,14 @@
       <Configuration>Release</Configuration>
       <Platform>Win32</Platform>
     </ProjectConfiguration>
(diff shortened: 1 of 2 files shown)


azum4roll:
Do you know why the MSVC Release build always get stuck in Visual Studio? I presume it's related to Windows 11 since it only started behaving like that after upgrading.

JohnsterID:
Tried building it, slow but finishes. vs2022 and win10ltsc. Might try profiling it.

azum4roll:
It's always stuck here.

I've tried leaving it on for a night.

JohnsterID:
My run completes same as the runner so I'm not sure. The other builds work for you?
Try turning on more verbose build output info to see if there is any hint, [MSBuild project build output verbosity](https://learn.microsoft.com/en-us/visualstudio/ide/reference/options-dialog-box-projects-and-solutions-build-and-run?view=vs-2022).

Issue #11331, opened 2024-10-19
CTD


RemitoL:
_1. Mod Version  
4.16.2 with EUI

_2. DirectX Version_
DX 9

_3. List of Other Mods_
none

_4. Describe the Issue_
CTD around turn 205

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[Enrico Dandolo_0206 AD-1460.zip](https://github.com/user-attachments/files/17445257/Enrico.Dandolo_0206.AD-1460.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

[logs.zip](https://github.com/user-attachments/files/17445259/logs.zip)

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

[CvMiniDump.zip](https://github.com/user-attachments/files/17445265/CvMiniDump.zip)

PR #11370, opened 2024-11-17
Fix CTD (#11331)

axatin:


azum4roll:
Image

I think it's better to add a NO_UNIT check here, and an assert to the start of `CvPlayer::IsUnitValidForVassalLevy()`.

axatin:
Does CvAssert work by now?

axatin:
Added the NO_UNIT check

JohnsterID:
> Does CvAssert work by now?

We can just use ASSERT, CvAssert still needs to re-implemented.

azum4roll:
ASSERT for making sure no logical errors are made, null checks for when null values are actually expected.

Preferably no check against modder stupidity (let it crash!)

axatin:
I agree. Changed it

azum4roll:
Oh I was expecting an ASSERT on eUnit not being NO_UNIT.

axatin:
Oh I see, makes sense. Done

PR #11384, opened 2024-12-03
Many assert fixes

axatin:


CvGameCoreDLL_Expansion2/CvCity.cpp
```diff
@@ -9759,26 +9759,24 @@ bool CvCity::IsBuildingLocalResourceValid(BuildingTypes eBuilding, bool bTestVis
 	bool bHasAnyOr = false;
 	bool bNeedAnyOr = false;
 	CvString tempTooltip;
-	for (int iResourceLoop = 0; iResourceLoop < GC.getNumResourceInfos(); iResourceLoop++)
+	for (int iPrereq = 0; iPrereq < GD_INT_GET(NUM_BUILDING_RESOURCE_PREREQS); iPrereq++)
```
azum4roll, line 9762: Stone Works has more than 5 prereqs so it'll break.

azum4roll: We should turn Building_LocalResourceAnds and Building_LocalResourceOrs into sets instead of arrays.

CvGameCoreDLL_Expansion2/CvBuilderTaskingAI.cpp
```diff
@@ -1835,7 +1842,7 @@ static vector<OptionWithScore<BuilderDirective>> FilterNonOptimalNoTwoAdjacentIm
 				continue;
 
 			BuildTypes eOtherBuild = it2->option.m_eBuild;
-			CvBuildInfo* pkOtherBuild = GC.getBuildInfo(eOtherBuild);
+			CvBuildInfo* pkOtherBuild = eOtherBuild != NO_BUILD ? GC.getBuildInfo(eOtherBuild) : NULL;
```
azum4roll, line 1845: Just continue instead of assigning null here if NO_BUILD imo

(1) Community Patch/Database Changes/Defines/CoreDefineChanges.sql
azum4roll, line 1: What's the point of these? Aren't they already initialized in the DLL?

azum4roll: I think that should be an expected case. Maybe print to Database.log as a warning instead of using assert?

azum4roll: Something like "WARNING: Integer Define Value not found for <name>, using default value <value> instead"

CvGameCoreDLL_Expansion2/CvCity.cpp
axatin: fixed it

Issue #8030, opened 2021-06-11
CvAssert macros are never hit


IanE9:
https://github.com/LoneGazebo/Community-Patch-DLL/blob/b34963d1356573117ea1c22c281f963d426de8db/CvGameCoreDLLUtil/include/CvAssert.h#L36-L44

``CvAssertDlg()`` is an imported function from a static library, but it compiles to just two instructions that unconditionally result in a false return:
```
35582510  xor         al,al  
35582512  ret
```
The function is probably ifdef'd to do this in the original source code at release and we have only a release build of the library.
Regardless we need an assert macro that actually works.

JohnsterID:
Potential code (with errors due to parameters changing and the need to include Windows.h): https://github.com/JohnsterID/Community-Patch-DLL/blob/CvAssertDlg/CvGameCoreDLLUtil/include/CvAssert.h

JohnsterID:
Wanted to get thoughts on the below.
https://github.com/JohnsterID/Community-Patch-DLL/commit/d1db0aae38d203dfd66b906dd6989d52fb6b563c
![CvAssertDlg-1a](https://github.com/user-attachments/assets/9aba3e37-e47c-40c6-985a-a65351d36c84)
https://github.com/JohnsterID/Community-Patch-DLL/commit/d480cf8890737702650705b912e6cd61a4ac1a61
![CvAssertDlg-2a](https://github.com/user-attachments/assets/c2337ad2-3234-4f29-b9df-dda514969633)

JohnsterID:
I've moved it to CvGameCoreUtils.cpp, to keep it within our solution and it seems the closest related cpp file. Best to not mix it with headers.

PR #11386, opened 2024-12-06
CvAssertDlg implementation

JohnsterID:
https://github.com/LoneGazebo/Community-Patch-DLL/issues/8030

I'm sure we can play with it more but most important note is that we need to maintain the exact same function signature to stay compatible with the rest of the code. Also, do we want a way to turn off CvAsserts completely whilst still using debug build?

image


azum4roll:
> Also, do we want a way to turn off CvAsserts completely whilst still using debug build?

Would be nice to have at the current state where assertion failures are so common, until we fix them all.

JohnsterID:
Added DISABLE_CVASSERT, just need to remove it.

PR #11444, opened 2024-12-17
CvAssert and PRECONDITION

axatin:
My suggestion of how to deal with Assertions:

- CvAssert and CvAssertMsg show a message dialog in Debug config (unchanged)
- CvAssert and CvAssertMsg show a message dialog also in Release config (can be turned off by undefining VPRELEASE_ERRORMSG in the project settings). The dialog informs the player that there is a bug in the code and asks them to report it on github. The player is allowed to continue the game, in which case no further message for this assert will be shown in the current session.

- Added macros PRECONDITION and PRECONDITION_MSG, which show message dialogs similar to the ones in CvAssert (depending on the project settings), but always cause the game to crash if the given expression is false. To be used for example for array bounds checks and other situations where we don't want the game to continue if the assertion fails.
- UNREACHABLE(); is now the same as PRECONDITION_MSG(false, "UNREACHABLE"), so it also produces an error message
- CvAssertFmt merged into CvAssertMsg.

todo: replace all occurrences of ASSERT() and FAssert() with CvAssert() or PRECONDITION


Cyclopia31:
> CvAssert hit in Release config Screenshot (956)
> 
> PRECONDITION hit in Release config, leading to a crash Screenshot (958) Screenshot (957)

Hi @axatin , do you know how it will behave with multiplayer games/modpacks ?

axatin:
> Hi @axatin , do you know how it will behave with multiplayer games/modpacks ?

Good point. I'm not sure, maybe the popup dialogs can cause problems in network games. I changed it so that the less severe warnings are now shown in the chat in MP games (can't test it but it should work fine). The severe errors will cause a crash anyway.

azum4roll:
> CvAssert hit in Release config Screenshot (956)
> 

Where does the message in CvAssertMsg show up?

PR #11448, opened 2024-12-19
TechEnhancedYield

axatin:


axatin:
todo: Building_YieldFromGPBirthScaledWithPerTurnYield should work on total empire yield, not city yield

PR #11457, opened 2024-12-22
Fix asserts

axatin:
- Fix asserts
- Fix Chartered Company giving extra resources only within working range

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvCity.cpp
+++ b/CvGameCoreDLL_Expansion2/CvCity.cpp
@@ -14643,11 +14643,11 @@ void CvCity::processBuilding(BuildingTypes eBuilding, int iChange, bool bFirst, 
 
 			// Add extra luxury counts
 
-			for (int iJ = 0; iJ < GetNumWorkablePlots(); iJ++)
+			for (int iPlotLoop = 0; iPlotLoop < GC.getMap().numPlots(); iPlotLoop++)
 			{
-				pLoopPlot = iterateRingPlots(getX(), getY(), iJ);
+				pLoopPlot = GC.getMap().plotByIndexUnchecked(iPlotLoop);
 
-				if (pLoopPlot != NULL && pLoopPlot->getOwner() == getOwner())
+				if (pLoopPlot != NULL && pLoopPlot->getOwner() == getOwner() && pLoopPlot->getOwningCityID() == GetID())
 				{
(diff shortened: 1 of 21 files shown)


CvGameCoreDLL_Expansion2/CvCity.cpp
azum4roll, line 1: Isn't there a faster way to loop all owned tiles of a city?

azum4roll: If there isn't, I think we should cache a list of them seeing how often we do that.

azum4roll: Or change this loop to a resource loop and use GetNumResourceLocal() instead.

azum4roll: Ok I've read the current code.

Currently it has to loop all plots and call addResourcesOnPlotToTotal(), removeResourcesOnPlotFromTotal(), addResourcesOnPlotToUnimproved(), or removeResourcesOnPlotFromUnimproved().

The first two is used to update total amount of resources connected by the player, as well as the amount of resources connected/worked by the city. The worked function can be removed as GetNumResourceWorked() is never called (also saving a bit of memory).

Instead of using the same m_paiNumResourcesLocal array to store the duplicated resources, we can use the original array to store the base amount and then a new one to store the modified amount, as that's what the building really does: modifying (duplicating) the number of (luxury) resources connected by the city. All the work on individual plots seems to be excessive.

Similarly, the latter two function calls are unnecessary. It'll affect the top panel UI on the unimproved resource part, but the amount really isn't duplicated before the resource is connected, so it technically isn't lying.

axatin: removing GetNumResourceWorked is fine for me and I agree that a cached list of city plots would be useful, but what would the change to m_paiNumResourcesLocal you suggest improve? I don't want to risk introducing new bugs by changing code that works fine

azum4roll: > but what would the change to m_paiNumResourcesLocal you suggest improve?

So buildings that modify the number of local resources can act on the city instead of updating every tile owned by the city.

axatin: let's consider this after 4uc integration, right now I don't think it has priority

PR #11462, opened 2024-12-23
Fix crash

KungCheops:
Fix crash when building an improvement, that spawns an adjacent resource, on an unowned tile.


KungCheops:
This currently only happens in my personal version of 4UC where the Zulu Unique GPTI spawns an adjacent cattle.

PR #11490, opened 2025-01-07
4.18.1 fixes

azum4roll:


azum4roll:
image
image

These methods are bad, trying to do two things at the same time (update + get) and violating const. The above fix split each of them into Get* (already existing) and Update* (still returning the count) methods.

The Lua API for CountCityFeatures and CountNumBuildings was adjusted accordingly. Not that any modders know about the second parameter.

m_piCityFeatures is currently unused in the DLL (and is inaccurate) but it's being updated. Not sure if it's ever useful. Lua API exists though so we can't really remove it.

Issue #11506, opened 2025-01-09
Delay increase in City Defense strength due to Defense process by 1 turn to prevent an exploit?


bsesar:
_1. Mod Version (X.Y.Z). Current Version: 4.18.1_

_4. Describe the Issue_

Currently, when a player uses a Defense process in a city, the city's defense strength immediately increases. This immediate increase allows the player to use the benefit during the turn of an attacking AI, and once the AI's attack is finished, but *before* the beginning of the player's turn, the player can turn off Defense process, switch to a building in the production queue, and obtain the benefit of production at the beginning of its (player's) turn.

An example of this exploit can be seen in Milae's video below

https://youtu.be/Biv6SU39FYE?t=488

The player uses Defense process in the city of Adwa and the city's defense strength increases to 70. After the AI (Germany) does its attack (turn), the player turns off the Defense process to switch to building Arsenal, gains the production, and the number of turns to produce the Arsenal decreases from 11 to 10. As a result, the player gains the increase in city's defense strength *for free*.

To prevent this obvious exploit (which the AI to my knowledge does not use), it seems reasonable to apply the increase in city's defense strength at the *beginning* of the turn that comes *after* Defense process is activated (i.e., delay the effect for one turn). This should prevent the above exploit and level the playing field between the player and AI.

This issue may be related to @azum4roll's comment in #10826.

azum4roll:
You aren't supposed to be able to change your production queue during someone else's turn. That should be guarded by the UI.

bsesar:
@azum4roll, would you recommend I change the title and description of the issue? Are there devs that can fix this UI issue or is this out of VP scope?

axatin:
> You aren't supposed to be able to change your production queue during someone else's turn. That should be guarded by the UI.

exactly. I'll fix it

Cyclopia31:
Does the defense bonus start at the end of the turn? Or when you select it during the turn? I can see problems with this in multiplayer simultaneous war between humans if the "when you select it during the turn" is the case.

bsesar:
Judging by the video, the defense bonus is immediate (in single player mode).

Issue #11505, opened 2025-01-09
4.18.1 Three errors during a single AI turn [Line: 19465; 19483; 19491]


brsoft1981:
_1. Mod Version (X.Y.Z). Current Version: 4.18.1 VP with EUI

_3. List of Other Mods: None

_4. Describe the Issue_
During an AI turn, I've got 3 errors (I think that Egypt is faulty)

![error19491](https://github.com/user-attachments/assets/1b1509a1-0015-45b1-9176-6efa3e0d4a73)
![error19483](https://github.com/user-attachments/assets/3faee0da-f332-49b5-a407-0ec56b3152c2)
![error19465](https://github.com/user-attachments/assets/5e2def73-3f47-4a85-9f98-dedcce1137f3)

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_
[3Errors_Savegame.zip](https://github.com/user-attachments/files/18361437/3Errors_Savegame.zip)

axatin:
Fixed. Instant production from events was implemented incorrectly as a negative overflow yield, as opposed to reducing the stored amount of production.

Issue #11503, opened 2025-01-09
4.18.1 Another Error [Line: 25220] (CS related)


brsoft1981:
_1. Mod Version (X.Y.Z). Current Version: 4.18.1 VP with EUI

_3. List of Other Mods: None

_4. Describe the Issue_
Error during an AI turn - when barbarians swarming a CS (Panama in this case), i've got that error:

![error 25220](https://github.com/user-attachments/assets/4ba74d7a-6823-4635-a247-2de3190f153a)

![Panama](https://github.com/user-attachments/assets/093d0383-c983-4453-af09-5dee7a6ba939)

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[ErrorSave.zip](https://github.com/user-attachments/files/18360313/ErrorSave.zip)

brsoft1981:
I've got exactly the same error later in the game, when I liberated Panama CS. There's a savegame and a screenshot
![Uploading 25220.jpg…]()


[Pre_liberate_CS_savegame.zip](https://github.com/user-attachments/files/18361674/Pre_liberate_CS_savegame.zip)

axatin:
That's something that needs to be fixed. When a CS is conquered, the bonuses it gives to a player are removed twice, so in this case cities end up with negative food per turn from city-states

Issue #11493, opened 2025-01-07
4.18.1 Error message on turn overs


skodkim:
_1. Mod Version (X.Y.Z). Current Version: 4.18.1_

4.18.1

_2. DirectX Version_

_3. List of Other Mods_

Mods included in zip-file: https://www.dropbox.com/scl/fi/kqhhn2p53zupfk1vkc8p4/Sid-Meier-s-Civilization-5.zip?rlkey=y5bzrhrf57mlt588dlw4sychz&st=xgmwyabl&dl=0
(also uploaded on https://github.com/LoneGazebo/Community-Patch-DLL/issues/11491)

_4. Describe the Issue_

CTD occus with error message when selecting End turn. The save game has been continued after the bug in https://github.com/LoneGazebo/Community-Patch-DLL/issues/11491 occured so may be linked.

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_

[QuickSave.zip](https://github.com/user-attachments/files/18327826/QuickSave.zip)

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_

[Logs.zip](https://github.com/user-attachments/files/18327820/Logs.zip)

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_

_8. Steps to reproduce the Issue (Optional)_

Happens if you select end turn after the quick save is loaded

_9. Screenshots of the Issue (Optional)_
![image](https://github.com/user-attachments/assets/580393b2-c059-4da9-a063-8ba55b2bb802)

skodkim:
And another one after turn 4:
[QuickSave.zip](https://github.com/user-attachments/files/18327882/QuickSave.zip)

![image](https://github.com/user-attachments/assets/9986b1f9-d994-47a2-875f-5267eafc7eae)

skodkim:
The latter bug also seems to be reported in https://github.com/LoneGazebo/Community-Patch-DLL/issues/11487

axatin:
The gold error is fixed

PR #11507, opened 2025-01-10
Fixes

axatin:


azum4roll:
Which code was changing gold to negative? If we just cap gold at 0 some penalty may be bypassed.

axatin:
An instant yield, but it's fine, if SetGold is called with a negative value the treasury is set to 0 anyway. The penalties are caused by having 0 gold and negative GPT

axatin:
Fix #11516 
Fix #11514 
Fix #11513 
Fix #11512 
Fix #11510 
Fix #11506 
Fix #11505 
Fix #11503 
Fix #11493 
Fix #11489 
Fix #11484

azum4roll (commented):
So this is different between traits (players) and buildings (cities)? How do Building_YieldPerFriend and Building_YieldPerAlly work?

axatin:
Building_YieldPerFriend and Trait_YieldPerFriend work the same way, but they use a meaning of "friend" that's different than the one used in the dll (`CvMinorCivAI::IsFriends` and `DoSetBonus`) and in the defines (for example `FRIENDS_CAPITAL_FOOD_BONUS_AMOUNT_PRE_RENAISSANCE` and `ALLIES_CAPITAL_FOOD_BONUS_AMOUNT`).

On the other hand, `CvPlayer::GetNumCSFriends()` seems to have been added for the YieldPerFriend tables and uses the same definition of friend as the tables. Very inconsistent.

PR #11508, opened 2025-01-10
Fix all policies being available from ancient era on in CP

axatin:
In 4.18 CP all policy branches are available already in ancient era because they have `GetNumPolicyRequirement = 0`. We can fix this either by disabling early policy branch unlocking in CP altogether, or by setting the default value of the column to a high value. The latter is how it was done in pre-4.17:
image

Any preferences?

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPolicyClasses.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPolicyClasses.cpp
@@ -5420,10 +5420,13 @@ bool CvPlayerPolicies::CanUnlockPolicyBranch(PolicyBranchTypes eBranchType)
 				}
 			}
 			//Using a system of numbers instead? Okay.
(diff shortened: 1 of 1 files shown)


axatin:
Third option: set the default value for NumPolicyRequirement to -1, and change the code to
```
if ( pkBranchEntry->GetNumPolicyRequirement() >= 0 && iNumPolicies >= pkBranchEntry->GetNumPolicyRequirement())
{
				bCanUnlockEarly = true;
}
```

azum4roll:
> Third option: set the default value for NumPolicyRequirement to -1, and change the code to
> 
> ```
> if ( pkBranchEntry->GetNumPolicyRequirement() >= 0 && iNumPolicies >= pkBranchEntry->GetNumPolicyRequirement())
> {
> 				bCanUnlockEarly = true;
> }
> ```

Do this.

PR #11536, opened 2025-01-15
Fix crash

axatin:
Error message CvTradeClasses, line 1916, "iIndex out of bounds"


azum4roll:
Oh cool an actual bug!

axatin:
Yes. I think this is the first time we actually found a case of writing beyond array bounds

azum4roll:
So this fixes a wrong trade route being ended when a canal or water polder is removed (or undefined behaviour). Kinda hard to notice.

axatin:
Yes, it would have taken us a long time to find this without the preconditions. So implementing them was worth it already

PR #11551, opened 2025-01-17
Compressed atlases, Hacienda&Ordo DX9 fixes

nobubblegums:
Compressed new atlases, making the project ~25MB smaller, hopefully this may also help with late game memory issues.
Fixed long standing red box texture glitch with Hacienda and Ordo for DX9. These were caused by broken references in the gr2 file, I gave Ordo the base texture of Tipi, and added generic height maps to Hacienda.


azum4roll:
Will need work to merge this into 4UC.

PR #11586, opened 2025-01-26
Exclude non-CP tables from remapping

axatin:


axatin:
savegame compatible for base VP games, but not for games with modmods

PR #11588, opened 2025-01-26
Synchronize

RecursiveVision:


azum4roll:
The old CvAssertMsg and VALIDATE_OBJECT are still in here.

Issue #11594, opened 2025-01-26
4.18.4 error LINE4912


cubapl:
Mod VP 4.18.4 and no additional mods
Happened before round 183 during move of AI
Cape Town was Holland spere of influence, my friend and Venezia took it over
Occured together with error LINE4912; LINE4926 ; LINE25094

![Image](https://github.com/user-attachments/assets/c7eb57c1-89fc-4dec-a026-b5e9accd9c51)

logs and save: 
https://limewire.com/d/650ad2f7-d65b-4e75-b146-25040d6566d0#K4x0zTW39or1Zi4zYBUmQWMn-YmSQcQO_fJA6bsubIs

azum4roll:
Inconsistent effects when a minor civ dies. Yields seem to be removed twice from the ally?

axatin:
Something weird is going on in `ChangeFriendshipWithMajor`. I've seen bugs with great diplomats lately, those seems to have the same cause.

PR #11601, opened 2025-01-28
Fix bug in AI project build evaluation

axatin:
Fixes an assert message "eBuildingNum > -1". Much more importantly, it fixes `GetBestProductionCity(BuildingTypes eBuilding, ProjectTypes eProject)` always returning NULL when called for a Project. This might have caused the AI to select a suboptimal city for Apollo Program or Citizen Earth Protocol, or even to not build them at all.


axatin:
Fix two bugs with resources given by city-states:
(1) When a city-state researches a technology, newly revealed resources are not given to its ally
(2) When a city-state's ally changes and the CS owns a resource that the old ally can see but the new one can not, the CS does not stop exporting the resource to the old ally.

axatin:
Fix #11594 .
Fix bug in ally status change when using a Great Diplomat.

Issue #11628, opened 2025-02-03
Version 4.18.4 error: Strange Error: Assertion Failed @ Line 49 of CvPlayer.AI.cpp - non-CTD


Giterrobo:
This error happened while using the full installation of the VP pack (CP patch + VP + UI), no other mods or alterations were used. No CTD occurred.

I'm not really sure how to describe this weird error, it just happened while moving my units around during the turn of the attached autosave, no crash occurred thankfully.

![Image](https://github.com/user-attachments/assets/5f55100d-0b57-4448-8c71-f16150b98e33)

[AutoSave_0210 AD-1500.zip](https://github.com/user-attachments/files/18647333/AutoSave_0210.AD-1500.zip)

azum4roll:
I can't load the save. Are you sure it's the correct one? Try saving manually.

cubapl:
@azum4roll  I just had this error on turn 220 after conquering Siamse city Muang Saluang. I have autosave from 218 and quicksave taken just moment after ctd. please see logs and save. i dont have screenshot but all the informations and error line number are the same. i play 4.18.4

https://limewire.com/d/1696a6c7-1ddb-4423-9721-6712a8109a2e#jxDfd_VwJbA1dDyIWxvC-MZ0upN4yB2m0euOHy1TY8g

axatin:
> I can't load the save. Are you sure it's the correct one? Try saving manually.

it's 43 civ

azum4roll:
> @azum4roll I just had this error on turn 220 after conquering Siamse city Muang Saluang. I have autosave from 218 and quicksave taken just moment after ctd. please see logs and save. i dont have screenshot but all the informations and error line number are the same. i play 4.18.4
> 
> https://limewire.com/d/1696a6c7-1ddb-4423-9721-6712a8109a2e#jxDfd_VwJbA1dDyIWxvC-MZ0upN4yB2m0euOHy1TY8g

This one's been fixed.

Issue #11624, opened 2025-02-02
Version 4.18.4 error: Assertion Failed on Line 38252: non-CTD error


Giterrobo:
This error happened while using the full installation of the VP pack (CP patch + VP + UI), no other mods or alterations were used.

This error occurred when I was attempting to replace an improvement with another.
Specifically, I attempted to replace an Eiki (the Hunnic unique Improvement), that was sitting on a road with a city connection to my capital, with a Village. The tile that I had the Eiki on was a tile that was claimed by the Eiki's claiming effect, and was not purchased through border growth or gold.

![Image](https://github.com/user-attachments/assets/5d532f66-7f34-4de6-8096-7bf9ee64e121)

Giterrobo:
The error happened again. Same error number, but this time it happened when my worker was completing an Eiki after it got "woken up" by a nearby enemy unit. The Eiki completed, but the error happened, no crash again.

![Image](https://github.com/user-attachments/assets/1b8a8c96-c346-42f9-b0df-5f5ea8176db6)

Issue #11604, opened 2025-01-29
Air Repair promotion does not function


elite-spud:
1. Mod Version (X.Y.Z).
4.18.4 with EUI

2. DirectX Version
DX 10 & 11

3. List of Other Mods
none

4. Describe the Issue
Fighter units in possession of the Air Repair promotion do not heal after performing Air Strike or Air Sweep actions.
Units do heal after Intercept actions, but that also occurs for units without Air Repair

5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[saves.zip](https://github.com/user-attachments/files/18580916/saves.zip)

6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

[Logs.zip](https://github.com/user-attachments/files/18580918/Logs.zip)

8. Steps to reproduce the Issue (Optional)
- Load the Autosave, and issue orders for triplanes in Gao if loading the `Pre_Action` save
- In the `Post_Action` save (pictured in the screenshots), one of the middle 2 triplanes is given an Air Strike order, while the other is given an Air Sweep order
- End the turn.
- Observe that only the triplanes which take no action or attempt interception will heal.
- The same behavior can be observed on the Jet Fighters in Rotterdam.

9. Screenshots of the Issue (Optional)

Pre-Action Screenshots:
<details>

![Image](https://github.com/user-attachments/assets/016a79d1-0851-4409-bbea-e869b8236721)
![Image](https://github.com/user-attachments/assets/b48df2d0-fc65-4d8c-8bdc-633183e7d90b)
![Image](https://github.com/user-attachments/assets/26897d8f-51d7-4998-96ee-b897075eae7e)
![Image](https://github.com/user-attachments/assets/7a60aede-ed9a-4776-99e8-9f6b3d2d0a24)

</details>

Post-Action Screenshots:
<details>

![Image](https://github.com/user-attachments/assets/d97830b0-e6fb-49f3-99d8-556fb3580054)
![Image](https://github.com/user-attachments/assets/80d46709-495d-4db8-9693-bdbb4d890a13)
![Image](https://github.com/user-attachments/assets/ca892754-961b-4ff1-803f-3f36b19b41a4)
![Image](https://github.com/user-attachments/assets/e87ddd92-ee71-4ae9-98f9-7faba7cb3d15)

</details>

Issue #11599, opened 2025-01-27
4.17.8 Expanding on issue #11587, Swedisch Great Generals give XP only to land units


Findorfin:
Expanding on issue #11587 (regarding the XP gain from Self Determination policy), the XP gain from the birth of Swedish Great Generals (granted by Swedens unique ability "Lion of the North") applies only to land units, not towards sea or air units. 

Patch 4.17.8, various game conditions and settings;

Thanks!

axatin:
in the code the healing is explicitly restricted to land units, so this behavior is intended and the text is wrong

![Image](https://github.com/user-attachments/assets/2b9afd41-2137-425b-b569-11681e1814af)

PR #11642, opened 2025-02-06
Misc fixes

axatin:
- Fix incorrect description of Sweden's UA (#11599)
- Fix air repair promotion not working (#11604)
- Fix Ekis built in neutral land not triggering CvPlayer::changeImprovementCount (#11624)
- Fix assert hit when using Squads (#11567) (I don't know if this makes squads fully multiplayer-compatible of if there are more issues)
- Fix a harmless assert hit caused by EUI (#11628)
- The city governor now updates tile assignments in Venetian puppets as frequently as in non-puppet cities (#11563)


axatin:
I don't have any further changes for 4.19 .

PR #11654, opened 2025-02-09
Add LuaCATS documentation for UI

vaderkos:
Added LuaCATS documentation for UI and other parts of the code.

I've tried to add documentation and types for each function where possible. Each time I didn't know what code does I left `TODO docs` for such function. 

This does not contain all the things from `UI` table but everything that I've found using `UI\.\w+\(` regex in Civ5 `Assets` directory and in `Community-Patch-DLL` repository. All types were taken either from usages that I've found or from http://modiki.civfanatics.com/index.php/Lua_and_UI_Reference_(Civ5)

P.S. Github may collapse `UI.d.lua`(It's big) where most documentation is.

UI_Docs_Example


LuaCATS/Pseudo.d.lua
azum4roll, line 1: What's this file for?

vaderkos: The idea is instead of using integers use something like branded types for the ids, to eliminate situation where, for example, TeamId is used where PlayerId expected. 
I've named this file `Pseudo.d.lua` because such identifiers are documented in [modiki](http://modiki.civfanatics.com/index.php?title=AreaID_(Civ5_Type)) as pseudo-types.
Let me know you have any better idea 😄 .
image

azum4roll: But how would LuaCATS enforce that? Like when we call city:GetTourismMultiplier(1), it's still an integer being passed in when the documentation expects a "PlayerID".

vaderkos: @azum4roll `PlayerId` type is de-facto integer, we cannot enforce it, unless type it as something else or, for example, intersection like `integer & { __brand: 'PlayerId' }`  (but as far as I know this is not possible yet in luals).

What this gives is type hints mention `PlayerId` instead of `integer` and it emits warning if something else comes in.
image

image

vaderkos: I've left it as just aliases to integers, I haven't tried yet to make it enforcing using LuaCATS type system.

vaderkos: IMHO little type hint is better than plain integer

azum4roll: If it's shown somewhere in the tooltip that PlayerId is integer.

vaderkos: @azum4roll I've changed definition to aliases instead of classes. I could also add it as union `-1 | integer` to signify `NO_PLAYER`, but integer already includes -1 so no need I think.
image
image

PR #11661, opened 2025-02-14
4.19.1 fixes

azum4roll:


azum4roll:
Note that the tooltip for concert tour mission is still having a dynamic number instead of the number set on the unit's birth. That will be fixed in 5.0.

PR #11710, opened 2025-03-11
Simplify some worker homeland logic

KungCheops:
Now that Ilteroi fixed the inefficient pathfinding for workers in for the homeland AI, we can remove some logic meant to avoid pathfinding.

Also remove double penalty for workers currently building an improvement.


KungCheops:
Why was this one closed?

PR #11711, opened 2025-03-11
Fix nasty typo

KungCheops:


azum4roll:
Why I prefer using getters. It's a struct though so can't be helped.

PR #11717, opened 2025-03-14
Traits:ArtistGoldenAgeTechBoost

THELANDOFLOVE:
Great artists start the Golden Age with research points equal to half the median of all currently researchable technologies.


azum4roll:
Some feedback:

- I know this is intended to give an amount of science equivalent to a research agreement, but the median thing is already unused since G&K. It's awkward to describe that when nothing else works like that anymore.
- A check for artist is unnecessary and should be omitted, to make the trait ability more flexible (in case some other modded civ wants to have both this and a unique writer that can start golden ages).

THELANDOFLOVE:
Thank you for your reply. However, the reason why I designed this is because I wrote a CIVILIZAION-Ming-Dynamic, which I need to use and think may make the gameplay more interesting. If you think it's not suitable, do you have any other suggestions for me? I would be very grateful.

PR #11727, opened 2025-03-20
fix: empty str in CvLuaPlayer::lGetOpinionTable

qingyin3545:
561840CD76CCDEFE0F7115383E648BF0
61E48311A7880D6B8FC05B384F36293A

### DIFF
--- a/(1) Community Patch/Database Changes/Text/en_US/UI/CoreNewUIText.xml
+++ b/(1) Community Patch/Database Changes/Text/en_US/UI/CoreNewUIText.xml
@@ -933,6 +933,9 @@
 			<Text>You are not competing for the favor of the same City-States.</Text>
 		</Row>
 		<!-- Victory dispute -->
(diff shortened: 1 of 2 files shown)


azum4roll:
I'll leave this to @RecursiveVision, but I think it should just not have a line instead.

azum4roll:
Actually, that should be unreachable! How is iValue not 0 when victory dispute level is none?

qingyin3545:
> Actually, that should be unreachable! How is iValue not 0 when victory dispute level is none? Unless you modded OPINION_WEIGHT_VICTORY_NONE?

We modified OPINION_WEIGHT_VICTORY_NONE in our mod to allow it to be a negative number. Perhaps we can also hide it in this case.

azum4roll:
Ah, so the AI is supposed to be friendlier if there's no victory dispute.

qingyin3545:
> [8e4a23a](https://github.com/LoneGazebo/Community-Patch-DLL/commit/8e4a23a4af2b00bf4ff842852df5af193061185e)

This is much more robust than what I wrote. Thank you very much!

PR #11729, opened 2025-03-22
Add RequiresResource improvement column

KungCheops:


KungCheops:
This is support for the Marajoara modmod

PR #11757, opened 2025-04-24
Fix crash

axatin:


azum4roll:
I'd advise against modifying a list while it's being iterated. Any way to code it better?

Issue #11731, opened 2025-03-23
(4.20.1) Crash during AI turn


GeneralAmadeus:
Mod Version: 4.20.1

DirectX Version: DX 10&11

List of Other Mods: Some maps and a key rebind mod. These can be removed mid game and does not change the outcome.

Describe the Issue: Game crashes during AI turn. There are actually 2 crashes in the same game at different points of time but i think the same underlying problem. First crash is end of turn 195. I tried going back to turn 191 and hooped the crash would not happen but it always did UNTIL i rejected a friendship offer from Denmark. That allowed me to continue the game until turn 212 where again i crash some time after ending the turn. 
I have provided the saves for turn 191, before the friendship offer (non sure what turn exactly that happens). 
Turn 195, having accepted the offer and crashing after end turn. 
Turn 212, having rejected the friendship offer ~20 turns ago but now crashing again after ending the turn.
The minidump is from turn 212.

[Boudicca_0191 AD-0265.zip](https://github.com/user-attachments/files/19409574/Boudicca_0191.AD-0265.zip)
[Boudicca_0195 AD-0325  dead.zip](https://github.com/user-attachments/files/19409573/Boudicca_0195.AD-0325.dead.zip)
[Boudicca_0212 AD-0580  dead2.zip](https://github.com/user-attachments/files/19409575/Boudicca_0212.AD-0580.dead2.zip)
[CvMiniDump_20250323_161630_4.20.1_Release.dmp.zip](https://github.com/user-attachments/files/19409577/CvMiniDump_20250323_161630_4.20.1_Release.dmp.zip)
[Logs.zip](https://github.com/user-attachments/files/19409576/Logs.zip)

axatin:
I can't reproduce either crash. Are you sure the save files are from the correct turns?

GeneralAmadeus:
Yes I am very sure.

GeneralAmadeus:
I just nuked civ5 and did a clean install for both the game and VP. Still crashes on both turns. As far as I can tell the latter save crashes during or after the Incas turn. I added the logs and minidump from that crash but have no idea if more of them are of any use.

[CvMiniDump_20250325_221044_4.20.1_Release.dmp.zip](https://github.com/user-attachments/files/19455994/CvMiniDump_20250325_221044_4.20.1_Release.dmp.zip)
[Logs.zip](https://github.com/user-attachments/files/19455993/Logs.zip)

bsesar:
[...]

PR #11759, opened 2025-04-25
Better implementation of fix for CTD #11731

axatin:


azum4roll:
I'd write the erase loop like this:

```cpp
	list<int>::iterator it = m_CurrentTurnUnits.begin();
	while (it != m_CurrentTurnUnits.end())
	{
		CvUnit* pUnit = m_pPlayer->getUnit(*it);
		if (pUnit->TurnProcessed()) // I don't think pUnit can be NULL here? Maybe add an assert
		{
			it = m_CurrentTurnUnits.erase(it);
			continue;
		}
		++it;
	}
```

axatin:
okay changed it. But I kept the null check, just in case. It's also there in the previous loop.

PR #11766, opened 2025-05-03
4uc Promotions

axatin:
4uc:
- All UnitPromotion tables (1 through 6)
- UnitPromotions_UnitCombatMods.Attack, UnitPromotions_UnitCombatMods.Defense

bugfixes:
- Fix AI not using damaged units, scouts, caravels or carriers to pillage trade units on the same tile as the unit
- Fix AI trying to pillage trade routes with units that have no movement points left when reaching the trade unit
- Fix unit sorting in FindUnitsForHarassing not working


axatin:
- Train explorers to plunder trade routes
- Allow explorers to enter enemy territory, taking into account danger level and unit HP
- Explorers take into account instant healing when considering whether it's safe to pillage a tile
- Fix explorers not making any further moves if they pillage a tile and still have MP left

PR #11807, opened 2025-06-09
Replace Improvement confirmation

axatin:
Don't show the confirmation popup for replacing an improvement if the old improvement didn't connect the resource on the plot but the new improvement does

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvGame.cpp
+++ b/CvGameCoreDLL_Expansion2/CvGame.cpp
@@ -3195,7 +3195,9 @@ void CvGame::handleAction(int iAction)
 								CvPlot* pPlot = pkHeadSelectedUnit->plot();
 								if(pPlot != NULL)
 								{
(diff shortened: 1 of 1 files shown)


azum4roll:
I can see a need for an exception for Moai.

RecursiveVision:
@azum4roll When replacing a Moai, or when creating a Moai?

azum4roll:
When replacing a good improvement that doesn't connect the resource (idk how to define it).

axatin:
I don‘t want a hard-coded check on IMPROVEMENT_MOAI there, and as you said it’s hard to determine when an improvement is "too good".

The confirmation popup is only there to prevent  accidentally building an improvement where one already exists, giving the order to build an improvement that connects a resource rarely happens accidentally.

RecursiveVision:
I'm gonna tweak this just a tiny bit because I decoupled ConnectsAllResources and CreatedByGreatPerson, but otherwise I think we're okay.

PR #11852, opened 2025-07-29
Minor improvements to beliefs text

ron-murhammer:
Some minor improvements to beliefs text for consistency.

### DIFF
--- a/(2) Vox Populi/Database Changes/Text/en_US/Beliefs/BeliefTextChanges.sql
+++ b/(2) Vox Populi/Database Changes/Text/en_US/Beliefs/BeliefTextChanges.sql
@@ -12,7 +12,7 @@ SET Text = 'God of Fire'
 WHERE Tag = 'TXT_KEY_BELIEF_EARTH_MOTHER_SHORT';
 
 UPDATE Language_en_US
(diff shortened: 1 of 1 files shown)


Hokath:
thanks for doing this.
the only question I have is about the Earth Mother one. If you have a Mine on a Resource it doesn't improve (because the resource was revealed after the Mine was built, e.g. a Sheep), then do you get the bonus from the Pantheon? If you don't then we need to keep the word "Improved"

ron-murhammer:
I'm actually not sure. I removed "improved" to make it consistent with God of the Sun "+1 [ICON_PEACE] Faith, [ICON_FOOD] Food, and [ICON_RESEARCH] Science from **Farms on Resources**."

So I guess "improved" should be added to both or neither of them as I assume they work the same way.

ron-murhammer:
Updated to add "improved" to Earth Mother and God of the Sun as they both require the resource to be improved to get the bonus yields.

azum4roll:
Can someone try *removing* the pantheon (via conversion) and check if it affects the yields of a mine on sheep? That could be bugged (again, on the lack of handling of revealed resource after researching a tech).

axatin:
> Can someone try _removing_ the pantheon (via conversion) and check if it affects the yields of a mine on sheep? That could be bugged (again, on the lack of handling of revealed resource after researching a tech).

I tested it, it works fine

PR #11855, opened 2025-07-31
Add database docs

vaderkos:


azum4roll:
So `DB.Query` doesn't actually query the DB? I've been avoiding using it due to that.

The comparison is unfair: it's only the `GameInfo.Units()` iterator that's slow, as it fetches the entire row (155 columns) and creates a table for each iteration. `GameInfo.Units["UNIT_WARRIOR"]` is much faster in comparison, but of course should still be avoided in nested loops.

`UsingGameInfo` can be roughly optimized as below:
```lua
function UsingGameInfo()
	for row in GameInfo.Civilization_UnitClassOverrides() do
		local c = GameInfo.Civilizations[row.CivilizationType]
		local u = GameInfo.Units[row.UnitType]
		u and consume(u.Description, c and c.Adjective or nil)
	end
end
```

I'm prefetching the commonly used columns (Type, ID, Description, etc.) of primary tables (Units, Civilizations, etc.) globally in my UI rewrite, so retrieval of each unit, civilization, etc. for simple use is much faster (~1s for 10000000 retrievals). EUI has been doing this but for whole query results, which results in faster performance but memory problems. My implementation should strike the middle ground.

vaderkos:
@azum4roll This information is related to selecting multiple data, I may adjust docs a bit..
I'm in the middle of optimizing of TechHelpInclude, and there is a lot of iterations over GameInfo objects which makes it slow. 
If you want a single entry of something by it's `Type` or `ID`, GameInfo.Units[?] is completely fine :)

P.S> Could you share more insights about UI rewrite, maybe you accept contributions?

vaderkos:
> So `DB.Query` doesn't actually query the DB? I've been avoiding using it due to that.

@azum4roll it queries DB. If you write proper sql that actualy fetches data that's needed and prepare it using `DB.CreateQuery` it should be faster.

I'll give u the example, consider writing these queries using `GameInfo`.

```sql
SELECT 
  CASE WHEN b.Time IS NULL THEN (
    SELECT 
      bf.Time 
    FROM 
      BuildFeatures bf 
    WHERE 
      bf.BuildType = c.BuildType
  ) ELSE b.Time END as Time, 
  c.TimeChange as Change, 
  b.Description as Description, 
  civ.ShortDescription as Civilization 
FROM 
  Build_TechTimeChanges c 
  INNER JOIN Builds b ON b.Type = c.BuildType 
  LEFT JOIN Improvements i ON i.Type = b.ImprovementType 
  LEFT JOIN Civilizations civ ON civ.Type = i.CivilizationType 
WHERE 
  c.TechType = ? 
ORDER BY 
  Civilization IS NOT NULL, 
  Description, 
  Civilization

```

or this

```sql
SELECT 
  b.Description, 
  c.Adjective 
FROM 
  Buildings b 
  INNER JOIN BuildingClasses bc ON bc.Type = b.BuildingClass 
  LEFT JOIN Civilization_BuildingClassOverrides o ON o.BuildingType = b.Type 
  LEFT JOIN Civilizations c ON c.Type = o.CivilizationType 
  OR c.Type = b.CivilizationRequired 
WHERE 
  b.ShowInPedia = 1 
  AND bc.MaxGlobalInstances < 0 
  AND b.PrereqTech = ? 
ORDER BY 
  c.Adjective IS NOT NULL, 
  b.Description, 
  c.Adjective

```

azum4roll:
Both of those feel like hell to write, either in Lua + GameInfo loops or in SQL. However, I think a GameInfo version (optimized) would be easier to read or customize than the SQL version.

I have done something similar to the second one in TechButtonInclude.lua. Instead of relying on the `Civilization_BuildingClassOverrides` table, I call `player:GetCivilizationBuilding()` instead and cache the result. Yours seems to be looking for ALL unique buildings instead?

Code snippet (determining which building buttons should show up on a given tech for the given player):
```lua
	local function CanPlayerEverBuild(eBuilding)
		if not pPlayer then
			return false;
		end

		local kBuildingInfo = GetInfoFromId("Buildings", eBuilding);
		if kBuildingInfo.CivilizationRequired and kBuildingInfo.CivilizationRequired ~= strCivType then
			return false;
		end

		local eCheckBuildingClass = GameInfoTypes[kBuildingInfo.BuildingClass];
		if pPlayer:GetCivilizationBuilding(eCheckBuildingClass) ~= kBuildingInfo.ID then
			return false;
		end

		return true;
	end

	local tCanEverBuildCache = {};
	local function CanPlayerEverBuildCached(eBuilding)
		if tCanEverBuildCache[eBuilding] == nil then
			tCanEverBuildCache[eBuilding] = CanPlayerEverBuild(eBuilding);
		end
		return tCanEverBuildCache[eBuilding];
	end

	for kBuildingInfo in GameInfo.Buildings{PreReqTech = kTechInfo.Type} do
		if kBuildingInfo.ShowInPedia and CanPlayerEverBuildCached(kBuildingInfo.ID) then
			GenerateNextButtonFromInfo(SetupBuildingButton, kBuildingInfo);
			if iButtonIndex > iButtonCount then return iButtonCount end
		end
	end
```

azum4roll:
> @azum4roll This information is related to selecting multiple data, I may adjust docs a bit.. I'm in the middle of optimizing of TechHelpInclude, and there is a lot of iterations over GameInfo objects which makes it slow. If you want a single entry of something by it's `Type` or 'ID`, GameInfo.Units[?] is completely fine :)
> 
> P.S> Could you share more insights about UI rewrite, maybe you accept contributions?

It's still WIP but you can check my 4uc-integration branch.

vaderkos:
> Yours seems to be looking for ALL unique buildings instead?

@azum4roll That is sql is responsible for getting tech info in `TechHelpInclude.lua` in this PR.
Another pros that you get from sql is sorting on the db level.
Also the problem with GameInfo object that it have usually much more information that you actually need for the task you doing. 
<details>
<summary>
Looks like this.
</summary>
image
</details>

azum4roll:
> > So `DB.Query` doesn't actually query the DB? I've been avoiding using it due to that.
> 
> @azum4roll it queries DB. 

If it really does then SQLite queries are actually very fast (0.7ms per query in your `db.md` example, including C++ overhead). I guess that's the advantage of being "lite".

PR #11856, opened 2025-07-31
Add more lua documentation

vaderkos:


azum4roll:
I'm working on some of these in the huge commits of UI rewrite, so please don't merge these yet.

azum4roll:
On the documentation, what do you think about the included functions like those in TechHelpInclude.lua? VSCode complains about undefined global unless the files are open.

vaderkos:
> On the documentation, what do you think about the included functions like those in TechHelpInclude.lua? VSCode complains about undefined global unless the files are open.

You could either add it to `workspace.library` in `.luarc.json` but it would make it super global, or you have to keep it open.
I think it'd work normally if civ5 lua used `require` but with `include` it doesn't know that file was imported.

vaderkos:
> On the documentation, what do you think about the included functions like those in TechHelpInclude.lua? VSCode complains about undefined global unless the files are open.

@azum4roll you could also try something like

```lua
if require then
  require('path to help info include')
end
```

It may trick lua into importing file into lsp, but wont do anything in civ5 as require is nil

azum4roll:
Hmm, I don't think we should change code just for the sake of suppressing warnings.

I'd rather have everything added to the library. Define the base Civ5 included functions in the LuaCATS folder, and I'll separate the UI and include files.

azum4roll:
I'm differentiating instance IDs and database row IDs (called `<X>Type` in my docs). Players only have `PlayerId` and techs only have `TechType`, but units have both `UnitId` and `UnitType` and they shouldn't be confused with each other.

vaderkos:
> I'm differentiating instance IDs and database row IDs (called `<X>Type` in my docs). Players only have `PlayerId` and techs only have `TechType`, but units have both `UnitId` and `UnitType` and they shouldn't be confused with each other.

But what about e.g. TechId which is in database, it's numeric.
So how do then they called?
Could you add some code snippet so I can take a look?

vaderkos:
> Hmm, I don't think we should change code just for the sake of suppressing warnings.
> 
> I'd rather have everything added to the library. Define the base Civ5 included functions in the LuaCATS folder, and I'll separate the UI and include files.

What about a PR with directory structure you want to introduce.
You can just add .gitkeep files with description of purpose of folders, and I could take a look into it.

vaderkos:
> Hmm, I don't think we should change code just for the sake of suppressing warnings.
> 
> I'd rather have everything added to the library. Define the base Civ5 included functions in the LuaCATS folder, and I'll separate the UI and include files.

@azum4roll I think I have found solution. We just have to develop little plugin that does `ResolveRequire` when `include` is called and add it to autoload in `.luarc.json`. 
And there is already a part of code that can be used to emulate VFS in plain lua lsp here https://github.com/vaderkos/Community-Patch-DLL/blob/efeb834e5e89db8ce991b5e8fe59545388e63efa/(1)%20Community%20Patch/Kit/CPK.lua#L29


Info here --> https://luals.github.io/wiki/plugins/#resolverequire

azum4roll:
> > Hmm, I don't think we should change code just for the sake of suppressing warnings.
> > I'd rather have everything added to the library. Define the base Civ5 included functions in the LuaCATS folder, and I'll separate the UI and include files.
> 
> @azum4roll I think I have found solution. We just have to develop little plugin that does `ResolveRequire` when `include` is called and add it to autoload in `.luarc.json`. And there is already a part of code that can be used to emulate VFS in plain lua lsp here https://github.com/vaderkos/Community-Patch-DLL/blob/efeb834e5e89db8ce991b5e8fe59545388e63efa/(1)%20Community%20Patch/Kit/CPK.lua#L29
> 
> Info here --> https://luals.github.io/wiki/plugins/#resolverequire

Sorry, I'm not good enough to understand that include function. Does it help with the warning problem?

azum4roll:
> But what about e.g. TechId which is in database, it's numeric. So how do then they called? Could you add some code snippet so I can take a look?

C++ enums have all of the database IDs as `<X>Types`. `TechTypes` is an enum that refers to the ID of the Technologies table. A Lua accessible enum example is `YieldTypes`.

Now C++ also has player IDs as `PlayerTypes`, but that really should've been `PlayerId`.

I'll standardize database IDs as `<X>Type` to not be confused with instance IDs (`<X>Id`) or enum tables (`<X>Types`).

Example:
```lua
--- @class YieldTypes
--- @field [string] YieldType
YieldTypes = {}

--- @alias YieldType integer
```

vaderkos:
Closing as @azum4roll added it here https://github.com/LoneGazebo/Community-Patch-DLL/pull/11864

PR #11857, opened 2025-07-31
Replace DisplayPercentage with common AsPercentage function

vaderkos:
* Added `CPK` aka Community Patch Kit.
* Replaced `DisplayPercentage` everywhere with common `AsPercentage` implementation from `CPK`
* Added module auto-loader for CPK.
* Added `StringBuilder` implementation for future optimizations (in following PR).

See [File](https://github.com/vaderkos/Community-Patch-DLL/blob/0f1d6ecf7dccdf1d819c3714ec266a6e76ae408c/(1)%20Community%20Patch/Kit/readme.md?plain=1#L1) for more info.

### DIFF
--- a/(1) Community Patch/Community Patch.civ5proj
+++ b/(1) Community Patch/Community Patch.civ5proj
@@ -1,6 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Deploy" ToolsVersion="4.0"
-  xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project DefaultTargets="Deploy" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Default</Configuration>
     <ProjectGuid>{0d66d522-b624-4bc5-acfe-15a0c5b729f4}</ProjectGuid>
@@ -972,8 +971,10 @@
     <Folder Include="Database Changes\WorldMap\Resources" />
     <Folder Include="Database Changes\WorldMap\Terrains" />
     <Folder Include="Database Changes\WorldMap\Yields" />
+    <Folder Include="Kit\Util\" />
     <Folder Include="LUA" />
     <Folder Include="Mapscripts" />
+    <Folder Include="Kit" />
   </ItemGroup>
   <ItemGroup>
     <Content Include="Assets\DoMImages\dom_catherine.dds">
@@ -3295,6 +3296,22 @@
     <Content Include="Database Changes\WorldMap\Yields\YieldTableChanges.sql">
       <SubType>Lua</SubType>
       <ImportIntoVFS>False</ImportIntoVFS>
+    </Content>
+    <Content Include="Kit\CPK.lua">
+      <SubType>Lua</SubType>
+      <ImportIntoVFS>True</ImportIntoVFS>
+    </Content>
+    <Content Include="Kit\Util\CPK.Util.AsPercentage.lua">
+      <SubType>Lua</SubType>
+      <ImportIntoVFS>True</ImportIntoVFS>
+    </Content>
+    <Content Include="Kit\Util\CPK.Util.lua">
+      <SubType>Lua</SubType>
+      <ImportIntoVFS>True</ImportIntoVFS>
+    </Content>
+    <Content Include="Kit\Util\CPK.Util.StringBuilder.lua">
+      <SubType>Lua</SubType>
+      <ImportIntoVFS>True</ImportIntoVFS>
     </Content>
     <Content Include="LUA\CityBannerManager.lua">
       <SubType>Lua</SubType>
--- a/(1) Community Patch/Kit/CPK.lua
+++ b/(1) Community Patch/Kit/CPK.lua
@@ -1,0 +1,357 @@
+do
+	if CPK ~= nil then
+		return
+	end
+
+	--- Controls if CPK should output logs
+	--- @type boolean
+	CPK_VERBOSE = false
+
+	local _lua_type = type
+	local _lua_next = next
+	local _lua_error = error
+	local _lua_tostring = tostring
+
+	local _lua_string_match = string.match
+	local _lua_table_insert = table.insert
+	local _lua_table_concat = table.concat
+
+	local _lua_getmetatable = getmetatable
+	local _lua_setmetatable = setmetatable
+
+	local _civ_include = include or (function() -- emulates include in none civ5 environment
+		--- @param dir string
+		--- @return string[]
+		local function GetFilePaths(dir)
+			local stream = io.popen('find "' .. dir .. '" -type f')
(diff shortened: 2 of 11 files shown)


azum4roll:
A few things:

DisplayPercentage can be easily be replaced by `{1: number #.#}%` in the text entry that formats a number with 1 decimal place, and `Locale.Lookup("TXT_KEY_SOMETHING', math.floor(x * 100))` in Lua. The current uses are being gradually converted as I rewrite the Lua files.

How useful is `StringBuilder` over what Lua already offers with `string.format`, `..`, `table.insert`, and `table.concat`?

Global variables in Civ5 are still scoped under each Lua context. If you want the modules to actually load only once in game, you can put them under the `ModMapData` table which seems to be shared across contexts. However, `ModMapData` is not available in PreGame just like other DLL globals, so there should be a fallback to `_G` when using it.

(You can also use other DLL globals like `GameInfoTypes`, but `ModMapData` is more commonly used by modders since it's mostly unused otherwise.)

azum4roll:
Oh and there's no need to have `.lua` in `include`.

vaderkos:
> Oh and there's no need to have `.lua` in `include`.

@azum4roll That's to be sure that correct file is included.
[Usually only the first eight characters are checked. This is likely only true if you do not specify the path. This means that if you registered two files named MapScannerA.lua and MapScannerB.lua, using include("MapScannerA") could either load the A file or the B file!](https://modiki.civfanatics.com/index.php/Specificities_of_the_Lua_implementation_in_Civ5)
 
> How useful is `StringBuilder` over what Lua already offers with `string.format`, `..`, `table.insert`, and `table.concat`?

Same thing with useful abstraction. In huge files it's hard to determine what's going on where `table.insert` or worse plain `..` concatenations are used. 

> Global variables in Civ5 are still scoped under each Lua context. If you want the modules to actually load only once in game, you can put them under the `ModMapData` table which seems to be shared across contexts. However, `ModMapData` is not available in PreGame just like other DLL globals, so there should be a fallback to `_G` when using it.

I forgot to mention, once **per lua context**.
Could you elaborate more on `ModMapData` or how to use it.. Never heard of it.

vaderkos:
> DisplayPercentage can be easily be replaced by `{1: number #.#}%` in the text entry that formats a number with 1 decimal place, and `Locale.Lookup("TXT_KEY_SOMETHING', math.floor(x * 100))` in Lua. The current uses are being gradually converted as I rewrite the Lua files.

@azum4roll Isnt it just better to just use `Locale.ToPercent`
image

azum4roll:
> > DisplayPercentage can be easily be replaced by `{1: number #.#}%` in the text entry that formats a number with 1 decimal place, and `Locale.Lookup("TXT_KEY_SOMETHING', math.floor(x * 100))` in Lua. The current uses are being gradually converted as I rewrite the Lua files.
> 
> @azum4roll Isnt it just better to just use `Locale.ToPercent` image

You still want to be controlling the number of decimal places in the text key either way.

azum4roll:
> > Global variables in Civ5 are still scoped under each Lua context. If you want the modules to actually load only once in game, you can put them under the `ModMapData` table which seems to be shared across contexts. However, `ModMapData` is not available in PreGame just like other DLL globals, so there should be a fallback to `_G` when using it.
> 
> I forgot to mention, once **per lua context**. Could you elaborate more on `ModMapData` or how to use it.. Never heard of it.

It's `_G`, but shared across contexts while in game (it seems to be unavailable in pregame menus). I also added code to remove cached data from the table when the `InGame` context shuts down, just in case it doesn't. (And indeed, somehow `GameInfoTypes` table becomes `nil` when using RESTART GAME without this piece of code.)

```lua
ContextPtr:SetShutdown(function ()
	print("Shutting down InGame.lua");
	MapModData.VP = nil;
end);
```

vaderkos:
> > > Global variables in Civ5 are still scoped under each Lua context. If you want the modules to actually load only once in game, you can put them under the `ModMapData` table which seems to be shared across contexts. However, `ModMapData` is not available in PreGame just like other DLL globals, so there should be a fallback to `_G` when using it.
> > 
> > 
> > I forgot to mention, once **per lua context**. Could you elaborate more on `ModMapData` or how to use it.. Never heard of it.
> 
> It's `_G`, but shared across contexts while in game (it seems to be unavailable in pregame menus). I also added code to remove cached data from the table when the `InGame` context shuts down, just in case it doesn't. (And indeed, somehow `GameInfoTypes` table becomes `nil` when using RESTART GAME without this piece of code.)
> 
> ```lua
> ContextPtr:SetShutdown(function ()
> 	print("Shutting down InGame.lua");
> 	MapModData.VP = nil;
> end);
> ```

I'm not sure what you mean?
Correct me if I'm wrong...

There is a global table ModMapData(that is de-facto _G while in game) that's shared among contexts,
except contexts that are in PreGame like `Main State` or ???
And there are some quirks related to it?

vaderkos:
@azum4roll So you say I could do something like this and have globally shared CPK.

`CPK.lua`
```lua
local CPK = {} -- imagine autoloader

MapModData.CPK = CPK
```
And this is shared among all states?

azum4roll:
> @azum4roll So you say I could do something like this and have globally shared CPK.
> 
> `CPK.lua`
> 
> ```lua
> local CPK = {} -- imagine autoloader
> 
> MapModData.CPK = CPK
> ```
> 
> And this is shared among all states?

You'll need a nil check and a fallback to use _G but basically yes.

azum4roll:
I'll merge this first. No idea what most of those utilities are for, but they don't take up much space.

azum4roll:
What's the `IncludeFileList` and `RefreshIncludeFileList` in CPK.lua?

vaderkos:
> What's the `IncludeFileList` and `RefreshIncludeFileList` in CPK.lua?

@azum4roll Inside `Main State`
`IncludeFileList` is a table that contains all `lua` files available for inclusion (at least it's what I observed by checking what is available where). I haven't seen it in other states.
image

It's sorted from latest to earliest registered files
image

I only assumed `RefreshIncludeFileList` updates the files inside `IncludeFileList` for example when game loads mods.

azum4roll:
I'm getting this when I try the kit.

image

The file is in modinfo.

vaderkos:
> I'm getting this when I try the kit.
> image
> 
> The file is in modinfo.

@azum4roll Should be `CPK.Misc.AsPercentage` ? Yeah I've missed it in Vox Populi's TechHelpInclude as it's removed in PR with TechHelpInclude optimization.

azum4roll:
Can you fix the readme? Files like `CPK.Util.lua` don't seem to be required.

azum4roll:
image

Getting these when attempting to `Disable` multiple controls in the same call.

vaderkos:
> image
> 
> Getting these when attempting to `Disable` multiple controls in the same call.

@azum4roll Could elaborate more about the issue? It'd be very helpful if you include a code snippet.

I've tested it and looks like it works

image

azum4roll:
Nevermind, it was due to accidentally calling `SetDisabled` with `.` instead of `:`

PR #11862, opened 2025-08-03
Update LeaderFlavorSweeps.sql to increase mobile for Mongols

ron-murhammer:
Mongols have probably the most mobile kit and history of any civ so should probably be a 10 in mobile. This should push them to build mostly cavalry and skirmishers which is what they have bonuses for.

### DIFF
--- a/(2) Vox Populi/Database Changes/AI/LeaderFlavorSweeps.sql
+++ b/(2) Vox Populi/Database Changes/AI/LeaderFlavorSweeps.sql
@@ -361,6 +361,7 @@ UPDATE Leader_Flavors SET Flavor = 7 WHERE FlavorType = 'FLAVOR_DIPLOMACY' AND L
 
 UPDATE Leader_Flavors SET Flavor = 10 WHERE FlavorType = 'FLAVOR_MILITARY_TRAINING' AND LeaderType = 'LEADER_GENGHIS_KHAN'; -- 8, primary: dom
 UPDATE Leader_Flavors SET Flavor = 10 WHERE FlavorType = 'FLAVOR_OFFENSE' AND LeaderType = 'LEADER_GENGHIS_KHAN'; -- 8, primary: dom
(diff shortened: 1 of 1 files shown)


RecursiveVision:
https://forums.civfanatics.com/threads/some-remarks-on-the-mongols.699304/#post-16851354

ron-murhammer:
Hmm. The flavor file increases a number of civs `FLAVOR_MOBILE` and comments that its because of skirmisher, so is that just incorrect then (example is https://github.com/LoneGazebo/Community-Patch-DLL/pull/11862/files#diff-7135644b9c848416346db9128687941cd30d5dd3ef9da60a8761732f57ada2a7L377)? Can `FLAVOR_SKIRMISHER` be added in this file as I didn't see it used at all? 

I'm wondering if the better change here is to take a pass on setting appropriate `FLAVOR_MOBILE` and `FLAVOR_SKIRMISHER` across the personality types and leaders in this file then?

axatin:
`FLAVOR_SKIRMISHER` only works in balparmak's unit production AI mod, which we could maybe integrate into the main mod at some point (and update the leader flavors while doing so). For the mongolia issue the better solution would be if unit production AI took into account free promotions when evaulating the value of a unit.

ron-murhammer:
@axatin Ok, would it be useful to at least remove the `FLAVOR_MOBILE` from `LeaderFlavorSweeps.sql` where it is being added for the reason of a skirmisher UU? As that I think is confusing when reading the file and could be actually making the AI worse (choosing to build too many melee cavalry instead of skirmishers). Also given Mongolia's UU, I would still think they should be a 8 or 9 `FLAVOR_MOBILE`, no?

axatin:
Yes, `FLAVOR_MOBILE` shouldn't be used for the skirmisher UUs. It might actually be detrimental, making the AI use their horses for mounted melee instead of skirmishers, so we shouldn't give Mongolia a high value for it

azum4roll:
In general, we shouldn't be increasing flavors for UUs since AI already has a higher score on them.

PR #11863, opened 2025-08-04
Update CityStateStatusHelper.lua

jarcast:
City State status tooltip is not compatible with civs plopping unique embassies (improvements with column IsEmbassy=1) in CS territory. In detail, it displays "embassy available" erroneously when it should be not. I edited the relevant part and fixed the issue (I tested it in game).


axatin:
Please update the non-EUI version as well.

Is the variable `iEmbassy` still needed?

jarcast:
iEmbassy is not anymore needed.

Where are the non EUI files? I cannot see them.

Screenshot_2

axatin:
(2) Vox Populi/LUA/CityStateStatusHelper.lua, the changes are probably the same as for the EUI version

azum4roll:
Update the (1) one too. Or maybe not and we wait for the rewrite to finally make embassies work in CP-only (no rush).

jarcast:
Done.

PR #11864, opened 2025-08-04
add more Lua documentation

azum4roll:
This contains all changes in #11856


LuaCATS/Global.d.lua
```diff
@@ -1,5 +1,9 @@
 --- @meta
 
+--- A super global table that persists across all Lua contexts, useful for sharing data between them
+--- @type table
+MapModData = {}
```
vaderkos, line 5: I'd rather say it's `table | nil`

LuaCATS/Game.d.lua
```diff
@@ -0,0 +1,13 @@
+--- @meta
+
+--- @class Game
+--- @field [string] function
+Game = {}
```
vaderkos, line 5: I'm wondering if we can define it as `Game | nil`. There are a lot of things that do not exist everywhere...

LuaCATS/Control.d.lua
```diff
@@ -1,3 +1,6 @@
 --- @meta
 
---- @class Control : Context
+--- @class ControlBase: UIElement
```
vaderkos: Is there any difference between `ControlBase` and `Control`?

LuaCATS/Game.d.lua
azum4roll: It's context-dependent. I don't want to add nil checks where they're not needed.

LuaCATS/Control.d.lua
azum4roll: Modiki defines it as so. I think the site was at least partially sourced from the defunct 2k site, so we'll have to trust it on those inaccessible code.

azum4roll: I searched all Civ5 Lua files and there isn't any global variables named `Control` being used, so it should be safe to use it as the class name.

azum4roll:
Reverted `ControlBase` to `Control`, also removed `UIElement` as it seems to be identical to `ControlBase`. `Context` is now a subclass of `Control`, which should be the superclass of all tags in a UI xml file.

image

(1) Community Patch/Kit/CPK.lua
```diff
@@ -28,7 +28,7 @@
 
 	local EmulateInclude = (function()
 		local function GetAvailablePaths()
-			local list = IncludeFileList --[[ @as string[] | nil ]]
```
vaderkos, line 31: Spaces were added because editor freaked out when it was just `string[]` like in this cast `--[[@as string[]]]`, because it messed lua comment and added extra `]` after it

LuaCATS/Network.d.lua
```diff
@@ -0,0 +1,15 @@
+--- @meta
+
+--- Contains functions that send network messages to every player.<br>
+--- Note that these are also used in single player.<br>
```
vaderkos, line 4: Why?

LuaCATS/Pseudo.d.lua
```diff
@@ -1,8 +1,30 @@
...
 
+--- Below refer to the ingame instance IDs
 --- @alias PlayerId integer
 --- @alias AreaId integer
 --- @alias CityId integer
 --- @alias PlotId integer
 --- @alias TeamId integer
 --- @alias UnitId integer
+
+--- Below refer to the database row IDs
```
vaderkos, line 11: How to call then string types in code? like `TECH_GUILDS` that's inside `Type` Column of `Technologies`
image

azum4roll: They're just strings. There's no need to alias them.

LuaCATS/enum/ButtonPopupTypes.d.lua
```diff
@@ -1,4 +1,123 @@
 --- @meta
 
 --- @enum ButtonPopupType
-ButtonPopupTypes = {}
\ No newline at end of file
+ButtonPopupTypes = {
```
vaderkos, line 4: It would be awesome if such files could be generated from C code that create that enums...

LuaCATS/Network.d.lua
azum4roll: Even in single player, player actions from the UI have to go through the network code back to the game core.

(1) Community Patch/Kit/CPK.lua
azum4roll: There's a nil after that so it works.

LuaCATS/enum/ButtonPopupTypes.d.lua
azum4roll: Should be possible to write a Lua script to print it out.

azum4roll:
As there's simply no way to tell which subclass of `Control` each of them is from the Lua perspective, I'm going to document every method as under the `Control` class. Developers would need to manually make sure the caller UI component has implemented the called method.

For example, only `Grid` and `Image` have the `DoAutoSize()` method, but LuaLS isn't going to complain if you do `instance.Button1:DoAutoSize()`, even if it likely won't work.

LuaCATS/Control.d.lua
```diff
@@ -1,3 +1,75 @@
 --- @meta
 
---- @class Control : Context
+--- @class Control
+--- @field [string] function
```
vaderkos, line 4: @azum4roll You're getting problems with luals **not complaining** about the functions that do not exist because of this line. Because you made each key that is string to be function

vaderkos:
> As there's simply no way to tell which subclass of `Control` each of them is from the Lua perspective, I'm going to document every method as under the `Control` class. Developers would need to manually make sure the caller UI component has implemented the called method.
> 
> For example, only `Grid` and `Image` have the `DoAutoSize()` method, but LuaLS isn't going to complain if you do `instance.Button1:DoAutoSize()`, even if it likely won't work.

That's because https://github.com/LoneGazebo/Community-Patch-DLL/pull/11864/files#r2256662244

I think separation is very important here, because otherwise we loose the main idea of types and're straight lying about the types leading to runtime errors!

LuaCATS/database/GameInfo.d.lua
```diff
@@ -0,0 +1,92 @@
+--- @meta
+
+--- Usage:<br>
+--- `GameInfo.Units()` returns an iterator that loops through all rows in the `Units` table<br>
```
vaderkos, line 4: You can enclouse this into lua block, it'll look nicer inside ide
\`\`\`lua

\`\`\`

vaderkos:
@azum4roll I think everywhere where we have dynamic fields, we should have

```lua
--- @field [string] <something> | nil 
--- @field [integer] <something> | nil
```
instead of 
```lua
--- @field [string] <something>
--- @field [integer] <something>
```

So during development we're aware that checks are required!

.luarc.json
```diff
@@ -2,7 +2,10 @@
   "workspace": {
     "library": [
       "./LuaCATS",
-      "./(1) Community Patch/Kit"
+      "./(1) Community Patch/Kit",
+      "./(1) Community Patch/Core Files/VPUI",
+      "./(1) Community Patch/Core Files/Overrides/Includes",
+      "./(1) Community Patch/LUA/Includes"
```
vaderkos, line 8: That would imply everything in those files is available in every context, which isn’t accurate... We have to live with it...

What's the difference between `LUA/Includes` and `Core Files/Overrides/Includes` ?

`LUA/Includes` is custom?
`Overrides/Includes` is base game files overriden?

vaderkos: What about adding such comment to each global in these includes?

```lua
--- **This global function is only available after calling** `include("<FILENAME>")` **in your script.** <br>
function FUNCTION_NAME() end
```

```lua
--- **This global variable is only available after calling** `include("<FILENAME>")` **in your script.** <br>
MyVar = BLABLA
```

vaderkos: Looks nice
image

LuaCATS/Control.d.lua
azum4roll: Intended. We're unlikely to ever document every single one of them (there are hundreds), so this would nicely get rid of all the false positives for now. Treating a `Control` field as anything other than function will still trigger a warning.

.luarc.json
azum4roll: > What's the difference between `LUA/Includes` and `Core Files/Overrides/Includes` ?
> 
> `LUA/Includes` is custom? `Overrides/Includes` is base game files overriden?

LUA folders (from both (1) and (2)) are removed from the installation of EUI version, or else they'll override (base) EUI files.

So any file that's in UI_bc1 would be in LUA, and other vanilla files would be in Overrides.

azum4roll:
> ```lua
> @field [string]
> ```

We should never need to nil check any of them, unless `[string]` is dynamic.

Imagine having to check `GameInfo.Technology_DomainExtraMoves` is not nil before iterating over it...

LuaCATS/Control.d.lua
vaderkos: But we're very unlikely to document them at all then. **Warning** motivates developers to document functions they're using. Then gradually we would have documented it. Currently warning would be everywhere, but step-by-step it could be covered.
Plus we could do as I did in `UI` and `UIManager`. Assign desired control to variable and using FireTuner just document each function, or at least just define them with `TODO docs`.

PR #11869, opened 2025-08-06
Add 61 New & Restored City-States

RecursiveVision:
1. Implements proposal (6-70), adding 42 City-States: https://forums.civfanatics.com/threads/6-70-add-new-city-states-when-major-civs-are-absent-ingame.686300

2. Reactivates 19 previously existing City-States disabled by Firaxis. This also restores some otherwise unused City-State screen audio clips.
- Makes use of existing assets that were not removed from the game database, and also restores text from very old Civ V versions for three City-States that were later replaced (Oslo, Copenhagen, and Seoul).

- Fixes a BNW bug where Cape Town would use Almaty's audio clip. Note that modifying the base game files is required for audio clip assignment to work correctly, both to fix this and to add clips for any new City-States.

- Applies a custom color to City-States affected by MajorBlocksMinor, changing them to use the same secondary color as their associated civ's background color in most cases.
  - In cases where the contrast with black is insufficient or the color is too close to Barbarians' red, a brightened or replacement color is used instead.

- Expands the Civilopedia entries of the Scramble for Africa scenario City-States for parity with the other City-States.

- Updates Hong Kong and Singapore's Civilopedia entries to reflect recent developments.

- Fixes a typo in Kyiv's Civilopedia entry.


azum4roll:
The database entries for ArtStylePrefix don't actually have the trailing space. I think strings from XML are always trimmed.

(2) Vox Populi/Database Changes/Minors/NewMinorColors.sql
```diff
@@ -0,0 +1,95 @@
+-- For the City-States added to the MajorBlocksMinor table, add a custom color wherein they use their associated major civ's background color instead of the normal CS colors.
+-- This is a cool way to highlight that these are special City-States while still keeping it subtle.
+CREATE TEMP TABLE NewCityStateColors (
+	CityStateColorName text,
+	CivBackgroundColor text
+);
+
+-- 15 out of the 43 civs have a dark background color which needs to be brightened to contrast well against black.
```
azum4roll: Please use NewColors.xml for this.

RecursiveVision:
Awaiting fact-checking on Zuunmod and the Scramble for Africa City-States, otherwise this is ready to be merged.

RecursiveVision:
My fact-checker is unavailable indefinitely; however, since Civilopedia text is going to be missing anyway in 5.0, a few minor historical errors aren't too big a deal. Will follow up on this eventually.

@axatin When the autoinstaller is used to uninstall VP, `Expansion2_Base.Civ5Pkg` needs to replace `Expansion2.Civ5Pkg` and the `MinorCivSounds_VoxPopuli.xml` file needs to be deleted.

RecursiveVision:
Please use Rebase and merge when this is merged. :)

PR #11872, opened 2025-08-08
UI rewrite and Lua documentations

azum4roll:
This is a huge UI rewrite for both InfoTooltipInclude and TechButtonInclude. Most (95%+) applicable columns and tables have been accounted for when generating tooltips for units, buildings, and techs (improvements, projects, processes coming soon).

VPUI mostly serves as the replacement of EUI library to save memory yet still caching commonly-used primary table columns. There should be little reason to iterate through the entire Buildings table now (`GameInfo.Buildings()`), or grabbing full rows out of it (`GameInfo.Buildings["BUILDING_FOO"]`). Use `(MapModData.)VP.GameInfoCache(Buildings)` and `(MapModData.)VP.GetInfoFromType("Buildings", "BUILDING_FOO")` instead to avoid querying the database.

This includes all commits from #11856 and #11864.


axatin:
if `TechButtonInclude` and `InfoTooltipInclude` are used for both (1) and (2), why are they in `LUA\Includes` and not in `Core Files\Includes`?

azum4roll:
> if `TechButtonInclude` and `InfoTooltipInclude` are used for both (1) and (2), why are they in `LUA\Includes` and not in `Core Files\Includes`?

That'll override EUI's files. TechButtonInclude may not matter since it's overridden again by (3a), but InfoTooltipInclude doesn't have a (3a) version.

axatin:
If they're in LUA they won't be available in non-EUI VP.

azum4roll:
> If they're in LUA they won't be available in non-EUI VP.

Non-EUI VP shouldn't be removing the (1) LUA folder. If it's the case right now, it should be fixed.

axatin:
are you sure? that's how it's been for years. then we need to change the manual installation instructions and the installer

azum4roll:
The Lua files in (2) should properly override those in (1) when the mods are loaded (and the ones in the Overrides folder already do). It's only the Lua files in DLC that need to be specially handled.

PR #11874, opened 2025-08-08
Fix issue template

axatin:


azum4roll:
I see nothing in new issues.

axatin:
Yes, that didn't fix it... Trying to find the cause of the issue

PR #11882, opened 2025-08-14
Fixed the issue where the commit_inc.inc fails to write when the solution folder contains whitespaces.

CIVITAS-John:
A very minor change :)

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvConnectionService.cpp
+++ b/CvGameCoreDLL_Expansion2/CvConnectionService.cpp
@@ -248,8 +248,12 @@ DWORD WINAPI CvConnectionService::NamedPipeServerThread(LPVOID lpParam)
 {
 	CvConnectionService* pService = static_cast<CvConnectionService*>(lpParam);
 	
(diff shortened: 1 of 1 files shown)


RecursiveVision:
What is the purpose of the CvConnectionService files? The pull request title does not seem particularly descriptive.

CIVITAS-John:
Hi! Sorry I must have mistakenly sent this pull request. But I want to learn about how to do it since it is better if those things get integrated. Basically, my changes provide capabilities for the DLL to expose certain things to an HTTP service via IPC:
- Calling an arbitrary Lua script or pre-registered functions;
- Allowing Lua script to call arbitrary external functions (exposed as HTTP endpoints);
- Streaming DLL events for subscription. 
The goal is to enable integration of LLMs within Civ 5's AI (more or less, not taking over everything).

RecursiveVision:
If that's the case I'll merge your intended change manually.

I'm not sure of what benefit integrating ChatGPT or a similar program would be, unless you meant to say some kind of machine learning. Although maybe ChatGPT could produce responses for the Diplomacy AI.

PR #11884, opened 2025-08-14
Add backwards compatibility to HandicapInfos

RecursiveVision:
~~Removes DifficultyChanges.sql, separating it into one file for the Ancient Ruins changes and five files for the triggered instant yields, one file for each difficulty.~~

~~This should now be way easier to modify for the end user compared to having to do math. And it's easy enough to read from the file if someone wants to do difficulty bonus related proposals.~~

Adds backwards compatibility to some renamed but previously used columns in the HandicapInfos table - if these normally unused columns are used by a modmod, their effects will stack with those of the new columns.

The columns unfortunately cannot be simply deleted by remaking the table, since remaking the table creates errors due to external code.

### DIFF
--- a/(1) Community Patch/Database Changes/Difficulty/CoreDifficultyChanges.xml
+++ b/(1) Community Patch/Database Changes/Difficulty/CoreDifficultyChanges.xml
@@ -16,7 +16,9 @@
 	- AITrainPercent
 	- AIWorldTrainPercent
 	- AIConstructPercent
+	- AIWorldConstructPercent
 	- AICreatePercent
+	- AIWorldCreatePercent
 -->
 <GameData>
 	<HandicapInfos>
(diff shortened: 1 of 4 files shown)


azum4roll:
To me this is actually harder to read and modify, since all relations between difficulties and between historic event types are now lost.

RecursiveVision:
My opinion is that the end user should not have to do math to modify the difficulty levels. Am fine with adding some kind of documentation as to how the numbers came to be.

It's up to @axatin to make the call.

axatin:
I tend to agree with azum here. The current approach shows the logic behind the difficulty bonuses more clearly, and the numbers can be modified quite easily by changing either base values or multipliers. If a modder is unsure about how the code works and wants to know the final values, they can simply look them up in the database.

RecursiveVision:
Mmm. I'm thinking more in terms of end users than modders, but on thinking about it, this would be a pain to modify if we push a change to the whole bunch of values like +20%. Very well.

PR #11904, opened 2025-09-08
4UC Egypt

RecursiveVision:


Hokath:
Update to map trading, nice.
I've personally moved it earlier to Navigation, and I saw you wrote
"- Add compatibility code for deal valuation if a modder makes map trading available earlier"
what happened before?

RecursiveVision:
Map valuation is now reduced by 60% in Ancient, 40% in Medieval, and 20% in Renaissance. That wasn't the case earlier.

Probably best discussed on the RSI pull request thread though. :)

Issue #11907, opened 2025-09-11
C4DF features should not be locked by MOD_BALANCE_VP.


mwhiter:
_1. Mod Version (X.Y.Z). Current Version: 4.22_
4.22

_2. DirectX Version_
N/A

_3. List of Other Mods_
N/A

_4. Describe the Issue_
I noticed that the C4DF code that affects trade item eligibility is locked behind the `MOD_BALANCE_VP` flag. This is incorrect; these items should be locked behind the following requirements:
1. Map Trading: A tech that has `MapTrading` enabled
2. Tech Trading: "No Tech Trading" game option is OFF, "No Science" game option is OFF, and a tech has `TechTrading`
3. Vassalage: "No Vassalage" game option is OFF (note: for Community Patch, this game option must be set ON to ensure vanilla compatibility)
4. Vassalge Revoke: "No Vassalage" game option is OFF (note: for Community Patch, this game option must be set ON to ensure vanilla compatibility)

Changes where this needs to happen:
**DLL**
1. `IsPossibleToTradeItem`
2. `GetReasonsItemUntradeable`

______________________________________________________________________________________
_5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)_
N/A

_6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)_
N/A

_7. CvMiniDump.dmp File (ATTACH IF REPORTING A GAME CRASH)_
N/A

_8. Steps to reproduce the Issue (Optional)_
N/A

_9. Screenshots of the Issue (Optional)_
N/A

RecursiveVision:
@mwhiter I do not completely remember Younger Me's reasoning, and let's be fair, that guy's an idiot so I can't be held accountable for his decisions. :)

But I would _guess_ that his reasoning went something like this:
- In version 2.0, the "Great Merge", we merged (2), (3), (4), and (5), so I was like "okay, so these features are always on", we don't need separate C4DF and CSD defines anymore, just use MOD_BALANCE_VP for everything.
- We merged the Share Opinion and Gifts stuff into Community Patch Only as those are AI improvements, not balance changes. The vassalage system is a balance change so we locked it behind VP along with all the other balance changes.
- I also likely didn't want to loop through all techs multiple times per turn to see if one of them unlocked maps or techs. This could be done by running the check once during game initialization and caching the value.
- Community Patch Only's LUA files don't have compatibility with vassalage code.

[...]

PR #11911, opened 2025-09-12
Remove MOD_BALANCE_VP for C4DF features

mwhiter:
Intended fix for #11907 

⚠️ BREAKING CHANGE ⚠️ 

Removes `MOD_BALANCE_VP` checks for all C4DF features.

For Community Patch, I am forcing GameOptions to be set on startup that disables the features. This is technically a breaking change, as I think previous saves will not have the correct GameOptions set and this may affect them. I am doing this by adding two new Lua files, which are copies of the original Civ 5 lua files with the modifications to set the GameOptions. `AdvancedSetup.lua` and `GameSetupScreen.lua`.

For Vox Populi, I am adding `AdvancedSetup.lua` which is an exact copy of the Civ 5 screen. This undoes the code to set the GameOptions, as now we have the game options available for editing. `GameSetupScreen.lua` is already overridden, so no changes are necessary.

# Community Patch only
image
image

# Vox Populi
image


azum4roll:
How I would've done this:

1. Change the existing No Tech Trading and No Vassalage to Enable Tech Trading and Enable Vassalage, and have them be default enabled in (2).
2. Move Transparent Diplomacy to (1). It should already work properly.
3. No need to change anything in Lua.

Even with the existing No ... options, I think you could've set them to Visible = 0 and Default = 1 in (1)?

axatin:
> Even with the existing No ... options, I think you could've set them to Visible = 0 and Default = 1 in (1)?

I'm not sure if that combination works, needs to be tested. If it does, it would indeed be the best solution: Don't make any changes in lua, move the game options from (2) VP/Database/NewGameOptions.xml to the corresponding file in (1) with Visible = 0 and Default =1 or = 0, use SQL to change the values in VP.

RecursiveVision:
Have not had a chance to review this yet, but:

> I'm not sure if that combination works, needs to be tested. If it does, it would indeed be the best solution: Don't make any changes in lua, move the game options from (2) VP/Database/NewGameOptions.xml to the corresponding file in (1) with Visible = 0 and Default =1 or = 0, use SQL to change the values in VP.

I would likely have done something similar: move them to (1), change them to "Enable" rather than "Disable", and switch them on by default in (2). If the goal is compatibility to facilitate C4DF standalone, making them invisible is not necessarily needed if it's disabled by default. Although if you want to have a separate mod for that, then yes, I'd make them invisible.

When I said the (1) LUA files were incompatible with LUA I was referring to the Trade screen (TradeLogic.lua), not Game Options.

mwhiter:
Okay I see what you are saying. I'll move this back into draft.

> Is this whole rewrite of AdvancedSetup.lua and GameSetupScreen.lua I don't see any diff in PR?

The changes are turning on game options for (1) and restoring the original file in (2) but I'll remove these as not needed. The files didn't exist before as not needed so the full diff is just adding the vanilla file with some changes.

New plan:
1. Update the game options to `ENABLE_TECH_TRADING`, `DISABLE_RESEARCH_AGREEMENTS`, `ENABLE_VASSALAGE`.
2. Keep `NO_TECH_BROKERING`
3. Update (2) to set `ENABLE_TECH_TRADING`, `DISABLE_RESEARCH_AGREEMENTS`, `ENABLE_VASSALAGE` to ON by default
4. No change to `NO_TECH_BROKERING`

> When I said the (1) LUA files were incompatible with LUA I was referring to the Trade screen (TradeLogic.lua), not Game Options.
My original thought was to not let them enable those features as it's game-breaking but I can just make those features enabled in (1) as originally done via the mod. So I'll port over `TradeLogic` and any other files that reference vassalage (`DiploGlobalRelationships` maybe? Gotta look through the files again...) so that feature works.

mwhiter:
This PR should be in a much better place now, thanks for the feedback y'all.

(1) Community Patch/Database Changes/CoreNewGameOptions.xml
```diff
@@ -42,5 +42,40 @@
...
+			<Default>0</Default>
+		</Row>
+		<Row>
+			<Type>GAMEOPTION_TRANSPARENT_DIPLOMACY</Type>
+			<Description>TXT_KEY_GAME_OPTION_TRANSPARENT_DIPLOMACY</Description>
+			<Help>TXT_KEY_GAME_OPTION_TRANSPARENT_DIPLOMACY_HELP</Help>
+			<Visible>0</Visible>
+			<Default>0</Default>
+		</Row>
 	</GameOptions>
```
RecursiveVision, line 80: Typically in XML we use true/false instead of 1/0 for clarity. In SQL this doesn't work so we use 1/0.

Also, no need to specify the value of a column if it's already the same value as the default value for that column. Based on the game options above, the default value for `Default` seems to be `false` and the default value for `Visible` seems to be `true`. So, only `Visible` needs to be specified.

Also also, you can make Transparent Diplomacy visible in Community Patch Only. It's not a balance change.

(2) Vox Populi/Database Changes/NewGameOptions.xml
```diff
@@ -6,32 +6,41 @@
...
+			<Visible>1</Visible>
+		</Replace>
+		<Replace>
 			<Type>GAMEOPTION_TRANSPARENT_DIPLOMACY</Type>
 			<Description>TXT_KEY_GAME_OPTION_TRANSPARENT_DIPLOMACY</Description>
 			<Help>TXT_KEY_GAME_OPTION_TRANSPARENT_DIPLOMACY_HELP</Help>
-		</Row>
+			<Default>0</Default>
+			<Visible>1</Visible>
+		</Replace>
```
RecursiveVision, line 43: Typically we use an SQL file rather than XML replace because it's more clear for moddability. See `CoreGameOptionChanges.sql` in (1). I would create a new `GameOptionChanges.sql` file here in (2).

See also comment on `CoreNewGameOptions.xml`.

(2) Vox Populi/LUA/TradeLogic.lua
```diff
@@ -20,10 +20,10 @@ local g_bAlwaysPeace = Game.IsOption( GameOptionTypes.GAMEOPTION_ALWAYS_PEACE );
...
 
 local g_PeaceDealDuration = GameInfo.GameSpeeds[PreGame.GetGameSpeed()].PeaceDealDuration;
-local g_bAllowResearchAgreements = Game.IsOption("GAMEOPTION_RESEARCH_AGREEMENTS");
+local g_bAllowResearchAgreements = not Game.IsOption("GAMEOPTION_DISABLE_RESEARCH_AGREEMENTS");
 local g_bDisableScience = Game.IsOption("GAMEOPTION_NO_SCIENCE");
-local g_bDisableTechTrading = Game.IsOption("GAMEOPTION_NO_TECH_TRADING");
-local g_bDisableVassalage = Game.IsOption("GAMEOPTION_NO_VASSALAGE");
+local g_bDisableTechTrading = not Game.IsOption("GAMEOPTION_ENABLE_TECH_TRADING");
+local g_bDisableVassalage = not Game.IsOption("GAMEOPTION_ENABLE_VASSALAGE");
 local g_bDisableLeague = Game.IsOption("GAMEOPTION_NO_LEAGUES");
```
RecursiveVision, line 28: You'll want to port this over to (1)'s version of TradeLogic.lua. Otherwise if someone enables the game option via Really Advanced Setup, they won't be able to make vassals, but the AI will.

In Community Patch Only, should probably hide the Vassalage trade option entirely if the game option isn't enabled.

May also want to add some code to turn on Vassal Overview only if the game option is enabled, for the same reason.

CvGameCoreDLL_Expansion2/CvGame.cpp
```diff
@@ -6458,7 +6458,7 @@ bool CvGame::IsShowHiddenOpinionModifiers() const
 /// This controls whether the AI should display the number value of each Opinion modifier in its table of modifiers.
 bool CvGame::IsShowAllOpinionValues() const
 {
-	if (IsDiploDebugModeEnabled() || (MOD_BALANCE_VP && isOption(GAMEOPTION_TRANSPARENT_DIPLOMACY)))
+	if (IsDiploDebugModeEnabled() || (isOption(GAMEOPTION_TRANSPARENT_DIPLOMACY)))
 		return true;
```
RecursiveVision, line 6462: Ew, those extra parentheses can go. >:o

RecursiveVision:
> You'll want to port this over to (1)'s version of TradeLogic.lua. Otherwise if someone enables the game option via Really Advanced Setup, they won't be able to make vassals, but the AI will.
> 
> In Community Patch Only, should probably hide the Vassalage trade option entirely if the game option isn't enabled.
> 
> May also want to add some code to turn on Vassal Overview only if the game option is enabled, for the same reason.

I would lean towards, instead of making/updating a C4DF standalone mod, just including all the functionality in Community Patch Only, due to the ability for modmods like Really Advanced Setup to expose game options that otherwise won't fully work.

Or if a standalone mod is used, it can simply set `Visible` and `Default` values for those game options. Note that the previous C4DF standalone mod is currently mod blocked by (1) Community Patch so you'd need to remove the mod blocker.

RecursiveVision:
Maybe we can add a (1a) mod to the repo.

RecursiveVision:
Alternative alternative idea: change the (1) text for the game options to something like "WARNING: Option is missing mod dependencies.", then update it in both (2) and (1a). Then nobody in RAS will activate it by mistake.

(2) Vox Populi/LUA/TradeLogic.lua
mwhiter: the scope of the vassalage changes are massive and should be its own PR honestly. I'm considering making it still subject to MOD_BALANCE_VP for now just to manage scope here. Otherwise there are hundreds of changes across tons of screens that need to come over to get that feature working

mwhiter:
There a lot of proposed options. I think it's a good long term goal to have vassalage enabled in Community Patch only. To manage scope here, I'll do the following:

1. Vassalage will remain a VP feature for now
2. Remove the vassalage game option from community patch only (hope this fixes really advanced setup mod)
3. Port the map and tech trading UI options to CPO. And add them to the various techs, but hide the game options. That way if someone uses really advanced setup, then they'll get use of the feature.

RecursiveVision:
@mwhiter I squashed all your commits and made the necessary edits in accordance with your proposition in the comment just above. See the `c4df-temp` branch.

Before this can be merged, the `TradeLogic.lua` (and probably another related xml file for the pocket hide/disable functions) needs to be reconciled between (2) and (1).

It's safe to port over *nearly* everything from (2) over to (1), although a few changes should be made:
- Vassalage should be either gated behind MOD_BALANCE_VP (ideal, then we can remove the (2) version entirely) or have the vassalage stuff removed from the (1) version.
- Techs and Maps should only be tradeable in (1) if the tech unlocking those trade options has been researched; otherwise, the entire option in the Trade screen should be hidden. Alternative solutions: looping through all techs and seeing if any unlock Map/Tech Trading, or (for techs) checking if the game option is enabled.
- I think there's some checks in the (2) version relating to the VP Dutch UA that will need to be adjusted as they're not applicable to (1).

@axatin pinging you as lead dev and also because you modified this file before
@azum4roll and you because you're also messing with lua files right now

azum4roll:
Moving all (2) Lua to (1) is the long-term goal. We don't have time to update 3 versions of UI files.

But until someone has time to rewrite the whole thing, small patches on all 3 versions is still the way to go.

axatin:
> * I think there's some checks in the (2) version relating to the VP Dutch UA that will need to be adjusted as they're not applicable to (1).

Oh, yeah, there is some very ugly hardcoding there
```
if (g_pUs:GetCivilizationType() == GameInfoTypes.CIVILIZATION_NETHERLANDS and (g_pUs:GetResourceExport(resType) > 0 or g_pUs:GetResourceImport(resType) > 0) and pResource.ResourceUsage == 2) then
				strString = "[ICON_CHECKBOX] " .. strString;
```
Instead of that, it should be checked if the civ has the trait `IsImportsCountTowardsMonopolies`, then it can also be used in (1).

mwhiter:
Thanks for updating this. I'm taking a bit of a break from modding so I won't be adding new commits to this for a bit. Feel free to make the necessary changes or I'll pick this back up when I'm back.

affom1:
Hello folks. This is not an actual comment just a random dude's praise from the internet. 
You people are doing god's work here 😘. Thanks for your awesome work - I am looking forward most excited for the new mod system 👍

RecursiveVision:
I'm not in the headspace to finish this but I rebased the c4df-temp branch for whoever does want to finish the Lua merge.

PR #11915, opened 2025-09-13
Change default AI on Conquistador

Hokath:
I noticed Spain is using their entire military supply to build Conquistador and then upgrading into recon unit. 
image

This means they becomes completely unable to fight, with the exception of Naval Units which seem to be handled seperately. Possibly explains poor performance in test games.

Discussed with KungCheops and he says the production AI reads the DefaultUnitAI column to decide what to build.  In lieu of a more sophisticated change, we have to stop Spain AI from committing sudoku every single game on UU unlock.

### DIFF
--- a/(2) Vox Populi/Database Changes/Civilizations/Spain.sql
+++ b/(2) Vox Populi/Database Changes/Civilizations/Spain.sql
@@ -83,7 +83,7 @@ SET
 		)
 	),
 	Combat = (SELECT Combat FROM Units WHERE Type = 'UNIT_EXPLORER') + 6,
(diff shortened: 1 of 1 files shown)


KungCheops (approved):

Hokath:
An alternative solution to this is for them to upgrade into a combat unit like a Fusilier or a Lancer.
Problem with Lancer is that it requires Horses.

RecursiveVision:
Should probably run a unit upgrade class change by the VP Congress.

azum4roll:
This will make Spain not build that many Conquistadors, and won't use Conquistadors to fight. I think this can be actually worse?

PR #11916, opened 2025-09-14
Fix reference to missing function

mwhiter:
A while ago, it seems a breaking change was made to the DLL without corresponding updates to (3a) EUI. This fixes it.


vaderkos:
@mwhiter I had very similar issue while developing lately, and someone said that the cause is incorrectly loaded DLL. I've managed to workaround [like this](https://github.com/LoneGazebo/Community-Patch-DLL/blob/d9b541f06bb0bf08db2b4c53b45db6aabae568b1/(1)%20Community%20Patch/Core%20Files/Overrides/Includes/TechHelpInclude.lua#L639), but is this the way...

mwhiter:
> @mwhiter I had very similar issue while developing lately, and someone said that the cause is incorrectly loaded DLL. I've managed to workaround [like this](https://github.com/LoneGazebo/Community-Patch-DLL/blob/d9b541f06bb0bf08db2b4c53b45db6aabae568b1/(1)%20Community%20Patch/Core%20Files/Overrides/Includes/TechHelpInclude.lua#L639), but is this the way...

I am using a DLL built from latest master. It appears this function is not defined. How would one get an incorrectly built DLL?

The code linked in that PR appears incorrect to me, since CP includes a custom DLL can't we enforce what functions are available? Do people swap out DLLs? That seems error prone to me.

axatin:
This PR is correct. A while ago I changed the function in the DLL and I didn't update EUI and CP only because azum was reworking files there. @azum4roll I thought you wanted to make the edits in the EUI version? that's why I documented the changes on discord (pinned message in my channel)...

> @mwhiter I had very similar issue while developing lately, and someone said that the cause is incorrectly loaded DLL. I've managed to workaround [like this](https://github.com/LoneGazebo/Community-Patch-DLL/blob/d9b541f06bb0bf08db2b4c53b45db6aabae568b1/(1)%20Community%20Patch/Core%20Files/Overrides/Includes/TechHelpInclude.lua#L639), but is this the way...

no, this is not necessary, GetResearchProgressTimes100 is always available when playing with our DLL. the issue you encountered was also caused by UI files that aren't updated yet

mwhiter:
If this PR looks good any objections to merging it?

azum4roll:
Unfortunately my game setup is going to be stuck in the non-EUI version until I'm done with the UI files. Switching between EUI and non-EUI simply takes too much effort.

PR #11919, opened 2025-09-14
4UC text import

Hokath:
Civ text keys and 'pedia text, including new entries for congress creations. Not tested so there may be formatting errors! 

Strategy text to come later from Alpakinator.

Someone needs to do promos. They don't go in this xml file and they all have different names now, it seems? I couldn't find them in order like the units and buildings so perhaps @azum4roll should do that part since he will remember what he did! :D


Hokath:
actually I can probably do promotions, the draft change log has them in order!

azum4roll:
There has to be a reason why it went unused?

azum4roll:
> actually I can probably do promotions, the draft change log has them in order!

Finish everything else first

Hokath:
Yeah I dunno, maybe? I assume it was left out as typo. As usual google doesn't show anything up for the name. 

> Finish everything else first

what still needs doing?

azum4roll:
> All pillaging is an improvement, after all.

It's not. You can pillage roads which aren't improvements. Very important to know (before people start complaining that it doesn't give yields)!

> Icon is preferable to colored text imo.

I agree, but only if there's an appropriate one.

Hokath:
That's an excellent point, I didn't actually realize that.
Does the 
"Pillaging costs no [ICON_MOVES] Movement and heals an extra [COLOR_POSITIVE_TEXT]10 HP[ENDCOLOR]."
work on roads too, then? What about XP.

Hokath:
https://github.com/IanE9/Community-Patch-DLL/blob/fd3a19d3371c93f2b07eaa1d6644da3ec285bc8d/CvGameCoreDLL_Expansion2/CvUnit.cpp#L10622

Seems the XP works for roads too, but not the HP, which is behind a if(bSuccessfulNonRoadPillage)
Although the indentation on this #if defined(HH_MOD_BUILDINGS_FRUITLESS_PILLAGE) is a bit confusing.

azum4roll:
> That's an excellent point, I didn't actually realize that. Does the "Pillaging costs no [ICON_MOVES] Movement and heals an extra [COLOR_POSITIVE_TEXT]10 HP[ENDCOLOR]." work on roads too, then? What about XP.

It's a bit of a wordplay. PIllaging (anything) costs no movement, and healing "extra" HP only applies if you already heal to begin with (so it would only apply to pillaging improvements).

azum4roll:
> https://github.com/IanE9/Community-Patch-DLL/blob/fd3a19d3371c93f2b07eaa1d6644da3ec285bc8d/CvGameCoreDLL_Expansion2/CvUnit.cpp#L10622
> 
> Seems the XP works for roads too, but not the HP, which is behind a if(bSuccessfulNonRoadPillage) Although the indentation on this #if defined(HH_MOD_BUILDINGS_FRUITLESS_PILLAGE) is a bit confusing.

The XP is improvements only too. Check the latest code.

Hokath:
Latest code?...
...
Oh I see, the PR I looked at gave me the link to the outdated fork. Wonderful.

Fixing text now

azum4roll:
This isn't meant to be merged... Oh well I'll just fix everything later.

Hokath:
Why not?
Trying to work with you is testing my psychic powers, azum. You don't explain anything willingly.

azum4roll:
I did say I was going to merge everything in myself. That avoids excessive commits to the master branch.

Issue #11920, opened 2025-09-15
Minor text error for Notre Dame Theme


Flamingcheesepie:
Notre Dame theme requires 2 Great Works of Art, made by 2 different civs that are not the actual owner. However, the current description of the string (TXT_KEY_BUILDING_NOTRE_DAME_THEMING_BONUS_HELP) is:

To maximize your bonus, make sure both Great Works were created by Players other than the owner.

Instead, it should also say that the Great Works are from separate players - probably something like

To maximize your bonus, make sure both Great Works were created by different Players other than the owner.

PR #11921, opened 2025-09-15
Update NewBuildingText.xml

Hokath:
notre dame theming bonus clarification #11920

we also need to give the actual theming bonus value somewhere.
@azum4roll will your auto-tooltip grab that?

### DIFF
--- a/(2) Vox Populi/Database Changes/Text/en_US/City/NewBuildingText.xml
+++ b/(2) Vox Populi/Database Changes/Text/en_US/City/NewBuildingText.xml
@@ -1105,7 +1105,7 @@
 			<Text>Holy Sanctuary</Text>
 		</Row>
 		<Row Tag="TXT_KEY_BUILDING_NOTRE_DAME_THEMING_BONUS_HELP">
-			<Text>To maximize your bonus, make sure both Great Works were created by Players other than the owner.</Text>
+			<Text>To maximize your bonus, make sure both Great Works were created by different Players other than the owner.</Text>
 		</Row>
 	</Language_en_US>
 </GameData>


azum4roll:
No, only the maximum theming yields (including the extra yields from Building_ThemingYieldBonus, and the Artistry boost in city view).

It's a bit too complicated to detail in a tooltip.

Hokath:
Ok maybe should be done in Help text then. 
We can revisit after 5.0 is working.

(2) Vox Populi/Database Changes/Text/en_US/City/NewBuildingText.xml
```diff
@@ -1105,7 +1105,7 @@
 			<Text>Holy Sanctuary</Text>
 		</Row>
 		<Row Tag="TXT_KEY_BUILDING_NOTRE_DAME_THEMING_BONUS_HELP">
-			<Text>To maximize your bonus, make sure both Great Works were created by Players other than the owner.</Text>
+			<Text>To maximize your bonus, make sure both Great Works were created by different Players other than the owner.</Text>
 		</Row>
```
RecursiveVision, line 1109: I would say "different foreign Civilizations".

vaderkos: After reading the previous version I finally understand what it meant...
Does it finally mean should be created by uniques civilizations other that players' civ?

RecursiveVision:
For additional clarity you might want to say "the Great Works" instead of "both Great Works" because "both" somewhat implies they're from the same foreign civ.

PR #11922, opened 2025-09-16
UI work

azum4roll:


azum4roll:
I don't have time to test every scenario for UI, but there are no lua errors, database errors or localization errors in the limited testing I have done.

TODO:
- TechTree.lua: Load on hover custom-sized tooltips, and generally rewrite
- InfoTooltipInclude.lua: Exclude potential boosts from unique components that are not from active player in non-generic mode
- Civilopedia.lua: Rewrite in (1)
- Civilopedia.lua: Remove non-icon panels (yields, etc.) from the left side of unit and building pages (they're all present in the new Game Info)
- Civilopedia.lua: Add more icon panels to the left side of unit and building pages to serve as links to their respective pages
- ~~LeaderHeadRoot.lua: Investigate why loading this breaks all global variables (game is unplayable without this fixed)~~
- ProductionPopup.lua: Merge production mode and purchase mode, add EUI-like purchase buttons on top of the production buttons
- CityView.lua: Make production queue infinitely long and drag-and-drop swappable like EUI
- CityView.lua: Show yields on buildings on the right. Hope info haters won't hate me for that

azum4roll:
Please don't squash

vaderkos:
I was working at the same time at `Civilopedia.lua` adding Fuzzy Search. Maybe we could split work here?

azum4roll:
Other than TechTree, other TODO changes have low priority and can wait until after 5.0.

I think you only need to modify `SearchForPediaEntry`?

vaderkos:
> Other than TechTree, other TODO changes have low priority and can wait until after 5.0.
> 
> I think you only need to modify `SearchForPediaEntry`?

That's the idea

vaderkos:
Can you do `format document` (`SHIFT+ALT+F` by default) action on new files so the formatting is consistent?

vaderkos:
Be careful with declaring classes, if you're trying to make something local try using `@package` or `@private` annotation to have limited scope. I'm looking at `Info` and wondering where it comes from..

vaderkos:
Can you move instead of creating new files in git for the files in `LUA` dir. It's hard to determine the changes in `TechTree.lua` and other files that were created instead of moved.

azum4roll:
`Info` isn't local. It's a `Row` that's returned by `GameInfo.<table>()` where `<table>` is a primary table.

> Can you move instead of creating new files in git for the files in LUA dir. It's hard to determine the changes in TechTree.lua and other files that were created instead of moved.

You mean separate commits for move and change? Git automatically determines whether a file is moved or re-created based on how different it is.

But then the entire file is rewritten (except TechTree due to time constraints) so there's not much point in comparing them.

azum4roll:
> Be careful with declaring classes, if you're trying to make something local try using `@package` or `@private` annotation to have limited scope. I'm looking at `Info` and wondering where it comes from..

Those are only available to functions according to the docs. Fields seem to be allowed to be private/package too, but not classes.

(1) Community Patch/Core Files/VPUI/VPUI_core.lua
```diff
@@ -193,20 +199,34 @@
 		return #tGameInfoCache[strTable];
 	end
 
+	--- IconHookup, but set the control texture to the random civ icon if it fails (instead of retaining the previous texture)
+	--- @param iPortraitIndex integer
+	--- @param iIconSize integer
+	--- @param strAtlas string
+	--- @param imageControl Control
+	local function IconHookupOrDefault(iPortraitIndex, iIconSize, strAtlas, imageControl)
+		if not IconHookup(iPortraitIndex, iIconSize, strAtlas, imageControl) then
```
vaderkos, line 209: Does IconHookup still exist after opening leaders screen bug?
When I last checked global scope was completely empty

axatin: a local copy of it is created at the beginning of the file

(1) Community Patch/Core Files/Overrides/Includes/IconSupport.lua
```diff
@@ -0,0 +1,256 @@
...
+				highlightControl:SetTextureOffsetVal(0, 0);
+			end
+
+			if (alwaysUseComposite or thisCivType == GameInfo.Civilizations["CIVILIZATION_MINOR"].ID) then
+				highlightControl:SetHide(false);
+			end
+		end
+		if shadowIconControl then
+			if (shadowed) then
+				shadowIconControl:SetHide(false);
```
vaderkos: `:SetHide(not shadowed)`?

(1) Community Patch/Core Files/Overrides/Includes/IconSupport.lua
```diff
@@ -0,0 +1,256 @@
+include("FLuaVector");
+
+local g_IconAtlases;
+
+local function PopulateIconAtlases()
+	g_IconAtlases = {};
+
+	for row in GameInfo.IconTextureAtlases() do
```
vaderkos: This would fill it with all data, even if it's not needed, why not fetch data on demand and allow it to be collected when not used. IMHO better use weak memoization

(1) Community Patch/Core Files/Overrides/Includes/IconSupport.lua
```diff
@@ -0,0 +1,256 @@
...
+		thisPlayerColorIndex = PreGame.GetCivilizationColor(playerID);
+		defaultColorSet = GameInfo.PlayerColors[thisCiv.DefaultPlayerColor];
+		nonDefaultColor = (defaultColorSet.ID ~= thisPlayerColorIndex);
+		if playerID >= GameDefines.MAX_MAJOR_CIVS then
+			nonDefaultColor = true;
+		end
+	end
+
+	if thisCiv and (alwaysUseComposite or nonDefaultColor) then
+		if (iconSize == 80) then
```
vaderkos: Use table lookup instead, it's faster in this particular case.

(1) Community Patch/Core Files/Overrides/Includes/IconSupport.lua
```diff
@@ -0,0 +1,256 @@
...
+			if (thisCivType == GameInfo.Civilizations["CIVILIZATION_MINOR"].ID) then
+				iconControl:SetColor(secondaryColorVector);
+				teamColorControl:SetColor(primaryColorVector);
+			else
+				teamColorControl:SetColor(secondaryColorVector);
+			end
+			teamColorControl:SetHide(false);
+		end
+		if highlightControl ~= nil then
+			if (iconSize == teamSize80) then
```
vaderkos: The only difference is offset x, use table lookup and do `highlightControl:SetTexture("CivIconBGSizes_Highlight.dds")` only once

(1) Community Patch/Core Files/Overrides/Includes/IconSupport.lua
```diff
@@ -0,0 +1,256 @@
...
+				highlightControl:SetTextureOffsetVal(77, 0);
+			elseif (iconSize == teamSize45) then
+				highlightControl:SetTexture("CivIconBGSizes_Highlight.dds");
+				highlightControl:SetTextureOffsetVal(32, 0);
+			elseif (iconSize == teamSize32) then
+				highlightControl:SetTexture("CivIconBGSizes_Highlight.dds");
+				highlightControl:SetTextureOffsetVal(0, 0);
+			end
+
+			if (alwaysUseComposite or thisCivType == GameInfo.Civilizations["CIVILIZATION_MINOR"].ID) then
```
vaderkos: Is it ok to fetch `GameInfo.Civilizations["CIVILIZATION_MINOR"]` every time?

(1) Community Patch/Core Files/Overrides/Includes/IconSupport.lua
```diff
@@ -0,0 +1,256 @@
...
+				shadowIconControl:SetHide(false);
+			else
+				shadowIconControl:SetHide(true);
+			end
+		end
+	else
+		if teamColorControl then
+			if thisCiv then
+				IconHookup(thisCiv.PortraitIndex, iconSize, thisCiv.IconAtlas, teamColorControl);
+				local white = {x = 1, y = 1, z = 1, w = 1};
```
vaderkos: Color can be shared and declared at the top.

(1) Community Patch/Core Files/Overrides/Includes/IconSupport.lua
```diff
@@ -0,0 +1,256 @@
...
+		local primaryColor         = GameInfo.Colors[thisPlayerColorSet.PrimaryColor];
+		local primaryColorVector   = Vector4(primaryColor.Red, primaryColor.Green, primaryColor.Blue, primaryColor.Alpha);
+		local secondaryColor       = GameInfo.Colors[thisPlayerColorSet.SecondaryColor];
+		local secondaryColorVector = Vector4(secondaryColor.Red, secondaryColor.Green, secondaryColor.Blue,
+			secondaryColor.Alpha);
+
+		iconControl:SetColor(primaryColorVector);
+
+		if teamColorControl then
+			if (iconSize == teamSize80) then
```
vaderkos: The only difference is offset x, use table lookup and do teamColorControl:SetTexture("CivIconBGSizes.dds") only once

(1) Community Patch/Core Files/Overrides/Includes/IconSupport.lua
```diff
@@ -0,0 +1,256 @@
...
+			end
+		end
+	end
+end
+
+---------------------------------------------------------------------------------------------------------
+-- This is a special case hookup for civilization icons that always uses the one-piece version
+function SimpleCivIconHookup(playerID, iconSize, iconControl)
+	local thisCivType = PreGame.GetCivilization(playerID);
+	local thisCiv = GameInfo.Civilizations[thisCivType];
```
vaderkos: There is no need to fetch whole rows, query for PortraitIndex and IconAtlas

vaderkos (commented):
I think IconSupport is used in so many places that we could gain a lot optimizing it

azum4roll:
> I think IconSupport is used in so many places that we could gain a lot optimizing it

This is WIP.

The global thing is likely more easily fixed by adding
```lua
if not PopulateIconAtlases then
	include("IconSupport");
end
```
in VPUI_core.lua.

I'm going to take the easier path and save the rewrite for later. This PR needs to be merged ASAP to allow me to work on text. Maybe you can do it?

EDIT: If you do, we won't even need IconHookupOrDefault anymore. Just change IconHookup to do that. Using the ? icon is much easier to spot mistakes than retaining the previous icon.

PR #11923, opened 2025-09-17
fix Scavenger promotion

azum4roll:


axatin:
We can now remove GainsXPFromPillaging entirely.

Civilopedia.lua also needs to be updated to analyze XPFromPillaging instead of GainsXPFromPillaging.

PR #11938, opened 2025-09-19
Update NewCivilizationText.xml

Hokath:
add missing examination hall keys


Hokath:
pulled to wrong place, ignore this

PR #11941, opened 2025-09-20
Fix debug log spam

axatin:
Related to this: @azum4roll  Can you fix non-EUI and BNW UnitPanel.lua calling CanHandleAction with 0 instead of an instance?

const int iAction = lua_tointeger(L, 1);
CvPlot* pkPlot = CvLuaPlot::GetInstance(L, 2, false);
const bool bTestVisible = luaL_optbool(L, 3, 0);
const bool bResult = GetInstance()->canHandleAction(iAction, pkPlot, bTestVisible);

if(action.Visible and Game.CanHandleAction( iAction, 0, 1 ) ) then

### DIFF
--- a/CvGameCoreDLL_Expansion2/Lua/CvLuaScopedInstance.h
+++ b/CvGameCoreDLL_Expansion2/Lua/CvLuaScopedInstance.h
@@ -166,10 +166,10 @@ InstanceType* CvLuaScopedInstance<Derived, InstanceType>::GetInstance(lua_State*
 #endif
 	}
 #if defined(VPDEBUG)
-	else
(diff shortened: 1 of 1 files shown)


azum4roll:
Currently it functions as if NULL is passed into CvGame::canHandleAction(). Is the desired behaviour different?

axatin:
I think we should explicitly use `nil` for 'no instance' and generate a warning if anything other than `nil` or a table is used. We might miss other bugs in lua function calls if we allow mixing numbers and instances.

azum4roll:
There's no rush. It can wait until we move UnitPanel.lua to (1).

PR #11945, opened 2025-09-21
Fix incorrect operator used

vaderkos:


axatin:
Already fixed in 8901ff8

PR #11947, opened 2025-09-21
Fix globals being nil in VPUI_core.lua

vaderkos:


axatin:
Merged in [e59428d](https://github.com/LoneGazebo/Community-Patch-DLL/commit/e59428da5677834b663b8c173dc98558db0cf9d3)

PR #11949, opened 2025-09-22
Luacheck

JohnsterID:
Improve .luacheckrc and LuaCATS


.luacheckrc
```diff
@@ -1,13 +1,117 @@
...
+	'611', --[[ Line contains only whitespace. ]]
+	'621', --[[ Inconsistent indentation (SPACE followed by TAB). ]]
+	'212', --[[ Unused argument (common in event callbacks) ]]
+	'213', --[[ Unused loop variable ]]
+	'311', --[[ Value assigned to variable is unused (common in UI code) ]]
+	'314', --[[ Value of field is unused (common in data structures) ]]
+	'542', --[[ Empty if branch (common in conditional compilation) ]]
+	'581', --[[ Negation of a relational operator can be simplified ]]
+	'211', --[[ Unused local variable (common in event handlers) ]]
+	'231', --[[ Variable is set but never accessed (common in data setup) ]]
```
vaderkos: Are we sure that all these have to be ignored and not fixed?

azum4roll: Empty if/else branch with a comment should be acceptable.

JohnsterID: Thanks for jumping on. It all depends how we want do it. I was thinking, ignore majority at the start then expand from there?

For review, the check is below on this page and current run direct link: https://github.com/LoneGazebo/Community-Patch-DLL/actions/runs/17900825305/job/50893647168?pr=11949

I could comment out all to see how large it grows as well.

Ideally, we can have Github annotations for future PRs as well. https://github.com/lunarmodules/luacheck/issues/57
Or we manually implement it, same as my other PR for the build CI, https://github.com/LoneGazebo/Community-Patch-DLL/pull/11944, example annotations: https://github.com/LoneGazebo/Community-Patch-DLL/actions/runs/17899999562?pr=11944

vaderkos: Ignoring majority won't motivate enough for fixing things. Without warning noone notices something is not correct. 
As for stylistic changes, I could make a PR that runs auto-formatting in every lua file in repo. What do you think about it?

JohnsterID: Only work if we have annotations or if people checked the output, it would be too many and scare them away. Annotations would be preferred.

Commit may be too large? Do we match existing format as close as possible or latest Lua 5.1 convention? I wouldn't mind doing the same for clang-format for the C++.

Want me to comment out the ignores so we can see it?

azum4roll: I still have to manually fix formatting after doing Alt+Shift+F in VSCode. It's useful for getting rid of trailing spaces though.

We should avoid doing large formatting commits until there's a need to rewrite the entire file.

vaderkos: What needs manual fixing?

azum4roll: Space before the table in function calls like `ExampleFunc {tParameter}`.

Placement of comment before conditionals like
```lua
if foo then
	DoStuff();
	-- this comment is supposed to describe the next line
elseif bar then
	DoSomethingElse();
end
```

azum4roll: When I rewrite Lua files, I only do auto format at the very start, then everything else is manual.

vaderkos: I think that's the default styling `ExampleFunc { }`, but otherwise IMHO better explicit than implicit `ExampleFunc({ })`. I'd stick to what is proposed by the default proposed by luacheck. There is also one much better lua linter and parser called emmylua-luals by xuhuanzy. But it works very strictly

JohnsterID (commented):
removed

vaderkos:
> removed

Why removed :D

LuaCATS/enum/ButtonStates.d.lua
```diff
@@ -0,0 +1,10 @@
+--- @meta
+
+--- Binary-only enum (registered by engine executable, not the DLL)
+--- @enum ButtonStates
```
vaderkos: Shall we do `ButtonState` instead of ButtonState**s**. The same as with types?

LuaCATS/Global.d.lua
```diff
@@ -44,7 +155,137 @@ function GridToWorld(x, y) end
...
 --- that's available in every state except `Main State`
 --- even if EnableLuaDebugLibrary = 0 in config.ini
 traceback = debug.traceback
+
+---------------------------------------------------------------------------
+-- Engine utility globals
+---------------------------------------------------------------------------
+
+--- @type string[]
+IncludeFileList = {}
```
vaderkos: That might not be available anywhere except main state

vaderkos:
@JohnsterID could you give me the command to run checks locally? I'd try fixing some :D

LuaCATS/enum/FeatureTypes.d.lua
```diff
@@ -0,0 +1,32 @@
+--- @meta
+
+--- @enum FeatureType
```
azum4roll: This will collide with the FeatureType that's a pseudo integer referring to IDs of the Features table.

azum4roll: Same issue with other table ID types.

azum4roll:
Instead of hardcoding the functions from vanilla lua files in Global.d.lua, I currently copy the files into LuaCATS/external in my branch.

.luacheckrc
JohnsterID: Rewrote this. Now only style warnings are suppressed (whitespace W6xx, line length W631). All real code warnings are kept.

LuaCATS/enum/ButtonStates.d.lua
JohnsterID: Good catch, fixed. Singularized 6 enum type names:
ButtonStates→ButtonState, CoopWarStates→CoopWarState,
GameStates→GameState, NetErrors→NetError,
NetKicked→NetKickedReason, TradeableItems→TradeableItem.

LuaCATS/enum/FeatureTypes.d.lua
JohnsterID: Fixed. Changed 7 files from `@enum FooType` to `@type table<string, FooType>`. This avoids the collision with the `@alias FooType integer` in Pseudo.d.lua while keeping the table shape typed. Affected: FeatureTypes, PlotTypes, YieldTypes, TerrainTypes, BeliefTypes, ReligionTypes, AccomplishmentTypes.

JohnsterID:
> could you give me the command to run checks locally? I'd try fixing some :D

Download `luacheck.exe` from https://github.com/lunarmodules/luacheck/releases (no install needed), drop it in the repo root, and run:

```
luacheck.exe --config .luacheckrc .
```

LuaCATS/Global.d.lua
JohnsterID: `IncludeFileList` and `RefreshIncludeFileList` are set in Main State globals only, not in UI context environments. You already had it in CPK.lua as it already nil-checks them for this reason. Fixed: typed as `string[]|nil` and `fun()|nil` respectively.

JohnsterID:
> Instead of hardcoding the functions from vanilla lua files in Global.d.lua, I currently copy the files into LuaCATS/external in my branch.

That works well for complex includes. For FLuaVector/TruncateString the stubs are small and let us add richer type annotations than the original code has. Let me know what you guys prefer? Is there a way to override by using external instead of the hardcode?

vaderkos:
> > Instead of hardcoding the functions from vanilla lua files in Global.d.lua, I currently copy the files into LuaCATS/external in my branch.
> 
> That works well for complex includes. For FLuaVector/TruncateString the stubs are small and let us add richer type annotations than the original code has. Let me know what you guys prefer? Is there a way to override by using external instead of the hardcode?

@JohnsterID I'd rather go with a per-path/file configuration of globals. I'm not so familiar with which files are available where. But in that way we can have most strict definitions.

The other way around I'm thinking right now that might work is... Generating custom include definitions, I'm investigating if this is possible. 

(Update) Looks like without modifying luals itself it's not possible... :(

PR #11954, opened 2025-09-27
Two small fixes for TechTree.lua

schnetziomi5:
The tech tree would not properly display [the icons for unlocked stuff]

- the call to `AddSmallButtonsToTechButton` was missing an argument

but still with that fixed the icons for buildings, improvements and units were missing...

- `GatherInfoAboutUniqueStuff( g_activeCivType )` in `TechButtonInclude.lua` needs to be called somewhere to fill the `ValidXXXBuilds` tables - I restored the call to the location it was before....

### DIFF
--- a/(3a) EUI Compatibility Files/ImprovedTechTree/TechTree.lua
+++ b/(3a) EUI Compatibility Files/ImprovedTechTree/TechTree.lua
@@ -262,7 +262,7 @@ local function RefreshDisplayOfSpecificTech( tech )
 	--print("RefreshDisplayOfSpecificTech afw Stuff done")
 	-- Rebuild the small buttons if needed
 	if g_NeedsFullRefresh then
-		AddSmallButtonsToTechButton( thisTechButton, tech, g_maxSmallButtons, 45, g_activePlayerID)
+		AddSmallButtonsToTechButton( thisTechButton, tech, g_maxSmallButtons, 45, 1, g_activePlayerID)
 	end
 	if g_EspionageViewMode then
 		thisTechButton.TechButton:SetDisabled(true);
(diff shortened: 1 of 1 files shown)


schnetziomi5:
Also my city view was rather empty, so I investigated the faulty [code]( https://github.com/LoneGazebo/Community-Patch-DLL/blob/78666310173d67dbec8ffe1723f1b7936bbe073f/(3a)%20EUI%20Compatibility%20Files/LUA/CityView.lua#L931
)

... turns out `YieldTypes.NUM_YIELD_TYPES`  apparently equates to `17` while `Game.GetNumYieldTypes()` or it's C++ counterpart `GC.getNUM_YIELD_TYPES()` only returns `12`

While most other function seem to silently return `0` in similar cases, the implementation for `city:GetBuildingYieldRateTimes100(buildingID, yieldID)` is more strict and throws an error. I "fixed" it by changing the iterator to use `Game.GetNumYieldTypes()` but insofar as the problem is `YieldTypes.NUM_YIELD_TYPES` and `Game.GetNumYieldTypes()` disagreeing, that's not an actual solution....

azum4roll:
The TechTree part was due to me backing out of making EUI tech tree use non-EUI TechButtonInclude but not fully removing the changes. The fix is correct.

azum4roll:
> YieldTypes.NUM_YIELD_TYPES

In general, replace all usage of this with `Game.GetNumYieldTypes()`, OR `VP.GetNumInfos("Yields")`, OR change the iterator to `for yieldID in VP.GameInfoCache("Yields") do` where `VP = MapModData and MapModData.VP or VP`. But make sure you cache the value if you use the former (only call DLL once).

PR #11960, opened 2025-09-27
Vox Deorum: opening IPC channels into Civ 5

CIVITAS-John:
See the protocol here https://github.com/CIVITAS-John/vox-deorum/blob/main/bridge-service/PROTOCOL.md

The mod doesn't do anything for real if the option is not enabled. However, it does add tiny overhead to many places (check if the flag is enabled, basically) because I want it to process Lua calls sooner rather than later. Late game AI operations can take minutes on potato computers/servers.

### DIFF
--- a/(1) Community Patch/Database Changes/NewCustomModOptions.xml
+++ b/(1) Community Patch/Database Changes/NewCustomModOptions.xml
@@ -993,6 +993,9 @@
 		<Row Class="6" Name="BATTLE_ROYALE" Value="0"/>
 
 		<Row Class="0" Name="UNIT_KILL_STATS" Value="0"/>
+		
+		<!-- Enables support for Vox Deorum modmod (opening up the IPC channel) -->
+		<Row Class="3" Name="IPC_CHANNEL" Value="0"/>
 	</CustomModOptions>
 
 	<!-- Events for JFD Piety -->
--- a/(1) Community Patch/Database Changes/Text/en_US/UI/CoreNewUIText.xml
+++ b/(1) Community Patch/Database Changes/Text/en_US/UI/CoreNewUIText.xml
@@ -2752,49 +2752,49 @@
 		</Row>
 		<!-- Debug messages -->
 		<Row Tag="TXT_KEY_DIPLO_REAL_APPROACH_WAR">
-			<Text>Real Approach is War.</Text>
+			<Text>Our real approach is War.</Text>
 		</Row>
 		<Row Tag="TXT_KEY_DIPLO_REAL_APPROACH_PLANNING_WAR">
-			<Text>Real Approach is Planning War.</Text>
+			<Text>Our real approach is Planning War.</Text>
 		</Row>
 		<Row Tag="TXT_KEY_DIPLO_REAL_APPROACH_HOSTILE">
-			<Text>Real Approach is Hostile.</Text>
+			<Text>Our real approach is Hostile.</Text>
 		</Row>
 		<Row Tag="TXT_KEY_DIPLO_REAL_APPROACH_DECEPTIVE">
-			<Text>Real Approach is Deceptive.</Text>
+			<Text>Our real approach is Deceptive.</Text>
 		</Row>
 		<Row Tag="TXT_KEY_DIPLO_REAL_APPROACH_GUARDED">
-			<Text>Real Approach is Guarded.</Text>
+			<Text>Our real approach is Guarded.</Text>
 		</Row>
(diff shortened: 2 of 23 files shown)


RecursiveVision:
@axatin Please wait before merging, I want to review this first.

CIVITAS-John:
@RecursiveVision I am also doing a round of refactoring to improve the event forward performance - will get that out tonight. That said, the change is limited to `ForwardGameEvent` and other things are unchanged. Thanks again

CIVITAS-John:
Refactor done, allowing the IPC endpoint to register a Lua function to post-process outgoing game events. This would avoid large numbers of Lua calls back and forth (postprocessing required a Lua function call to add metadata for LLM consumption, e.g., the name of the unit class, etc) and greatly improve the performance especially late game.

RecursiveVision:
@CIVITAS-John
You should change CvLuaPlayer::SetGrandStrategy() to change the Diplomacy AI's *current* victory pursuit to have the intended effect.

It's worth noting that each turn if victory competition is disabled or certain other conditions are met (see `IsSeriousAboutVictory()`, the current victory pursuit will be updated to the primary victory pursuit.

I'd advise using a different function to retrieve/set the Primary and Secondary Victory Pursuits. It's worth noting that these two values may as well be flavors in and of themselves.

CIVITAS-John:
Hi! Thanks for pointing that out. I have checked the DoUpdateCurrentVictoryPursuit(). It seems that if I only override CurrentVictoryPursuit, it is only going to be overridden there. On the other hand, once we change both the grand strategy AND the primary pursuits, it will guarantee the current victory pursuit *unless* the player is close to the victory. Am I missing something here?

Either way, it sounds good to separate the two functions (GrandStrategy/VictoryPursuits). I will revisit this once the main request gets merged (that commit has been somehow merged; I am pretty awkward with Git in general)

RecursiveVision:
> It's worth noting that each turn if victory competition is disabled or certain other conditions are met (see `IsSeriousAboutVictory()`, the current victory pursuit will be updated to the primary victory pursuit.

See my edit to the above comment for what you're probably missing. :)

CIVITAS-John:
Cool! So I will split into SetGrandStrategy(), SetPrimaryVictoryPursuit(), and SetSecondaryVictoryPursuit() in a later commit. SetCurrentVictoryPursuit doesn't make sense since it is only getting overriden soon.

RecursiveVision:
On second thought, if your goal is for it to persist between turns and not get updated each turn by that check, I can add an option for that.

CIVITAS-John:
Hmmmm, I think the goal is for the LLM to set both the grand strategy and the corresponding victory pursuit so they get aligned with each other (I cannot see why any human would make the two different, though I can see why our current AI does that)

RecursiveVision:
Hmm, my original comment there was assuming the LLM would be updating the values every turn, which might be an incorrect assumption.

If that's the case then your existing code should work, but I'd just change it to also update the current victory pursuit so that the change happens instantly.

PrimaryVictoryPursuit and SecondaryVictoryPursuit do have additional uses in changing the values of various diplo modifiers so there may be a reason not to align them.

CIVITAS-John:
Make sense! I will do that. 

Also, I wonder which is the easiest way for LLMs to add/remove additional diplomatic modifiers? e.g. imagine you chat with a LLM and it can raise or lower the relationship based on the conversation :)

RecursiveVision:
Check out `MOD_EVENTS_DIPLO_MODIFIERS`.

CIVITAS-John:
Interesting. Seems that I need to add an item to DiploModifier and then a custom Lua hook to return a dynamic value (set by the LLM and possibly saved in Modding:OpenSaveData(). Or maybe just use ScenarioModifier.

However, I am not sure how much performance impact this would have vs. create a new modifier and get it serialized in the save itself?

RecursiveVision:
`However, I am not sure how much performance impact this would have vs. create a new modifier and get it serialized in the save itself?`

I am not sure, though we avoid gating serialization behind CustomModOptions.

Also, there is no point in merging this yet because the beta version of 5.0 will be released as a modpack.

CIVITAS-John:
I think if I am to do that, the serialization should be for all and the same do the APIs. It shouldn't add too much storage space anyway. 

What do you mean by released as a modpack? Do we have a way to automate the modpack building? That sounds really cool.

RecursiveVision:
Author has said they will submit a revised PR after the beta release of 5.0.

PR #11968, opened 2025-09-28
Various fixes/refactorings

KungCheops:
Move function only used in CvBuilderTaskingAI to CvBuilderTaskingAI (from CvPlot).

Don't apply negative modifier for repairing Great Person improvements on top of existing/planned roads.

Allow AI players to replace Great Person improvements and Landmarks.

Change defensive improvement downprioritization in safe territory to only apply to forts.

Fix several possible bugs in HomelandAI when evaluating improvements on the same turn as another improvement was finished.

Fix minor bug with improvement planning regarding forests for vanilla Celts.

Limit maximum scouting units for AIs to 1/4 of total supply.

Fix theming bonus being set to 0 in almost all cases.

Fix great work tooltip.

Fix emergency units bought by military AI not getting promoted on the same turn as they are bought.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvBuilderTaskingAI.cpp
+++ b/CvGameCoreDLL_Expansion2/CvBuilderTaskingAI.cpp
@@ -1313,7 +1313,8 @@ bool CvBuilderTaskingAI::ExecuteWorkerMove(CvUnit* pUnit, BuilderDirective aDire
 			UNREACHABLE();
 		case BuilderDirective::BUILD_IMPROVEMENT_ON_RESOURCE:
 		case BuilderDirective::BUILD_IMPROVEMENT:
-		case BuilderDirective::REPAIR:
+		case BuilderDirective::REPAIR_IMPROVEMENT:
+		case BuilderDirective::REPAIR_ROUTE:
 		case BuilderDirective::BUILD_ROUTE:
 		case BuilderDirective::REMOVE_FEATURE:
 		case BuilderDirective::REMOVE_ROAD:
@@ -1355,8 +1356,11 @@ bool CvBuilderTaskingAI::ExecuteWorkerMove(CvUnit* pUnit, BuilderDirective aDire
 				case BuilderDirective::BUILD_IMPROVEMENT:
 					strLog += "On plot,";
 					break;
-				case BuilderDirective::REPAIR:
+				case BuilderDirective::REPAIR_IMPROVEMENT:
 					strLog += "Repairing,";
+					break;
+				case BuilderDirective::REPAIR_ROUTE:
+					strLog += "Repairing route,";
 					break;
 				case BuilderDirective::BUILD_ROUTE:
 					strLog += "Building route,";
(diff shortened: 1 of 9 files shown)


azum4roll:
> Limit maximum scouting units for AIs to 4.

Why set a cap to it instead of letting it continue to be based on map size/unexplored plots?

Can you look into switching unit AI between UNITAI_EXPLORE/UNITAI_EXPLORE_SEA and others defined in Unit_AITypes for the unit? Currently AI switching is only one way into UNITAI_EXPLORE/UNITAI_EXPLORE_SEA.

Ideally they should switch to the best AI available when there are more than 4 recon units, before considering disbanding. EDIT: looks like GetExplorersNeeded() isn't even considered for disbanding.

KungCheops:
It's not very well implemented, was playing a game as Polynesia on standard size and it wanted to build 7 scouts early game (when I had 12 supply). I think 4 should be more than enough until I figure out a better way for it to calculate this.

KungCheops:
It's basically based on how large your "frontier" is, which I don't think is a particularly good metric. For Polynesia in particular, the frontier gets insanely large when you send all your scouts in different directions and you're not limited by deep ocean.

azum4roll:
> Can you look into switching unit AI between UNITAI_EXPLORE/UNITAI_EXPLORE_SEA and others defined in Unit_AITypes for the unit? Currently AI switching is only one way into UNITAI_EXPLORE/UNITAI_EXPLORE_SEA.
> 
> Ideally they should switch to the best AI available when there are more than 4 recon units, before considering disbanding. EDIT: looks like GetExplorersNeeded() isn't even considered for disbanding.

If you can do this, we can add secondary AI to recon units, and Conquistadors and their upgrades will work for the AI again.

KungCheops:
> > Can you look into switching unit AI between UNITAI_EXPLORE/UNITAI_EXPLORE_SEA and others defined in Unit_AITypes for the unit? Currently AI switching is only one way into UNITAI_EXPLORE/UNITAI_EXPLORE_SEA.
> > Ideally they should switch to the best AI available when there are more than 4 recon units, before considering disbanding. EDIT: looks like GetExplorersNeeded() isn't even considered for disbanding.
> 
> If you can do this, we can add secondary AI to recon units, and Conquistadors and their upgrades will work for the AI again.

I have not looked into that part of the code at all and I wasn't planning to tbh.

KungCheops:
This fix isn't to prevent Spain from building too many conquistadors, they already won't. This is to fix Polynesia building too many scouts (or all civs on any map that is fully explorable from turn 1, like any land based map). The function for calculating needed scouts is just bad.

azum4roll:
Spain SHOULD spam Conquistadors and still use the upgraded units to fight.

KungCheops:
> Spain SHOULD spam Conquistadors and still use the upgraded units to fight.

That has nothing to do with this commit though. So we can drop it and attend to it when we decide to fix AI conquistador use.

KungCheops:
Change it to cap at 1/4 of total supply limit instead, should be fine in all cases.

PR #11970, opened 2025-09-28
fix assert/precondition related problems

azum4roll:


axatin:
We should also set bHideAssert to true when checking the prisoners of war promotion, or we'll get an assert hit in Community Patch

axatin:
I'll add it in my branch to reduce the number of commits.

Issue #11876, opened 2025-08-10
Setting Vassal Taxes not supported in Multiplayer


axatin:
In `CvTeam::DoApplyVassalTax` a network message needs to be sent. And maybe in other vassal-related functions as well (DoLiberateVassal?).

axatin:
Should be fixed now

PR #11972, opened 2025-09-28
Squads improvements

artemigkh:
* During group movement, non-combat units will now be escorted by combat units on the same starting tile, linking their movement until the destination is reached or another move order is issued. This behavior is configurable in the squads options menu and enabled by default.
* Group movement destination tiles now have an accurate preview instead of a generic 2 tile radius ring.
* Improved group movement destination tile selection logic. Units now select tiles that have a better grouping around the selected destination.
* Fixed an issue where embarked units were not routed to the same tiles as naval units when available and such tiles would be closer to the target destination.
* Fixed an issue where non-combat units were sometimes routed to the destination tile while no combat units in the squad were.
* Fixed an issue where box select did not work correctly around the map world wrap.

### DIFF
--- a/(4a) Squads for VP/UI/Squads.lua
+++ b/(4a) Squads for VP/UI/Squads.lua
@@ -18,6 +18,7 @@ include( "SquadNames.lua" );
 
 
 local bHighlightSquadUnits = true;
+local bEscortNonCombatUnits = true;
 local SquadsEndMovementType = 0;
 
 SQUADS_MODE_NONE = 0;
@@ -30,11 +31,15 @@ local currentMode = SQUADS_MODE_NONE;
 
 local bIsBoxSelecting = false;
 local pSelectBoxStartPlot = nil;
+local bBoxSelectingLeftToRight = nil;
+local mapW, mapH = Map.GetGridSize();
 
 -- Options Logic
 function SquadsOptionChanged(optionKey, newValue)
     if optionKey == "HighlightSquadUnits" then
         bHighlightSquadUnits = newValue;
+    elseif optionKey == "EscortNonCombatUnits" then
+        bEscortNonCombatUnits = newValue;
     elseif optionKey == "SquadsEndMovementType" then
         SquadsEndMovementType = newValue;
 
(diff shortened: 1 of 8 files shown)


Hokath:
Accurate preview sounds awesome!

(4a) Squads for VP/UI/Squads.lua
```diff
@@ -118,14 +123,29 @@ function UpdateMouse(gridX, gridY)
 
 
             if bIsBoxSelecting then
-                --print(string.format("Start plot for box select: (%i,%i)", pSelectBoxStartPlot:GetX(), pSelectBoxStartPlot:GetY()))
-                --print(string.format("End plot for box select: (%i,%i)", currentPlot:GetX(), currentPlot:GetY()))
+                -- Determine box select direction across world wrap using mapW
+                if bBoxSelectingLeftToRight == nil and currentPlot:GetX() ~= pSelectBoxStartPlot:GetX() then
+                    if currentPlot:GetX() > pSelectBoxStartPlot:GetX() then
+                        if currentPlot:GetX() - pSelectBoxStartPlot:GetX() > mapW / 2 then
```
vaderkos: `bBoxSelectingLeftToRight = currentPlot:GetX() - pSelectBoxStartPlot:GetX() < mapW / 2` is enough

(4a) Squads for VP/UI/Squads.lua
```diff
@@ -138,11 +158,14 @@ function UpdateMouse(gridX, gridY)
 
         HighlightSquadUnits(pPlayer, currentSquadNumber);
         Events.ClearHexHighlightStyle("SquadsGroupMovement");
-        -- Draw the movement preview shadow (2 rings worth)
-        HighlightPlot(currentPlot, nil, "SquadsGroupMovement");
-        for pLoopPlot in PlotAreaSpiralIterator(currentPlot, 2) do
-            -- TODO: only highlight tiles current squad units can end on
-            HighlightPlot(pLoopPlot, nil, "SquadsGroupMovement");
+        -- Draw the movement preview region
+        for unit in getSquadUnits(pPlayer, currentSquadNumber) do
```
vaderkos, line 154: This loop is not needed?
```lua
local unit = getSquadUnits(pPlayer, currentSquadNumber)()
-- logic below
```

artemigkh: The loop is needed to access the first element of the iterator, as lua has nothing equivalent to pythons `next()`, nor is it able to index iterators. Looping over it then breaking after the first element is the best (and only) solution I could find on the internet, though if you know of a cleaner way to do it let me know. 

Your current suggestion doesn't work, errors with
```
Squads.lua:162: Not a valid instance.  Either the instance is NULL or you used '.' instead of ':'.
```

(4a) Squads for VP/UI/Squads.lua
artemigkh: good call,
```
if currentPlot:GetX() > pSelectBoxStartPlot:GetX() then
    bBoxSelectingLeftToRight = currentPlot:GetX() - pSelectBoxStartPlot:GetX() < mapW / 2
else
    bBoxSelectingLeftToRight = pSelectBoxStartPlot:GetX() - currentPlot:GetX() > mapW / 2
end
```
 is much cleaner, I've made this change

vaderkos:
Is there any possibility to make it work in multiplayer ?

artemigkh:
> Is there any possibility to make it work in multiplayer ?

@vaderkos Outside of two issues, I don't see why it wouldn't already work in multiplayer, as at its core all the Squads modmod does really is issue movement commands via `gDLL->sendPushMission(pLoopUnit->GetID(), CvTypes::getMISSION_MOVE_TO() ...` which I believe should be handled correctly for MP. It also uses stable sort where sorting is required so I think it shouldn't cause desyncs?

The two problems are:
1) The saving of squad IDs isn't hooked up to any net message code, so after a game is reloaded everyone but the host will lose their squad assignments
2) (hotseat only) The lua code for squads assumes only one human player, so you get an issue with enemy units in your squad, lol. (Though this one is fixable fairly easily)

axatin:
> 1. The saving of squad IDs isn't hooked up to any net message code, so after a game is reloaded everyone but the host will lose their squad assignments

I can fix that one. There are also some other parts of the code where network messages need to be added (https://github.com/LoneGazebo/Community-Patch-DLL/issues/11876). Which function needs to be changed?

PR #11973, opened 2025-09-29
Add mod options for letting AI control certain parts of civ management

KungCheops:
Add mod options for letting AI control certain parts of civ management

Adds the following mod options that provide the option to let the AI control the corresponding part of civilization management:

MOD_AI_CONTROL_CITY_MANAGEMENT
MOD_AI_CONTROL_CITY_PRODUCTION
MOD_AI_CONTROL_DIPLOMACY
MOD_AI_CONTROL_ECONOMY
MOD_AI_CONTROL_ESPIONAGE
MOD_AI_CONTROL_EVENT_CHOICE
MOD_AI_CONTROL_FAITH_SPENDING
MOD_AI_CONTROL_GREAT_PERSON_CHOICE
MOD_AI_CONTROL_POLICY_CHOICE
MOD_AI_CONTROL_RELIGION_CHOICE
MOD_AI_CONTROL_TECH_CHOICE
MOD_AI_CONTROL_TOURISM
MOD_AI_CONTROL_UNIT_PROMOTIONS
MOD_AI_CONTROL_UNITS
MOD_AI_CONTROL_WORLD_CONGRESS

Some combinations of mod options may not work perfectly since the AI systems are interconnected in some places. A lot of it is compartmentalized though, and works quite well without the others.

To achieve this, the isHuman functions in CvPlayer, CvTeam, CvUnit and CvCity have been modified to take an enum argument that corresponds to one of the mod options. Since most AI functionality is controlled by an isHuman check, this works quite well. For the few places where this isn't the case, the mod options are used directly. The number of places where this is the case has been kept to an absolute minimum to avoid introducing bugs.

There are some enums that don't have a corresponding mod option, these were added purely for documentation purposes. E.g. for logging or achievement checks.


CIVITAS-John:
Wow! Is it possible to make it changable during a game run? Like if I don't want to control all army movement each turn late game but I surely want to micro my first scout

KungCheops:
> Wow! Is it possible to make it changable during a game run? Like if I don't want to control all army movement each turn late game but I surely want to micro my first scout

You can change the mod options mid game, but you have to reload your save. I think it should work though.

KungCheops:
Haven't tested all combinations yet, not sure how the military AI works without the Diplo/economic AI to set up goals and some values. Some things are done by the so-called "shadow AI" that always runs for the human player but not sure exactly how much that entails.

KungCheops:
I think this should actually be safe to merge now, all the mod options do nothing when set to false. Tried to clean it up so all the mod options are applied in a controlled manner in the same place.

RecursiveVision:
A couple of the ISHUMAN reasons are incorrect. Will point out which ones I spotted later.

It does seem like an intrusive change to make every isHuman() call have an extra parameter. Not sure I like this.

Does the UI work properly?

KungCheops:
> A couple of the ISHUMAN reasons are incorrect. Will point out which ones I spotted later.
> 
> It does seem like an intrusive change to make every isHuman() call have an extra parameter. Not sure I like this.
> 
> Does the UI work properly?

I've gone through all the isHuman checks two times, but I can see how some could still be wrong. The UI works.

KungCheops:
It's either being intrusive or being extremely unsafe and unmaintainable, I tried the other way first and it got very messy.

RecursiveVision:
> It's either being intrusive or being extremely unsafe and unmaintainable, I tried the other way first and it got very messy.

It's not a critique of your implementation, more a critique of 'are people actually going to use this for anything?'

It's cool, but is it practical enough to justify the intrusion?

KungCheops:
> > It's either being intrusive or being extremely unsafe and unmaintainable, I tried the other way first and it got very messy.
> 
> It's not a critique of your implementation, more a critique of 'are people actually going to use this for anything?'
> 
> It's cool, but is it practical enough to justify the intrusion?

Yeah I'm not sure, it's been really nice for me personally, but I have fairly unique interests. I was thinking it could possibly be used for challenge runs, e.g. can I win without controlling my units, etc.

The AI systems are ofc fairly interdependent so it's not perfect. I was thinking about some possible additions like the strategic AI sending notifications to the player telling them to attack certain cities (if they are in control of the units). Or (even more difficult to implement) to make UI elements for the human to set strategic goals for its armies.

KungCheops:
But that would be future considerations, if this were to be merged.

RecursiveVision:
Aside from the intrusion on `isHuman()` my concern would be, do you plan on actively maintaining this? It's a lot of interconnected code and potential to introduce bugs.

KungCheops:
> Aside from the intrusion on `isHuman()` my concern would be, do you plan on actively maintaining this? It's a lot of interconnected code and potential to introduce bugs.

I've tried to design it in such a way that when the mod options are disabled, nothing can happen. I would suggest to anyone wanting to try them out that they should expect some buggy behavior when some/all of them are enabled. If people find them fun to play with and they notice inconsistencies or bugs I'd be happy to try and solve them.

I don't think there are more than maybe a handful of places where the mod options are used outside of isHuman, and isHuman is designed to work like before when the mod options are disabled. So in my opinion I believe the changes to be safe. But I'm open to be convinced otherwise.

If there are bugs introduced even when the mod options are disabled, I'll commit to fixing those for sure.

I've been playing with these mod options enabled quite a bit now and will continue to do so if I get it merged (it would be my preferred way of "playing" or for e.g. debugging the AI). I will probably find some way to play this way even if it's not merged but it will be a fair bit more annoying. :)

If anyone wants to add new isHuman calls in the future, the OTHER_ISHUMAN_REASON argument is always safe to use, although I suppose people would know the specific reason for adding the checks.

I have some minor changes on a local branch that I'll force push soon-ish.

RecursiveVision:
Well I'm currently trying to reorganize, fix, and prune CustomModOptions that have stopped working properly. Even if the bugs are only when it's enabled I'm still not fond of that. :P

KungCheops:
I can't promise I'll be around forever but I'm not planning on letting this rot, especially since I'll want to use all mod options myself whenever I play in the future.

KungCheops:
Cleaning up some embarrassing mistakes

axatin:
If you want to make it less intrusive, you can just make OTHER_ISHUMAN_REASON the default value in `bool isHuman(IsHumanReason eIsHumanReason) const;`

It's better for performance if the pregame::isHuman check is done first, the switch-statement can be skipped if `pregame::isHuman()` is false (which is will be most of the time).

KungCheops:
> If you want to make it less intrusive, you can just make OTHER_ISHUMAN_REASON the default value in `bool isHuman(IsHumanReason eIsHumanReason) const;`
> 
> It's better for performance if the pregame::isHuman check is done first, the switch-statement can be skipped if `pregame::isHuman()` is false (which is will be most of the time).

Sounds reasonable, getting on it!

KungCheops:
Maybe I should remove the ones that don't have a corresponding mod option, don't know how much value they add.

PR #11980, opened 2025-10-04
AI fixes for exploration, city production and more

KungCheops:
#### QoL change:

Units that can pop ancient ruins will now cancel their current move order if an ancient ruin is revealed during it.

#### AI strategic planning:

AI will now use recon units in combat when they aren't needed for exploration.

AI will now properly use non-recon units to explore if they have too few recon units. This will be particularly noticable early game, as the AI should now explore their initial surroundings much faster.

AI will now properly assign naval units as explorers when needed.

AI will now consider deep ocean plots as explorable as soon as it can build ocean-faring units, instead of when all units can cross oceans. This means that they will start assigning caravels to explore as soon as they are available in VP.

Fix typo causing AI to basically never adopt the "settle other continents" strategy.

#### AI Exploration:

Explorers now coordinate with each other so they don't all run to the same place.

Non-recon exploring units will no longer try to get ancient ruins when exploring if they can't pop them.

Explorers should now be much less prone to ignore nearby ancient ruins.

Exploring AI units will no longer think barbarian camps are undefended (and try to snipe them) when they are in fog of war.

Exploring AI units will no longer suicide into unseen enemies.

Exploring AI units will no longer randomly stop or do stupid moves when they run into previously unseen enemies. Instead they will choose a new exploration target.

AI will now send scouts home to be upgraded to explorers when they are available (in VP).

Naval explorers should now be less prone to getting stuck because they want to heal and think moving is dangerous. They will try to move towards safe harbor.

#### City production AI:

Change city production logic for building work boats. Will now not build work boats in cities which can't reach their own resource plots. Should now be better at coordinating with other cities, supporting them with work boats. Can now build two work boats in direct succession if that is needed.

AI cities should now actually be able to build work boats for other nearby cities, even if they don't need any work boats themselves.

AI should no longer build units in small landlocked oceans (unless they are contested).

Fix typo causing AI to not build air carriers when needed, and to build them when not needed.

#### Military/tactical/homeland AI:

AI units should now be much more willing to stay and heal instead of running around.

Fix AI sometimes sending non-recon units out to claim ancient ruins.

AI workers should now heal when not taking terrain damage.

AI no longer assigns slingers the same way as other archers and skirmishers in formations, this would cause them to occasionally stand two tiles away from enemy units and do nothing.

Teach AI how to use conquistadors to found cities.

#### Worker AI:

Force AI workers to never build on dig sites (if we can build landmarks).

Greatly increase AI worker evaluation of late game strategic resources.

#### Civ changes:

Revert VP conquistador to UNITAI_FAST_ATTACK.

#### Various fixes:

Fix some incorrect tech triggers for military strategies in VP.

Enable mod option for half move from terrain promotions.

Fix AI Spain starting in the renaissance assigning their starting explorer as a settler.


CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -9484,16 +9484,22 @@ bool CvPlayer::HasActiveSettler()
 	return false;
 }
 
-/// Returns number of Units a player has with a particular UnitAI.  The second argument allows you to check whether or not to include Units currently being trained in Cities.
-int CvPlayer::GetNumUnitsWithUnitAI(UnitAITypes eUnitAIType, bool bIncludeBeingTrained)
+/// Returns number of Units a player has with a particular UnitAI.
+// The second argument allows you to check whether or not to include Units currently being trained in Cities.
+// The third argument allowsy you to check for default AI type rather than what it is currently assigned to.
+int CvPlayer::GetNumUnitsWithUnitAI(UnitAITypes eUnitAIType, bool bIncludeBeingTrained, bool bDefaultAI)
```
RecursiveVision, line 9491: Typo: allowsy

RecursiveVision:
I think we can make the Unit Production AI tweaks perma-enabled. Who doesn't want AI improvements?

(2) Vox Populi/Database Changes/Civilizations/Spain.sql
```diff
@@ -83,7 +83,7 @@ SET
 		)
 	),
 	Combat = (SELECT Combat FROM Units WHERE Type = 'UNIT_EXPLORER') + 6,
-	DefaultUnitAI = 'UNITAI_EXPLORE', -- fixme: this should be changed to UNITAI_FAST_ATTACK once a solution is found for the AI spamming Conquistadors that upgrade into exploration units
+	DefaultUnitAI = 'UNITAI_FAST_ATTACK', -- fixme: this should be changed to UNITAI_FAST_ATTACK once a solution is found for the AI spamming Conquistadors that upgrade into exploration units
```
RecursiveVision: Can remove the comment now

CvGameCoreDLL_Expansion2/CvHomelandAI.cpp
```diff
@@ -2466,8 +2507,10 @@ bool CvHomelandAI::ExecuteExplorerMoves(CvUnit* pUnit)
 		if (!IsValidExplorerEndTurnPlot(pUnit, pEvalPlot))
 			continue;
 
+		bool bCanPopGoody = MOD_BALANCE_CORE_GOODY_RECON_ONLY ? pUnit->getUnitCombatType() == (UnitCombatTypes)GC.getInfoTypeForString("UNITCOMBAT_RECON", true) || pUnit->GetGainsXPFromScouting() : true;
+
```
RecursiveVision: Should move this code up to where iMoveFlags is initialized. No need to run it for every plot.

Also I would write this as
`!MOD_BALANCE_CORE_GOODY_RECON_ONLY || pUnit->getCombatType() == (UnitCombatTypes)GC.getInfoTypeForString("UNITCOMBAT_RECON", true) || pUnit->IsGainsXPFromScouting();`

KungCheops:
> I think we can make the Unit Production AI tweaks perma-enabled. Who doesn't want AI improvements?

I concur

CvGameCoreDLL_Expansion2/CvHomelandAI.cpp
KungCheops: sounds good, will fix when I'm home or tomorrow

CvGameCoreDLL_Expansion2/CvHomelandAI.cpp
```diff
@@ -2625,12 +2692,67 @@ bool CvHomelandAI::ExecuteExplorerMoves(CvUnit* pUnit)
...
-	pUnit->AI_setUnitAIType(pUnit->getUnitInfo().GetDefaultUnitAIType());
-	ExecuteMovesToSafestPlot(pUnit);
+	if (!pBestPlot)
+	{
+		// Take this unit off exploration duty since we couldn't find any target plot to explore
+		UnitAITypes eUnitAI = pUnit->getUnitInfo().GetDefaultUnitAIType();
+		eUnitAI = eUnitAI != UNITAI_EXPLORE ? eUnitAI : UNITAI_FAST_ATTACK;
+		pUnit->AI_setUnitAIType(eUnitAI);
+	}
+
```
RecursiveVision, line 2813: Is it necessary to switch their UnitAI back if plots become available to explore later? (e.g., once ocean crossing is unlocked)

KungCheops: This is to let the AI use them as military units until then. Unfortunately they are a bit prone to leaving them open to attacks but I think that's better than just having them afk.

KungCheops: The AI will turn them back into exploration AI if they are still around when ocean travel becomes available.

RecursiveVision:
> Teach AI how to use conquistadors to found cities.

In Community Patch Only, Conquistadors have `FoundAbroad` instead of `Found`, does your new AI logic account for this?

KungCheops:
> > Teach AI how to use conquistadors to found cities.
> 
> In Community Patch Only, Conquistadors have `FoundAbroad` instead of `Found`, does your new AI logic account for this?

I think that's covered in the "get best settle plot for unit" function, but can double check.

Hokath:
Looks amazing. The first two points especially I think are going to have a large effect. Will be interesting to watch it happen in Observer Mode.

KungCheops:
> > Teach AI how to use conquistadors to found cities.
> 
> In Community Patch Only, Conquistadors have `FoundAbroad` instead of `Found`, does your new AI logic account for this?

It is indeed covered in CvPlayer::GetBestSettlePlot (it calls pUnit->canFoundCity(pPlot)).

KungCheops:
Added: AI will no longer stop building work boats when they have many cities and only a few of them need work boats. E.g. they wouldn't build work boats when they had 5 cities and one of them needed work boats.

Edit: I may have been a bit hasty, this part of the code may be unused.

KungCheops:
Good news number 1: the old code was working as intended
Good news number 2: I think I've figured out why AI cities maybe weren't able to build work boats for each other in some cases

azum4roll:
> AI should no longer build units in small landlocked oceans (unless they are contested).

What if they aren't sure whether that ocean is landlocked or not?

KungCheops:
> > AI should no longer build units in small landlocked oceans (unless they are contested).
> 
> What if they aren't sure whether that ocean is landlocked or not?

In this case, they cheat. Mostly for performance reasons. Although it shouldn't make much of a difference I think, by the time they start building boats they should know whether a city is connected to a tiny ocean or not.

KungCheops:
Added a little buffer after a unit is built before they can be recruited for exploration so they have time to join operations (like settling)

RecursiveVision:
> Added a little buffer after a unit is built before they can be recruited for exploration so they have time to join operations (like settling)

Is that exclusive to units that don't have a DefaultUnitAI of EXPLORE or EXPLORE_SEA?

KungCheops:
> > Added a little buffer after a unit is built before they can be recruited for exploration so they have time to join operations (like settling)
> 
> Is that exclusive to units that don't have a DefaultUnitAI of EXPLORE or EXPLORE_SEA?

Those units start with AI set to explore, so they don't need to be recruited to exploration after being produced.

RecursiveVision:
> Force AI workers to never build on dig sites.

Egypt's new UI is an improvement that is built on Dig Sites.

azum4roll:
I'm not sure if Obelisks should still be built after you can dig up the artifacts.

CvGameCoreDLL_Expansion2/CvTacticalAI.cpp
```diff
@@ -4488,6 +4488,9 @@ CvUnit* CvTacticalAI::FindUnitForThisMove(AITacticalMove eMove, CvPlot* pTarget,
 			}
 			else if(eMove == AI_TACTICAL_GOODY)
 			{
+				if (MOD_BALANCE_CORE_GOODY_RECON_ONLY && !pLoopUnit->GetGainsXPFromScouting() && !pLoopUnit->getUnitCombatType() == (UnitCombatTypes)GC.getInfoTypeForString("UNITCOMBAT_RECON", true))
+					continue;
+
```
RecursiveVision: AI_TACTICAL_GOODY =/= pPlot->isGoody()

See the existing usage of MOD_BALANCE_CORE_GOODY_RECON_ONLY in CvTacticalAI.cpp.

KungCheops: I see, for some reason an AI decided to send a slinger across the whole continent to claim an ancient ruin though. So something is wrong with the old code.

KungCheops: I'll revert it anyway, must have been something else that caused it I guess.

RecursiveVision: How odd. But if that's so, the old code needs to be fixed!

KungCheops: That's gonna have to be some other time, this scope has creeped enough :)

RecursiveVision:
> Enable mod option for half move from terrain promotions.

I'm gonna be deleting this mod option, no sense in enabling it.

RecursiveVision:
It is deleted.

RecursiveVision:
@axatin There will be merge issues with my own changes so to save you some work I'll just merge this now.

PR #11981, opened 2025-10-04
Bugfixes

axatin:


CvGameCoreDLL_Expansion2/CvPlot.cpp
```diff
@@ -6650,10 +6649,6 @@ void CvPlot::setOwner(PlayerTypes eNewValue, int iAcquiringCityID, bool bCheckUn
 						}
 					}
 				}
-
-				// Remove credit for building Landmark if player took it from the minor
-				if (eImprovement == eLandmarkImprovement && eOldOwner == GetLandmarkCreditMinor() && GET_PLAYER(eBuilder).getTeam() == GET_PLAYER(eNewValue).getTeam())
-					SetLandmarkCreditMinor(NO_PLAYER);
 			}
```
RecursiveVision, line 6657: What was the issue here? It's intended that if you steal away the tile, you lose the bonus.

axatin: The bonus is lost anyway as the minor doesn't own the tile anymore

RecursiveVision: Yeah but if the minor say, captures the owning city, then the bonus would be reinstated.

axatin: ok I can add it back in

PR #11994, opened 2025-10-07
MP desync debugging with ungated features

JohnsterID:
For review, keep commits separate. First is MP debugging and second so we can test other FINAL_RELEASE code with VPDEBUG.


axatin:
interesting... how did you find out about that?

JohnsterID:
I was looking at Modded MP that was available prior to FINAL_RELEASE, couldn't figure it out besides UI so far. Moved on to all the gated FINAL_RELEASE gated code we still have that isn't compiled out in the binary we can use.

Changed back to draft as there are more gates that need to be reviewed.

edit: We only define _DEBUG in ResourceCompile not ClCompile so there are bunch of _DEBUG code that is not compiling in our debug builds as well.

_DEBUG could be added to our debug builds but we will need to test. Original plan, was to test gated FINAL_RELEASE code with VPDEBUG but the below showed a larger issue. Future issue, we can look at.

PR #11978, opened 2025-10-04
Alpa strategy text

Hokath:
Merged from Alpa's fork.
Have done a pass for spelling and grammar.
Need to check for apostrophe in sql files.

I've used newline in some xml strings to make double [NEWLINE] easier to find. Please delete if you think it's not appropriate.

Moved some UU text back to Civ folder as they were in Units folder as CS gifts etc. 
There might be others need same treatment?

PR #11962, opened 2025-09-27
GOTTA, SWEEP SWEEP SWEEP!

RecursiveVision:
Balance:
- `CvMinorCivAI::DoFirstContactWithMajor()` will now attempt to use the actual meeting player (where possible) instead of the team leader every time
- When Militaristic CS don't give a unit to the meeting player, they will instead give a bit of extra Influence
- Units lacking Strategic Resources can no longer heal from pillaging, Repair Fleet, etc.
- Adjusted the exception to conversion resistance for City-States when a Spread Religion quest is active
- City flat damage reduction is now applied before any damage is redirected to the city garrison

General:
- Most instances of `#if defined` have been removed
- Removed Personality column in Leaders table and associated trigger; replaced with helper tables
- Add NOT NULL checks to leader flavor tables
- Moved Barbarian flavor sweep to (1)
- Only change Cape Town's MinorCivSound in Community Patch Only
- AI compatibility for MOD_GLOBAL_CS_OVERSEAS_TERRITORY
- Performance improvement in `CvCity::DoSellBuilding()`
- Autoinstaller now deletes some very old VP versions
- Add GameSpeeds.ExperiencePercent
- Removed MountainsDoubleMove column entirely - use UnitPromotions_Terrains instead
- Removed HillsDoubleMove column from the DLL, but made it backwards compatible
- Use UnitPromotions_TerrainModifiers for vanilla Desert Warrior and Altitude Training
- Add IDEOLOGY_UNLOCK_BASE_FREE_TENETS

Text fixes:
- Fix Salon theming bonus text
- Clarify Indonesia UA
- The notification when buildings are consuming more Strategic Resources than you have now tells you the quantity that individual city is consuming above the limit

Bugfixes:
- Fix issues with `CvUnit::CanUpgradeInTerritory()` and MOD_GLOBAL_CS_UPGRADES
- Fix MOD_GLOBAL_CITY_JUNGLE_BONUS
- Fix MOD_GLOBAL_GRATEFUL_SETTLERS
- Fix MOD_GLOBAL_CS_OVERSEAS_TERRITORY
- Fix MOD_SANE_UNIT_MOVEMENT_COST
- Fix MOD_BALANCE_CORE_QUEST_CHANGES
- Fix MOD_BALANCE_CORE_ARCHAEOLOGY_FROM_GP
- Fix a MOD_EVENTS_REBASE event being called without the option being enabled
[...]

PR #11997, opened 2025-10-10
4UC tooltips, text changes and bugfixes

azum4roll:
Will rebase on @RecursiveVision's changes (#11962) when that PR is merged. Some of the mod options referenced in the UI files have been removed.

#11978 will also be merged into this later.

This is nowhere near done - still adding tooltips for corporation tables which I missed earlier.


azum4roll:
When we get to add a base influence column to Units (instead of relying on promotions), we won't need Help text for diplo units as well.

PR #12001, opened 2025-10-11
Remove BALANCE_GPTI_ARCHAEOLOGY; Better AI Archaeology Choice

axatin:


RecursiveVision:
If you delete a CustomModOption make sure to remove the MOD_OPT_CACHE in CustomMods.cpp and the MOD_OPT_DECL in CustomMods.h.

CvGameCoreDLL_Expansion2/CvCultureClasses.cpp
```diff
@@ -2934,42 +2958,37 @@ ArchaeologyChoiceType CvPlayerCulture::GetArchaeologyChoice(CvPlot *pPlot)
...
 			}
 		}
+		if (!bWithinWorkingRange)
+		{
+			eRtnValue = ARCHAEOLOGY_ARTIFACT_PLAYER1;
+		}
+
+		// Otherwise go for Landmark if from Medieval Era or earlier and we're not going for a cultural victory, or if we're very unhappy
+		else if ((!m_pPlayer->GetDiplomacyAI()->IsGoingForCultureVictory() && pPlot->GetArchaeologicalRecord().m_eEra <= 2) || m_pPlayer->IsEmpireVeryUnhappy() || m_pPlayer->IsEmpireSuperUnhappy())
+		{
```
RecursiveVision, line 2989: `IsEmpireVeryUnhappy() || IsEmpireSuperUnhappy()` is functionally equivalent to `IsEmpireVeryUnhappy()`.

CvGameCoreDLL_Expansion2/CvCultureClasses.cpp
```diff
@@ -2898,31 +2898,55 @@ ArchaeologyChoiceType CvPlayerCulture::GetArchaeologyChoice(CvPlot *pPlot)
...
 	if (pPlot->HasWrittenArtifact())
 	{
-		eRtnValue = ARCHAEOLOGY_ARTIFACT_WRITING;
+		PolicyBranchTypes eIdeology = m_pPlayer->GetPlayerPolicies()->GetLateGamePolicyTree();
+		if (eIdeology == NO_POLICY_BRANCH_TYPE)
+			return ARCHAEOLOGY_CULTURE_BOOST;
+
+		if (m_pPlayer->GetPlayerPolicies()->GetNumTenetsOfLevel(eIdeology, 3) < 2)
+			return ARCHAEOLOGY_CULTURE_BOOST; 
 	}
```
RecursiveVision, line 2929: So if I'm reading this right, the ArchaeologyOnly=1 Great Works of Writing give the player who obtains them the choice to collect their Great Work or get an instant Culture Boost, similar to the Great Writer's ability? I was actually unaware of this bit, I thought you had to collect the Great Work.

It's true that I almost always go for instant Culture when I get a Great Writer, but is it really worth always going for the instant boost over the per-turn yields?

RecursiveVision:
In terms of AI improvements, currently the AI will call `CvDiplomacyAI::IsBadTheftTarget()` with THEFT_TYPE_ARTIFACT to determine whether or not to send an Archaeologist into another civ's lands at all, but it would be smarter to use this function to instead determine whether the AI extracts artifacts instead of building a Landmark.

_That being said,_ it's probably more worthwhile for the AI to prioritize getting Culture and Tourism over small Happiness and relationship boosts, so if we're not dealing with a City-State (which gives a Resting Influence bonus and can fulfill a Quest) it may be wise to downprioritize players the AI doesn't intend to steal from.

RecursiveVision:
Related: Because the archaeology choice popup is partially hardcoded and janky, I resorted to adding `m_bWaitingForDigChoice` to the Diplomacy AI, setting it to true when a dig is completed, and setting it to false when a dig choice is made.

But this is a) janky, and b) prone to error if multiple plots with dig sites are completed on the same turn. Feel free to improve it.

CvGameCoreDLL_Expansion2/CvCultureClasses.cpp
axatin: Yes, it gives the player the choice. The instant culture option is very strong, it's the same amount of culture as a great writer would give, that's clearly better than all the other options.

axatin:
```
-- Third choice: Create Landmark or Cultural Renaissance
local itemInstance = g_ItemManager:GetInstance();
if (not bWrittenArtifact) then
	itemInstance.Name:LocalizeAndSetText("TXT_KEY_CHOOSE_ARCH_LANDMARK_HEADER");
	local strLandmarkText = Locale.ConvertTextKey("TXT_KEY_CHOOSE_ARCH_LANDMARK_RESULT");
	...
else	
	itemInstance.Name:LocalizeAndSetText("TXT_KEY_CHOOSE_ARCH_RENAISSANCE_HEADER");
	local strCultureText = Locale.ConvertTextKey("TXT_KEY_CHOOSE_ARCH_RENAISSANCE_RESULT");
	...
end
```
Looking at the lua file more closely, the landmark option isn't available when there's a written artifact, so I need to change my code.

axatin:
> Related: Because the archaeology choice popup is partially hardcoded and janky, I resorted to adding `m_bWaitingForDigChoice` to the Diplomacy AI, setting it to true when a dig is completed, and setting it to false when a dig choice is made.
> 
> But this is a) janky, and b) prone to error if multiple plots with dig sites are completed on the same turn. Feel free to improve it.

Changed it to an integer counter to solve the multiple dig sites issue

PR #12003, opened 2025-10-11
Fix enough settlers strategy condition

KungCheops:
AI would adopt the "stop building settlers" strategy as soon as they started building a settler. This meant that if the city did a reevaluation, it would stop building the settler. Could cause some AI players to build very few settlers.


azum4roll:
But wouldn't this lead to many cities building settlers at the same time when they need only one?

KungCheops:
> But wouldn't this lead to many cities building settlers at the same time when they need only one?

No

KungCheops:
The change I've made is "if this city is currently building a settler, don't count that one in the total number of settlers for this city evaluation". If another city is building a settler, that one will be counted.

PR #12007, opened 2025-10-12
Auto/AI great work optimizing/swap rework+fixes

KungCheops:
Swapping a great work with another civ now puts the great work you swapped away to them in their great work swap slot.

Automatic+AI great work optimization/swapping functionality now works.

Automatic+AI great work swapping is no longer limited to one foreign work swap per turn/button click. This means that the AI should finally be able to theme buildings that require only foreign works.

Fix possible desync issues with automatic great work swapping.

AI will now reevaluate their great work assignments whenever they create a great work during their turn instead of waiting until the next turn.

Worker AI fix: if a human player gets their first worker from a policy, it will now be able to start automatic working in the same turn.


Hokath:
wow ok that sounds like quite involved, can you explain simply what changed to make it function for the AI?

KungCheops:
> wow ok that sounds like quite involved, can you explain simply what changed to make it function for the AI?

The biggest issue was that it was assuming it could use the result of a DLL call to know where great works are. But that DLL call isn't handled instantly, essentially it was a race condition I suppose. I changed that so it keeps track of where things will be instead and then queue up all the DLL calls basically.

KungCheops:
That change was maybe 30 minutes of work, then I completely redid the foreign great work swap logic so it is much neater and also better, which took a couple of days.

PR #12087, opened 2025-10-19
Improvement text

Hokath:
Updated text descriptions for various improvements, including Aplekton, Oppidum, Obelisk, Villa, and Isibaya, to provide detailed information on their functions and historical context.

Question: The changelog suggests Villa doesn't claim tiles. Text is changed to reflect that. Correct?

Note: the improvement help text (as opposed to _build_ help text) is what appears in the tech tree.

Also made help text for Salon more similar to the automated tooltip (replace words with just icons)


azum4roll:
Spawning an adjacent resource claims the tile if it's unowned.

PR #12092, opened 2025-10-19
Tactical AI tuneup and Homeland AI improvements

KungCheops:
### Tactical AI:

Tactical AI evaluations were incorrectly including intermediate unit position scores, meaning most of the evaluation space was being taken up by nonsense moves. This has now been fixed with the evaluation only including final unit positions + the value of all attacks/pillaging/etc.

One effect of this is that the Tactical AI will no longer try to maximize the amount of moves for all units. This means, among other things, that it is now properly able to fortify melee units on the front lines and letting ranged units attack from behind. In general it should now be acting much more sensibly.

Improve tactical AI distance to target estimation, AI should no longer get stuck on mountain chains or on the wrong side of islands.

Add support in tactical AI evaluation for AoE damage on move.

Add support in tactical AI evaluation for restore moves from improvement.

Tactical AI evaluation for AoE damage from fortification is now consistent with other similar evaluations.

Tactical AI now takes current heal rate into account when decide whether or not to stay and heal (previously it was based on damage taken).

Because units are no longer incentivized to always move as much as possible, a small boost has been given to get them to move towards their targets.

Fix swap moves always being evaluated twice *and* having twice the priority of other moves in the evaluation queue. These two fixes should help free up a lot of evaluation space for better moves.

Should now be better at not letting ranged units die, e.g. when sieging cities.

Should now position fast ranged naval units further back in formations, meaning that they can now bombard cities with very fast ranged naval units without taking damage. (Sometimes they choose to stay in range of the city, probably because it's moving its frontline forward, not entirely sure).

Should be a bit better overall at positioning ranged units.

Should now be a bit better at taking barbarian camps with skirmishers.

Should now be a bit better at using recon units to scout but keeping them out of danger.

Should now be better at using relevant units as garrisons.

Will now consider healing as progress when evaluating tactical positions. (Positions that aren't considered as doing progress are discarded).

Will now try harder to put ranged units in positions where they can fire at enemies.

Will now pillage more aggressively with units without free pillage moves.

### Homeland AI:

Will no longer try to put garrison in all cities always, instead it will try to maximize garrison bonuses and leave tactical garrison moves to the tactical AI.

Can now perform opportunity attacks with homeland combat units.

Will now try to garrison units in a way that maximizes garrison bonuses.

Will no longer send workers across the world while at war.

AI exploring ships will no longer go home to heal as soon as they take damage.

AI will no longer use ranged ships as explorers.

AI explorers will now accept taking a bit more damage before healing.

Fix cases where exploring units would not take nearby ancient ruins if another explorer further away was moving there.

Sentry units will now heal even when not heavily damaged.

### Both:

Now able to pull garrisoned units out of cities if the city isn't threatened.

Will no longer treat naval units as if they give garrison bonuses they can't give.

Should be significantly better at using ranged naval units.

Skirmishers can now perform opportunity attacks from cities even if the target is not adjacent to the city (if the skirmisher can make it back to the city afterwards).

### Other:

AI settler operations should now be a bit better at sticking together.

Unit AI Production mod tweak, will now not build any bad units when unit balance is bad.

Small fix for unit production AI regarding evaluation of garrisoned units.

AI will now purchase work boats in the cities that need the work boats.

AI will no longer build work boats for unworkeable bonus resources.

Reorder "close to victory" conditions, go for the easiest ones first.


KungCheops:
This is currently being tested, hopefully I will be done with it by tomorrow, at the very least by next week.

CvGameCoreDLL_Expansion2/CvUnitProductionAI.cpp
```diff
@@ -1563,25 +1563,25 @@ int CvUnitProductionAI::CheckUnitBuildSanity(UnitTypes eUnit, bool bForOperation
 			if (kPlayer.GetMilitaryAI()->IsUsingStrategy(eStrategyArcher))
 				iBonus += iBonusValueFromStrategy;
 			else if (kPlayer.GetMilitaryAI()->IsUsingStrategy(eStrategyEnoughArcher))
-				iBonus -= iBonusValueFromStrategy;
+				return SR_STRATEGY;
```
RecursiveVision: Think it should be `return SR_BALANCE;` - might wanna check the other SR_ instances to make sure they're actually correct with the enum.

RecursiveVision:
Good to see that we'll still get Tactical AI fixes despite ilteroi's retirement!

CvGameCoreDLL_Expansion2/CvUnitProductionAI.cpp
KungCheops: Will check SR usages and update

KungCheops:
I think this is working well enough for my liking now, gonna do a little more testing then I'll set it to ready

RecursiveVision:
Worth noting that (AFAIK) the Tactical AI has no support for withdrawal from melee / heavy charge / withering fire. These were adjusted to be more deterministic rather than random for AI friendliness, but the actual compatibility has yet to be added.

Note also that the CORE_CIVILIANS_RETREAT_WITH_MILITARY option might be worth checking in escort missions.

RecursiveVision:
Tactical AI is also unable to handle the Battering Ram, if memory serves. Unsure if this also applies to the Siege Tower or Great Bombard.

Not necessarily things that need to be addressed now, but if you plan on fixing up Tactical AI, worth bringing these to your attention.

CvGameCoreDLL_Expansion2/CvAStar.cpp
```diff
@@ -2330,6 +2331,60 @@ static int SeaWorkerUnitSafeValid(const CvAStarNode* parent, const CvAStarNode*
...
+	//we could check the trait directly but theoretically we could have an old army of non-oceangoing ships ...
+	if (pToPlot->isDeepWater() && finder->HaveFlag(CvUnit::MOVEFLAG_NO_OCEAN))
+		return FALSE;
+	if (pToPlot->isIce() && !kPlayer.CanCrossIce())
+		return FALSE;
+
+	PlayerTypes ePlotOwner = pToPlot->getOwner();
+	if (ePlotOwner != NO_PLAYER && ePlotOwner != data.ePlayer && !kPlayer.IsAtWarWith(ePlotOwner) && !kPlayer.GetDiplomacyAI()->IsHasOpenBorders(ePlotOwner))
+		return FALSE;
+
```
RecursiveVision, line 2394: Would rather not rely on GetDiplomacyAI() for non-major civs. Longterm it'd be ideal to get rid of all Diplo AI calls for non-majors. You can move the helper function to CvPlayer if it'd be helpful; if you do, just change the Diplo AI function to call the CvPlayer function.

KungCheops: On it

Issue #12034, opened 2025-10-14
String error


AleksM-glitch:
### 1. Mod version

5.0-alpha.01

### 2. Installed Components

Vox Populi with EUI

### 3. Describe the Issue

Just a minor issue, but it seems like the the strings here are broken for getting an unit from a militaristic CS the first time you meet them. 

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e97843a1-9532-4395-8203-361fbb86a114" />

How to reproduce: discover the CS of Sogut with the Scout. 

### 4. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[Bug save.zip](https://github.com/user-attachments/files/22909922/Bug.save.zip)

azum4roll:
So you're somehow getting a unit despite not being the first civ to meet the CS?

azum4roll:
Seems you can indeed always get a unit now. @RecursiveVision is this intended?

AleksM-glitch:
That can't be quite right. I've encountered multiple militaristic CS, and I've not gotten any units (though maybe that's related to the personality).

azum4roll:
It's still RNG, but it's rolled for every player, not just the first one they meet.

azum4roll:
@axatin how should we handle this?

axatin:
I think it's unintended. The logic was changed in https://github.com/LoneGazebo/Community-Patch-DLL/commit/96968ae017d64d79e17e3f63147889985d395d05, which is a larger refactor. The commit message doesn't mention any change to how units gifts depend on being the first player or not

PR #12112, opened 2025-10-22
Fix first contact gift logic (#12034)

axatin:
This code was accidentally deleted in 96968ae


RecursiveVision:
Not sure whether or not this was the case prior, but I don't see the reason for zeroing out Gold and Food. The amounts are already not super significant, and would be halved if treated like the other yields.

Maybe the free unit makes sense, but halving the odds would also work there IMO.

axatin:
I think it's better if no free units can be given to subsequent players. Halved Gold and Food yields instead of zeroing them out.

PR #12119, opened 2025-10-23
Add mod option for no resource yields with GP improvement

KungCheops:
Also includes any improvement with the ConnectsAllResources property.


RecursiveVision:
To maintain standardization:
- The #define should have the MOD_ prefix.
- Balance changes that aren't enabled by default should be placed in the files in alphabetical order.

Issue #12076, opened 2025-10-18
CTD during AI turn


villeinkinen:
### 1. Mod version

5.0-alpha.01

### 2. Installed Components

Vox Populi with EUI

### 3. Describe the Issue

In two separate games I had crash to desktop at around turn 100 during AI turn. There is no error message.

### 4. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[Theodora_0107 BC-0200.zip](https://github.com/user-attachments/files/22982708/Theodora_0107.BC-0200.zip)

### 5. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

[Logs.zip](https://github.com/user-attachments/files/22982712/Logs.zip)

axatin:
It's a reproducible crash, but it originates in the exe, not in the game core dll. Does anyone know how to debug this? Maybe @vaderkos @JohnsterID ?

<img width="1351" height="518" alt="Image" src="https://github.com/user-attachments/assets/6939554a-aaca-4c5f-9fb1-dea5eac0f3d6" />

azum4roll:
Is the whole stack in the exe?

axatin:
yes

vaderkos:
@axatin is this the whole stack?

Hello71:
from registers and disassembly in #12105, looks like edi points to a valid memory location, but eax is -1*0x4b0. this should be the corresponding Linux function:

<details>
<pre>
 887db58:       55                      push   ebp
 887db59:       53                      push   ebx
 887db5a:       57                      push   edi
 887db5b:       56                      push   esi
 887db5c:       83 ec 1c                sub    esp,0x1c
 887db5f:       e8 00 00 00 00          call   887db64 <_ZN19cvLandmarkVisSystem17LandmarkRenderJob7ExecuteEj@@Base+0x1790>
 887db64:       58                      pop    eax
 887db65:       81 c0 9c 94 a8 01       add    eax,0x1a8949c
 887db6b:       89 44 24 18             mov    DWORD PTR [esp+0x18],eax
 887db6f:       8b 44 24 34             mov    eax,DWORD PTR [esp+0x34]
 887db73:       3d ff ff ff 0f          cmp    eax,0xfffffff 
 887db78:       0f 84 cb 00 00 00       je     887dc49 <_ZN19cvLandmarkVisSystem17LandmarkRenderJob7ExecuteEj@@Base+0x1875>
[...]

PR #12125, opened 2025-10-24
Fix CTD/game freeze caused by wrong Aplekton file names (#12076)

axatin:


JohnsterID:
Great work, anyway we could add validation or assert somewhere in the DLL?

Issue #12082, opened 2025-10-18
Civilopedia doesn't capture key presses


JohannesGaessler:
### 1. Mod version

5.0-alpha.02

### 2. Installed Components

Vox Populi with EUI

### 3. Describe the Issue

Not sure if this is a VP or a Civ 5 bug but I'm noticing that if you open the Civilopedia it doesn't capture key presses unless the text field for searching is focused. So if you e.g. select a settler and press B in the Civilopedia you found a City in the background.

axatin:
@vaderkos

azum4roll:
Hasn't this always been the case? Pressing enter on the search bar would pass your turn (if possible), for example.

vaderkos:
Maybe input is not consumed correctly...

vaderkos:
@JohannesGaessler how do you open the Civilopedia: 
1. using mouse on some entry to check info?
2. using keyboard shortcut?
3. or using using civilopedia button at the top right?

Is it reproducible in each case?

JohannesGaessler:
When I press F1 or when I e.g. right click on a Technology the Civilopedia is opened and the text box is focused automatically. When I open the Civilopedia by clicking its symbol on the top right the text box is not automatically focused (but the foxue state from having it opened previously is conserved). When I click on the text on a Civilopedia page the text box is unfocused. When I press any keys while the text box is not focused those keys are affecting the units I have selected in the background.

PR #12130, opened 2025-10-25
Fix civilopedia input handler

vaderkos:
Should fix https://github.com/LoneGazebo/Community-Patch-DLL/issues/12082


azum4roll:
Can ContextPtr actually be nil?

vaderkos:
> Can ContextPtr actually be nil?

It should not ever be nil. However, after the latest bugs with disappearing globals, I do not trust anything XD

PR #12133, opened 2025-10-26
Change Klepht CombatClass from SIEGE to RECON

Hokath:
typo is giving it siege properties in the sweeps!

I think this is the only issue? Are there other fields anyone can think of that might need updating?


Hokath:
fixes #12135

Issue #12161, opened 2025-11-01
University of Coimbra lacks all Theming bonuses


Alpakinator:
### 1. Mod version

5.0-alpha.03 (but with removed IGE)

### 2. Installed Components

Vox Populi with EUI

### 3. Describe the Issue

University of Coimbra should have +4 Science theming bonus on top of the 10 culture and 5 tourism bonus. Same as Imperial College. However it doesn't have both, so literally no theming bonuses. Even if filled with 2 works of writing from different civilizations and eras as seen on the screenshots:

<img width="1445" height="849" alt="Image" src="https://github.com/user-attachments/assets/19acaa70-16e8-4aab-9474-93c25a49ed8a" />

<img width="1447" height="851" alt="Image" src="https://github.com/user-attachments/assets/2d9d8264-d4d4-4575-9ec8-125b8d20e3d5" />

<img width="1446" height="851" alt="Image" src="https://github.com/user-attachments/assets/d505d831-c289-41e8-aeda-b4e77b9f18bf" />

in 1st 2 imgs you can see the different great works, and on all 3 the culture and science per turn are the same so it's not a visual bug. Uni of Coimbra lacks theming code in Portugal.sql.

PR #12162, opened 2025-11-01
Imperial College theme copied to Coimbra

Alpakinator:
University of Coimbra lacks any theming at all. It should have theming identical to Imperial College according to the passed proposal.


Alpakinator:
fixes #12161

PR #12182, opened 2025-11-09
Add CPK.Misc.CustomModOptionEnabled

vaderkos:


vaderkos:
@azum4roll, @axatin let me know if you'd like it to rather throw an error if there is no custom mod option corresponding to name (typo in name).

azum4roll:
The CvLuaGame version doesn't seem to throw errors. Default false is the current behaviour so I guess it should stay.

azum4roll:
Can you also make one for Defines?

vaderkos:
> Can you also make one for Defines?

Do we have flags in Defines? Is ok something like `DefineEnabled`?

azum4roll:
Defines are name-value pairs, accessed using the GameDefines table ingame.

PR #12192, opened 2025-11-15
Slightly Increase the High Level AI Early Game Bonuses

ron-murhammer:
Currently, its a bit too easy to conquer the Deity AI early in the game (first 3 ages) as it seems VP shifted too far from vanilla which gave too many early game bonuses (see recent Deity playthroughs on discord). I would propose the following to give the AI a bit more supply (needs this especially after the naval rework as it builds ships too early and frequently now) and adjust building/combat unit/non-world project cost so that it starts lower but scales a bit slower to help them build up a little faster while keeping about the same late game discount.

Deity
- Increase AI base supply from 12 to 15
- Update AI building/combat unit/non-world project cost from 70% to 60%
- Update AI building/combat unit/non-world project cost per era from 10% to 8%

Immortal
- Update AI building/combat unit/non-world project cost from 75% to 70%
- Update AI building/combat unit/non-world project cost per era from 9% to 8%

PR #12208, opened 2025-11-20
[Balance] Increase Deity AI Base Supply

ron-murhammer:
Revised PR with reduced scope based on feedback on https://github.com/LoneGazebo/Community-Patch-DLL/pull/12192

Currently, its a bit too easy to conquer the Deity AI early in the game (first 3 ages) as it seems VP shifted too far from vanilla which gave too many early game bonuses (see recent Deity playthroughs on discord). I would propose the following to give the AI a bit more supply as it struggles to focus enough units to counter a human player (needs this especially after the naval rework as it builds ships too early and frequently now which wastes its limited supply).

Deity
- Increase AI base supply from 12 to 15


Hokath:
Seems quite targeted and safe. 
If Deity players want it a bit harder at the start militarily, this sounds like a minimal way to get that?
Is this shared amongst those players? Probably? It's hard to know --> maybe congress

PR #12218, opened 2025-11-21
[Balance] Reduce Ulticur Per War Prod Bonus From 10% to 5%

ron-murhammer:
With a standard 7 major opponents, the Huns can DoW on everyone while they are often only adjacent to 1-3 enemies to get a total of +70% prod bonus which is way too high. This cuts that in half to a max of +35% which is more reasonable but still significant.

Requires a Project Overseer to approve out-of-congress balance change. Deity play through here shows how abuseable this is for the human player while the AI won't consider doing something similar: https://discord.com/channels/707403015845576735/1430509448790540478


azum4roll:
Maybe add 1 production to Arenas to compensate.

ron-murhammer:
Yeah I think that's reasonable to compensate for the decreased per war prod percentage. Updated.

Hokath:
Might it be better to have a cap instead?
Besides the identified problem, this scales with map size in a way that's not healthy.

So "+10% (max 30%)"

ron-murhammer:
@Hokath Maybe. The downside I see is 10% per war is potentially enough to justify declaring war on faraway civs that you have no intention of fighting just for the bonus. So then you could declare war on say 3 civs on another continent for +30% to build your military to then invade your actual neighbor :)

Hokath:
Yeah the bonus just incentivizes the human to play in a weird way I guess.
We should think up an alternative bonus. 
Perhaps we could make "wars won (25+ warscore)" an accomplishment, and have like 5% per war won (table already exists).

ron-murhammer:
@Hokath Yeah ideally the bonus makes you have to actually fight/win wars in some way not just declare them. That I think reduces the weird abuse that human player can do with it. But that is probably a Congress change :)

azum4roll:
Huns specifically want to stay at war with the war weariness bonus. Leave the war victory thing to Aztecs.

PR #12219, opened 2025-11-22
[Balance] Reduce Bandeirante Yields From Scouting From 3 To 2

ron-murhammer:
Bandeirante yields from scouting are way too high and generally push Brazil to minimize its scouting until they are unlocked. On a standard map, they give around 2-3 policies and 8-12 techs in quite a short period of time for exploring the rest of the map. This makes them one of the best UUs even though they should be more of a secondary part of Brazil's kit. This change makes it so they are strong but don't give a crazy amount of gold/science/culture.


Hokath:
We should do this, it is indeed very broken.

PR #12233, opened 2025-11-27
Try fix CPK loader in specific situations

vaderkos:
Fix possible missing global variables.


axatin:
With the missing variables bug fixed this is no longer needed, right?

vaderkos:
> With the missing variables bug fixed this is no longer needed, right?
You can merge or close. Should not influence much if fix in LoadScreen works.

PR #12242, opened 2025-12-01
Add changeMinorPlayer

jwanderson2000-hash:
Adds api extension for changing minor player using Game.ChangeMinorPlayer(PlayerTypes eNewPlayer, MinorCivTypes m)


azum4roll:
Need checking so you don't set a major civ slot as minor.

PR #12249, opened 2025-12-03
Building theming bonuses derived help text

Hokath:
Added the long-missing information about theming bonus values (not to be confused with yield-if-themed!) to building and wonder HELP entries. We don't automate these, yet.


azum4roll:
Should we be mentioning the civ name in UBs instead of using "your"? What if the trait of capturing UBs ever come back?

Hokath:
Personally I would not use the civ name, and just stick with "Domestic" because its consistent between tooltips that way.
Here I kept the name for consistency with Recursive's earlier suggestion for the Salon.
Shall we change?

azum4roll:
Yes. But if "Domestic" a standard term? I don't see that being used in any description.

Hokath:
Standard in game, no. Closest is the tooltips with "X works created by you"
I liked Domestic because 
a) its the opposite of Foreign, with which it is mutually exclusive, so they can go as a "prefix" in the same place in every tooltip
2) its short
iii) "created by you" is maybe ambiguous for Artifacts

PR #12250, opened 2025-12-04
add BALANCE_HALF_XP_FAITH_PURCHASES

hyuckkim:
A custom option has been added to halve the experience points gained when a unit is purchased with Faith.

The change is not applied unless the additional setting is configured/enabled.


RecursiveVision:
If it's disabled by default please put it in the Other Balance Options category, not the Vox Populi Balance Changes category. To keep things organized :)

hyuckkim:
Is this right? ;>

PR #12251, opened 2025-12-04
[Balance] Decrease Sovereignty Exponent From -20% to -10%

ron-murhammer:
Sovereignty's exponent decrease is way too high as it goes from 1.35 to 1.08. This has led to the OP "border blob" strategy of taking Authority plus grabbing Sovereignty for insane border growth yields. This cuts the percentage in half to make it more reasonable yet still make Tradition feel like you get cities with more tiles. 

Details can be found in this forum thread: https://forums.civfanatics.com/threads/vox-populi-congress-proposal-workshop.683872/page-35#post-16830658. I went with a more conservative adjustment than proposed in that thread to start and figure it can be further adjusted if this still seems too strong.


axatin:
This can wait until the next congress. The AI doesn't use that strategy, and humans are free not to do so either.

ron-murhammer:
@axatin Ok. I actually thought because the AI won't consider not finishing policy trees that adjusting this was high priority since it seems that one of VP's focuses is on making sure the AI is able to play the same game as the human player?

PR #12261, opened 2025-12-09
Part 1 (or POC) of a reworked include system, for massive reduction of lua's memory consumption and (not measured yet, but likely) loading time

schnetziomi5:
Description/explanation: See comments
...for additional context also see Vaderkos' Discord channel where I wrote about the ideas behind this PR

* Commit title 'Adjust GameInfo and InfoTooltipInclude in (1)' should say adjust 'InGame' instead of 'GameInfo'
* When referring to memory consumption in tests, unless indicated otherwise, this always refers to the value returned by lua's collectgarbage("count") function, after the (very same) savegame is loaded with EUI active and LuaJIT replacing the stock lua51.dll

### DIFF
--- a/VPUI/LuaVP/newGameInfo.lua
+++ b/VPUI/LuaVP/newGameInfo.lua
@@ -389,7 +389,7 @@ local MT_GAMEINFO_OBJECT = {
 				GI[k] = nil ;
 				collectgarbage("collect") ;
 				local m2 = collectgarbage("count") * 1024;
-				--print( m - m2, k );
+				print( m - m2, k );
 				tot = tot + m-m2 ;
 				m = m2 ;
 			end
@@ -412,7 +412,7 @@ end
 GameInfo = newGameInfo ;
 
 
-
+--[[
 
 -- DEBUG STUFF
 --


schnetziomi5:
Backstory: I recently made a reworked LeaderHeadRoot screen, that is, I started with a blank file incrementally adding UI/logic to match the current versions capabilities. Among the way I developed system to declare dynamically updated Buttons/UI components that made for very efficient and concise writing of code( resembling the look and functionality of paradox' scripting language used in e.g. Stellaris or CK2) but also involved declaring a lot of lua tables. That's way I then built some sort of memory tracking system, which - because data to compare the footprint of my files with was needed - I eventually expanded to cover every lua file ever included.

This tracker was enabled to work by putting a single line of code on top of every lua file that gets loaded as a context( e.g. by ContextPtr:LoadNewContext or by being declared in a xml file as a  <LuaContext> object):
`if debug and debug.TRACKME and debug.TRACKME() then return end`
debug.TRACKME was a function set by the earliest loaded lua file I could find, which was ToolTips.lua. Basically it would log current memory usage of lua, call include(env().StateName) to load/run the file that just called once more, then upon being called immediately again just return false to let the code execute, and upon returning from include(env().StateName) log the memory usage again. It would also replace the original include of the calling thread with a wrapped version that allowed it to log all subsequent includes during execution of the original file and log memory consumption before/after those subsequent include calls (and also the files included). The original call would then return true, so that execution of the original caller would not go beyond the if clause (since it was already run before anyway)

Two important findings: 
* Some file were included quite often, creating the same objects every time. In in terms of how often InstanceManager is leading with being loaded 86 times, for ~3700 bytes of persistent memory consumption, and another ~8400 that got collected after the file was finished executing(compile/IO data). Not too much individually, but multiplied by 86 it stacks up. InfoTooltipInclude is loaded 35 times creating identical objects(functions) - it doesn't keep state of anything, for ~4mB of total memory consumption after garbage collection.

* The gameinfo object provided by fireaxis is terrible. And that's an understatement for language reasons. Some files showed huge spikes in (persistent) memory consumption during their first run only. When investigating I found that those memory increases came from calls to GameInfo. Worst offender: The first time GameInfo.UnitPromotions was accessed, memory consumption would permanently rise by 10mB. One has to assume the people who coded this were not just not trying to be efficient, but actually the opposite - or had no understanding of lua whatsoever. The worst crime they committed is, that for each row of a database table, GameInfo creates second internal lua-table-object just holding 'true' for all columns that exist in that database table - not once per table, _once per row_. That alone basically doubles the memory footprint of its cache...

schnetziomi5:
This PR proposes a solution to both of those problems. Importantly it does so purely in Lua, that is without using debug functions or needing dll/exe support. It is also intentionally designed to try to not break any code that is not aware of the new system proposed.

Core Concept:
Similar to how the memory tracker mentioned above worked, every lua file controlling a root context makes a call to a globally available function, which can then overwrite the predefined globals (most notably include and GameInfo) of the calling thread(think file) to whatever suits us better...

With that in mind I will now explain the changes/additions in this PR:

**Folder Lua_Expansion2/UI**
This is just a collection of all lua files that are present in the unmodded game with an additional line added on top.
`if CodeBuddy.vpRegisterContext and not CodeBuddy.vpRegisterContext() then return end`
Manually picking those that are not overwritten when using CP/VP or EUI would have been significantly more work and would be more unstable when files in the actual mods are added/removed. The Civ5Pkg file registers the Lua_Expansion2 folder as with priority 201, actual expansion 2 uses priority 200, VPUI/EUI are using 250. As a semi intended(POC) side effect, having only VPUI in the dlc folder, but not loading mods, now reduces lua memory consumption ingame by about 9mB(down to 13mB from 22mB)

**Files in LuaVP**
LuaVP holds some newly added files. InGame.lua, WaitingForPlayers.lua and ToolTips.lua are essentially used to signal 'vploader' what's happening with the engine.

InGame.lua is basically the same as InGame.lua in Lua_Expansion2/UI( that is expansion2's InGame.lua with a line on top), but this one also sets a callback for ContextPtr:OnShutdown(), which it uses to reset the references to gamecore objects in the thread 'vploader' is living in. 

WaitingForPlayers does the opposite, it's a context that doesn't live for long, but exists for a short time even before the LoadScreen, and crucially has already access to Game, Players, Teams, etc... - it packs those into a table and hands it to 'vploader'. 

ToolTips is another fascinating context that lives through loading and unloading a game, so it's the context 'vploader' lives in as well. It's also the first context that's loaded when the game boots, which ensures that the callback CodeBuddy.vpRegisterContext is already there when the UI gets built by the engine.

newGameInfo is, as the name suggests, a replacement for the GameInfo object that's available by default. It fetches the data from the database and uses a mix of dense arrays to store highly populated columns of a table and string indexed tables where only the non default entries of the remaining columns for each row are set. During testing I found that caching every table accessed during a game load takes about ~1,7mB of memory (compared to ~25mB for fireaxis' GameInfo). Also it's supposed to be fully compatible with any code using GameInfo...

(Vaderkos is also developing a replacement for GameInfo, so this one might be swapped out)

**vploader.lua**
The core file of the proposed system. Basically it includes a set of predefined files, fetching the objects(mostly functions) they generate and stores them in a globally available table: CodeBuddy.vpStaticObjects.

Also it sets the vpRegisterContext function that now every context is supposed to call when first loaded. As of currently this function always overwrites GameInfo with a (global) instance of newGameInfo and include with vpinclude, a function that when called checks if the required file is statically available and, if so, imports the globals from vploader into the callers thread instead of including the file again. Otherwise it will run an actual include operation

**InstanceManager and InfoTooltipInclude in Lua_Expansion2/AdaptedIncludes; Changes to InfoTooltipInclude in (1/CP)**
These are meant to be first examples(POC) of how to tweak files to make them work when statically included with the proposed system.

InstanceManager showcases the slight tweaks required to make it so that it can be used from any calling thread. ContextPtr has to be replaced by loadstring("return ContextPtr")().

InfoTooltipInclude in AdaptedIncludes and InfoTooltipInclude in (1) are meant to show that's possible to support statically including files that come with different objects based on what mods are loaded. InfoTooltipInclude in (1) is also modified slightly so that iTrainPercent works.

**Other files**
Notably content switch is a context that has its Context-Visibility toggled by the game engine directly - every time when mods are loaded/unloaded. So a callback is put in place to signal 'vploader' to update its library of static objects when that happens.

Any lua files already in VPUI hav had the hook line added on top
On Shutdown in InGame.lua in (1) now also calls vpSetGameCore({}) to delete the references to the gamecore's objects from vploader's thread..

schnetziomi5:
In this PR basically no lua file in (1),(2) or (3) is yet changed to make use of this new system. Loading the game with the proposed VPUI files and either playing without mods or loading VP on top works just fine and breaks nothing.

Further steps if this gets adopted:

* Every lua file in (1) and (2) that is associated with a context needs to have the hook line added on top for full efficiency.

* VPUI and CPK should have some of their logic changed and probably not care about ModMapData

* Other common includes would need to be checked and possibly tweaked if they were to be made available statically

* EUI can also be made to work with this, but that's a little more complicated because it also lives in the dlc folder and per Azum4roll's explanation we aren't allowed to change anything in there because of copyright.

Please everyone interested, comment, criticize and ask questions....

axatin:
@azum4roll can you review this?

schnetziomi5:
Locally I have now also added the hook line to the lua files in (1) and (2) - everything works fine and memory is down to 19Mb from 46Mb, however when committed to a temporary branch, Github thinks all files are completely new e.g. -1700/+1701 lines. Likely because I used a script? Obviously won't push it like that, working on it atm..

schnetziomi5:
This was mean to be pushed as one commit with this comment but I have to yet learn Github's ways I guess:

**General improvements to this PR and preparations for integration with lua files in (1) and (2)**
* Fix filters for GameInfo iterators in TechButtonInclude in (1), that somehow work with the stock GameInfo, even though they actually shouldn't
* Properly gate all function calls in (1) behind 'if' statements so that everything will work normally if vploader is disabled
* make sure `local VP = VP or MapModData.VP` is used everywhere in (1) instead of various different retrieval logic

* newGameInfo: Add warning for non existing column access, but only shown once per table. Fix the warning shown when string conditions are used to iterate on a table that has no ID column
* Added new functionality to vploader:
* -> iTrainPercent in InfoTooltipInclude can be a number again, that is updated when the gamecore is set/unset
* -> CPK gets a special include that makes sure everything is included in vploader's context
* -> added warnings when the same UI is loaded twice or when the same context registers twice (neither is a critical problem)

schnetziomi5:
@azum4roll @axatin 
See [here](https://github.com/schnetziomi5/Community-Patch-DLL/pull/6) for the branch where all the lua contexts in (1),(2) and (3) use vploader. The game loads normal, memory consumption (EUI) is down by whopping 40Mb (~16 from ~56) after loading a savegame. That's almost suspiciously much... but then again the stock GameInfo is overwritten in every context and all includes of (for now) InstanceManager, InfoTooltipInclude, CPK, VPUI, EUI_utilities and EUI_tooltips are rerouted, reusing the already existing objects instead...

Also, notably, there now exists a kill switch, that is by deleting 'enable_vpl.lua' the whole system is disabled(after reloading mods or restarting)... which will break nothing, since the changes to the existing files are minimal and function calls are gated by ifs....

PR #12268, opened 2025-12-10
5.0 fixes

azum4roll:

PR #12271, opened 2025-12-10
Fix Gumey help text from Each to If TRs

ron-murhammer:


azum4roll:
Fix included in [bc82357](https://github.com/LoneGazebo/Community-Patch-DLL/pull/12268/commits/bc82357b1b8a2a688d330e8234bcbc640635a895)

PR #12276, opened 2025-12-11
perf: Trade route from Buildings and Techs

qingyin3545:
Cached the number of trade routes provided by Buildings and Technologies to reduce redundant calculations and slightly improve performance.


qingyin3545:
Reopened the PR after accidental closure. No changes to the content.

PR #12309, opened 2025-12-17
GPTI Build Help fix and check-up

Hokath:
as per discord general chat


azum4roll:
There's no reason to stick to using CONSUMED as vanilla (which is known to be inconsistent anyway).

Hokath:
The existing one (Ordo) used it, so I stuck with it.
Probably this would trip someone up though so yeah, shall we just ditch it?

azum4roll:
Can you also restore the order in NewBuilds.xml?

Hokath:
Sorry Azum I missed your message. What was the problem with sectioning off the GPTI, it should match somewhere else or?

azum4roll:
Well the UIs were sorted by civ name (and portrait index) and there isn't really a reason to have another section.

PR #12318, opened 2025-12-20
Replace _diff for Oromo

Hokath:
It was DXT compressed, but the size was wrong, 1 pixel too many and one too few. Probably someone moved the layer with copy/paste?  Anyway, rounding to the nearest multiple of 4 and DXT5+mipmaps fixes it in-game for me


Hokath:
this is in #12370 somehow

PR #12339, opened 2025-12-27
Resolve Happiness Tooltip Rendering Issues in Non-English Locales

### LINKED ISSUE #12338: Brace formatting is not working with non-ascii text
### 1. Mod version

5.0.1

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

[Handani's Korean translation](https://github.com/handanikr/vp_kr)

### 4. Describe the Issue

When translate we changed the Happiness to 행복 like that:

```sql
-- Building
UPDATE Language_en_US
SET Text = '{1: number "''[ICON_HAPPINESS_1] Happiness: ''#;''[ICON_HAPPINESS_3] Unhappiness '':#"}'
WHERE Tag = 'TXT_KEY_PRODUCTION_BUILDING_HAPPINESS';
```

```sql
-- Building
UPDATE Language_ko_KR
SET Text = '{1: number "''[ICON_HAPPINESS_1] 행복: ''#;''[ICON_HAPPINESS_3] 불행: ''#"}'
WHERE Tag = 'TXT_KEY_PRODUCTION_BUILDING_HAPPINESS';
```

And it's not shown.

Image

_No response_

_No response_

hyuckkim:
### Overview
This PR addresses a UI bug where building happiness tooltips would fail to display or render incorrectly in certain localized versions of the game (specifically reported in the Korean locale).

### The Problem
Commit ef20092 attempted to simplify the happiness tooltip by using a single XML key with positive/negative number formatting (#;#). While efficient in English, this approach is incompatible with the game engine's handling of multi-byte characters.

When the parser encounters Korean characters wrapped in literal delimiters ('') within a formatting block, it fails to align bytes correctly, causing the entire tooltip line to disappear.

### The Solution
I have partially reverted the happiness logic in InfoTooltipInclude.lua to use Lua-side branching:

VP Path: Maintains the current separation of Happiness and Unhappiness lines.

Standard Path: Restores the if (iHappinessTotal < 0) check to explicitly switch between HAPPINESS and UNHAPPINESS keys.

This ensures that the UI remains robust regardless of the language setting and fixes the regression introduced 3 months ago.

### Related Issue
Fixes #12338

### DIFF
--- a/(1) Community Patch/Core Files/Overrides/Includes/InfoTooltipInclude.lua
+++ b/(1) Community Patch/Core Files/Overrides/Includes/InfoTooltipInclude.lua
@@ -1878,7 +1878,14 @@ function GetHelpTextForBuilding(eBuilding, bExcludeName, _, bNoMaintenance, pCit
 		AddTooltipNonZero(tYieldLines, "TXT_KEY_PRODUCTION_BUILDING_HAPPINESS", iHappinessTotal);
 		AddTooltipNonZeroSigned(tYieldLines, "TXT_KEY_PRODUCTION_BUILDING_UNHAPPINESS", kBuildingInfo.Unhappiness);
 	else
-		AddTooltipNonZero(tYieldLines, "TXT_KEY_PRODUCTION_BUILDING_HAPPINESS", iHappinessTotal + kBuildingInfo.Unhappiness);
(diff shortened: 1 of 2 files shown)


azum4roll:
This wouldn't even work. One text key is signed while the other is not.

PR #12344, opened 2025-12-28
fix/workaround for #12332 (MaxGlobalInstances precondition hit)

### LINKED ISSUE #12332: Crash: getmaxglobalinstances
### 1. Mod version

5.0.1

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

Enabled mods:
IGE+ v46
squads for VP
VP
CPP
3 units per tile
cities 4 hexes away
capture great people
reforestation

### 4. Describe the Issue

I spent 3 hours setting up my game using IGE, getting all my civs cities placed, bonuses started, etc etc. Upon clicking next turn I get this bullshit error. It looks like it's saying there's a problem with the amount of a specific building. I gave some wonders to multiple Civs. Using IGE I have tried to determine what the issue is with a specific building and cannot find one. The issue just happens eventually either on the first turn or multiple turns in without modifying civs at all. I'm at a loss.

Image

[IGE3.zip](https://github.com/user-attachments/files/24332039/IGE3.zip)

_No response_

schnetziomi5:
somewhat fix #12332 by changing the MaxGlobalInstances precondition check into a (less disruptive) assertion...

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvGame.cpp
+++ b/CvGameCoreDLL_Expansion2/CvGame.cpp
@@ -8103,8 +8103,6 @@ bool CvGame::isBuildingClassMaxedOut(BuildingClassTypes eIndex, int iExtra)
 		return false;
 	}
 
-	ASSERT(getBuildingClassCreatedCount(eIndex) <= pkBuildingClassInfo->getMaxGlobalInstances(), "Index is expected to be within maximum bounds (invalid Index)");
-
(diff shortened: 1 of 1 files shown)


axatin:
Since modmods can always add buildings that exceed the instance limits, I think it's better to remove the assert completely. That's unfortunate, but so far there weren't any issues with max instances, so it shouldn't be a problem.

schnetziomi5:
👍 adjusted

PR #12351, opened 2025-12-29
Trait_FreePromotionUnitCombats prereqtech check

Hokath:
Added prerequisite tech check for Trait_FreePromotionUnitCombats promotions.

Right now this causes no change, but I would like it for an implementation.
If the non-tech-respecting version were necessary in the future, one can either use Trait_FreePromotions, which I believe inserts as if they were Unit_FreePromotions ("the xml" as the comments call it), or just define a new promotion without the restriction that only that civ uses and assign that instead.

Questions for the reviewer:
1) can I compare to NO_TECH with equality like that? (I saw != elsewhere but maybe that's slightly different) 
2) Is there anywhere else this code block needs to go? This is for unit initialization, where does it check if the situation has changed?

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvTeam.cpp
+++ b/CvGameCoreDLL_Expansion2/CvTeam.cpp
@@ -7926,7 +7926,7 @@ void CvTeam::processTech(TechTypes eTech, int iChange, bool bNoBonus)
 					for (CvUnit* pLoopUnit = kPlayer.firstUnit(&iLoop); pLoopUnit != NULL; pLoopUnit = kPlayer.nextUnit(&iLoop))
 					{
 						// If we're in friendly territory and we can embark, give the promotion for free
-						if (pLoopUnit->plot()->IsFriendlyTerritory(*it))
+						if (pLoopUnit->plot()->IsFriendlyTerritory((PlayerTypes)iI)) 
 						{
 							// Civilian unit or the unit can acquire this promotion
 							if (IsPromotionValidForUnitCombatType(ePromotion, pLoopUnit->getUnitType()) || IsPromotionValidForCivilianUnitType(ePromotion, pLoopUnit->getUnitType()))


CvGameCoreDLL_Expansion2/CvUnit.cpp
```diff
@@ -836,7 +836,12 @@ void CvUnit::initWithNameOffset(int iID, UnitTypes eUnit, int iNameOffset, UnitA
 			const PromotionTypes promotionID = (PromotionTypes)iJ;
 			if(kPlayer.GetPlayerTraits()->HasFreePromotionUnitCombat(promotionID, unitCombatType))
 			{
-				setHasPromotion(promotionID, true);
+				TechTypes ePrereqTech = static_cast<TechTypes>(GC.getPromotionInfo(ePromotion)->GetTechPrereq());
+				CvPlayer& kOwner = GET_PLAYER(getOwner());
```
azum4roll: There's already `CvPlayer& kPlayer = GET_PLAYER(getOwner());` up above.

CvGameCoreDLL_Expansion2/CvUnit.cpp
```diff
@@ -836,7 +836,12 @@ void CvUnit::initWithNameOffset(int iID, UnitTypes eUnit, int iNameOffset, UnitA
 			const PromotionTypes promotionID = (PromotionTypes)iJ;
 			if(kPlayer.GetPlayerTraits()->HasFreePromotionUnitCombat(promotionID, unitCombatType))
 			{
-				setHasPromotion(promotionID, true);
+				TechTypes ePrereqTech = static_cast<TechTypes>(GC.getPromotionInfo(ePromotion)->GetTechPrereq());
```
azum4roll: `ePromotion` doesn't exist.

azum4roll:
> 1. can I compare to NO_TECH with equality like that? (I saw != elsewhere but maybe that's slightly different)

`m_iTechPrereq` defaults to `NO_TECH`, so yes.

> 2. Is there anywhere else this code block needs to go? This is for unit initialization, where does it check if the situation has changed?

If existing units are supposed to gain the promotions when the tech is researched, do something in `CvTeam::setHasTech()`.

CvGameCoreDLL_Expansion2/CvUnit.cpp
Hokath: good catch

Hokath:
Had a go at CvTeam::setHasTech() addition

CvGameCoreDLL_Expansion2/CvTeam.cpp
```diff
@@ -7951,8 +7951,79 @@ void CvTeam::processTech(TechTypes eTech, int iChange, bool bNoBonus)
 						kPlayer.SetWorkersIgnoreImpassable(true);
 					}
 				}
+				// What about if the promotion comes through Trait_FreePromotionUnitCombats?
+				if (GC.getPromotionInfo(ePromotion)->GetTechPrereq() == pTech->GetID())
+				{
+				    CivsList veMembers = getPlayers();
```
schnetziomi5: This entire code already sits in a for loop that iterates over all players, and skips them if they aren't in the team, so running another loop here _should_ be unnecessary. The code above, not added by you, also does this for some reason though.

CvGameCoreDLL_Expansion2/CvTeam.cpp
```diff
@@ -7951,8 +7951,79 @@ void CvTeam::processTech(TechTypes eTech, int iChange, bool bNoBonus)
...
+				    for (CivsList::iterator it = veMembers.begin(); it != veMembers.end(); ++it)
+				    {
+				        CvPlayer& kPlayer = GET_PLAYER(*it);
+				        if (!kPlayer.isAlive())
+				            continue;
```
schnetziomi5: This entire code already sits in a for loop that iterates over all players, and skips them if they aren't in the team, so running another loop here _should_ be unnecessary. The code above, not added by you, also does this for some reason though.

CvGameCoreDLL_Expansion2/CvTeam.cpp
```diff
@@ -7951,8 +7951,79 @@ void CvTeam::processTech(TechTypes eTech, int iChange, bool bNoBonus)
...
+						}
+					}
+
+					CvPromotionEntry* pkPromotionInfo = GC.getPromotionInfo(ePromotion);
+					if (pkPromotionInfo && pkPromotionInfo->IsCanMoveImpassable())
+					{
+						kPlayer.SetWorkersIgnoreImpassable(true);
+					}
+				}
+			}
```
schnetziomi5: This whole loop is duplicated ... I wildly guess you copied it, inserted your code, and forgot to delete the original?

Hokath: must be? thanks!

CvGameCoreDLL_Expansion2/CvTeam.cpp
Hokath: yeah it's a bit weird, isn't it
kPlayer is already defined 
so 7925--7931 could just be cut and the block unindented (and again below)

Hokath:
@schnetziomi5 does that look right to you?

CvGameCoreDLL_Expansion2/CvTeam.cpp
```diff
@@ -7922,25 +7922,16 @@ void CvTeam::processTech(TechTypes eTech, int iChange, bool bNoBonus)
...
+					for (CvUnit* pLoopUnit = kPlayer.firstUnit(&iLoop); pLoopUnit != NULL; pLoopUnit = kPlayer.nextUnit(&iLoop))
 					{
-						CvPlayer& kPlayer = GET_PLAYER(*it);
-						if (!kPlayer.isAlive())
-							continue;
-
-						int iLoop = 0;
-						for (CvUnit* pLoopUnit = kPlayer.firstUnit(&iLoop); pLoopUnit != NULL; pLoopUnit = kPlayer.nextUnit(&iLoop))
+						// If we're in friendly territory and we can embark, give the promotion for free
+						if (pLoopUnit->plot()->IsFriendlyTerritory(*it))
```
schnetziomi5: `if (pLoopUnit->plot()->IsFriendlyTerritory((PlayerTypes)iI))` to make it compile

Looks good to me otherwise :)

Hokath:
I built the DLL using Github Actions (amazing)
and it all seems to work as intended.

Big thanks to the two reviewers

Issue #12346, opened 2025-12-28
Field Gun uses two different icons in Strategic view


paulSoryu:
### 1. Mod version

5.0.1

### 2. Installed Components

Vox Populi with EUI

### 4. Describe the Issue

When camera is zoomed out and city badge changes to a minimalistic one, Field Gun has a different icon from the one when camera is zoomed in.
<img width="440" height="385" alt="Image" src="https://github.com/user-attachments/assets/36ee7de7-8d21-4138-a716-9281cb9fd498" />
<img width="669" height="593" alt="Image" src="https://github.com/user-attachments/assets/1f8efde5-5321-451e-81c6-18109b3b872c" />

It also uses a different icon outside of city, when camera is zoomed in and "secondary" unit icons appear in the centre of a tile (I assume to be able to view unit's promotions)
<img width="333" height="285" alt="Image" src="https://github.com/user-attachments/assets/ac8bd3b1-2886-4c16-8501-9cb0ec285eee" />

Hokath:
I suspect that the strategic view icon is still the old one, which only loads in at a given zoom level (so they overlap)

The little icon below, I have no idea. It has promo flags so maybe to do with that? What happens if you turn promo flags off?

Hokath:
Yes, svfieldgun.dds is the old flag

Do we know where this new icon came from?
The model is a smitty one, I found that, but the icon is sourced separately and I don't a reference anywhere atm.

paulSoryu:
Yes, these icons in the center appeared a long time ago, either when I started using Promo Flags when it was a modmod, or when it got integrated into VP, can't remember. I don't know if they are intended.

1) With Promo Flags disabled, icons in the middle don't appear at all. If it isn't possible to display promos above usual icons, then I don't mind if these secondary icons were removed completely. They are not terribly useful, some units don't have them at all, and they pollute the screen a lot.
2) If I enable Promo Flags (from the main menu), but don't reload the game (as it tells me to do), secondary icons in strategic view don't appear at all, but in normal view they work as intended.

paulSoryu:
So I think the second part of this issue will be fixed as well by changing the icon. I created a separate issue for the "secondary" icons in SV in general.

axatin:
Unfortunately we have very little control over the Strategic View UI because we don't have access to most of its code. The secondary unit icons are apparently unintended, but I don't know what makes them appear and I don't see a way to show them consistently.

[...]

PR #12352, opened 2025-12-29
Change SV asset for field gun unit to parrott cannon

Hokath:
matching the new icon from (6-08)
ty to Jarcast & Loupgarou for helping find* the asset


Hokath:
fixes the first part of #12346

PR #12356, opened 2025-12-29
Fix Notifications table

vaderkos:
Fix an issue where the .Urgent column was defined as `bool` and contained mixed representations `('false', 'true', 0, 1)`. Civ engine treats bool columns as mixed-type fields, which leads to incorrect serialization/deserialization.
The column is migrated to boolean, and all existing data is normalized to integers to ensure consistent behavior.


azum4roll:
I wonder if this really needs to be fixed if the column is unused.

vaderkos:
> I wonder if this really needs to be fixed if the column is unused.

To make it consistent with the other tables.

PR #12357, opened 2025-12-30
Fix Building_ThemingBonuses table

vaderkos:
Fix an issue where the column was defined as bool. Civ engine treats bool columns as mixed-type fields, which leads to incorrect serialization/deserialization.
The column is migrated to boolean, and all existing data is normalized to integers to ensure consistent behavior.


azum4roll:
I don't see a problem with this table currently. The existing values (in CP) are 1 or NULL, never 'true' or 'false'.

There will be an issue if we ever handle this table in Lua (e.g. automating theming tooltips), but that's for later.

vaderkos:
To make it consistent with the other tables.

PR #12384, opened 2026-01-04
Modify population consumption logic for (non)settlers

Hokath:
This code is intended to reduce pop when making settler units, but actually it reduces pop when making any unit that converts food to production (which in some mods can be combat units).

The new check, copied from code above, tightens it to require the bSettler is also satisfied.

I think this use of `CvUnitEntry* pkUnitInfo = pUnit->getUnitInfo();` is correct? Reduces the number of calls.


axatin:
use `kUnitInfo.IsFound()`, not  `kUnitInfo->IsFound()`

azum4roll:
This didn't change the tooltip...

axatin:
> This didn't change the tooltip...

fixed

PR #12386, opened 2026-01-04
Attempt at New policy table: Policy_YieldFromUnitProduction

Hokath:
I've tried to follow Axatin's guide to add a table based on the existing Building_YieldFromUnitProduction

It's mostly there but I'm stuck atm trying to find where the City effects are updated when Policies are gained.
Right now I was trying to mimic the 
`// Free Culture-per-turn in every City from Policies
ChangeBaseYieldRateFromPolicies(YIELD_CULTURE, GC.getPolicyInfo(ePolicy)->GetCulturePerCity());`
but I'm not seeing where else it gets called atm

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlayer.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlayer.cpp
@@ -42260,6 +42260,19 @@ void CvPlayer::processPolicies(PolicyTypes ePolicy, int iChange)
 			}
 		}
 		pLoopCity->ChangeBaseYieldRateFromPolicies(YIELD_CULTURE, iCityCultureChange);
+		// Yield from training units
+		for (int iI = 0; iI < NUM_YIELD_TYPES; iI++)
+		{
+			YieldTypes eYield = (YieldTypes)iI;
+			if (eYield == NO_YIELD)
+				continue;
+
+			//Simplification - errata yields not worth considering.
+			if ((YieldTypes)iI > YIELD_CULTURE_LOCAL && !MOD_BALANCE_CORE_JFD)
+				break;
+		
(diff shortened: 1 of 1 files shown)


Hokath:
Azum pointed me to CvPlayer::processPolicies

axatin:
You're also doing dll modding now? Nice!

This looks fine to me

Hokath:
I'm not a c++ person, but with some feedback I can hopefully reduce the workload overall ;)

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -42285,6 +42260,19 @@ void CvPlayer::processPolicies(PolicyTypes ePolicy, int iChange)
 			}
 		}
 		pLoopCity->ChangeBaseYieldRateFromPolicies(YIELD_CULTURE, iCityCultureChange);
+		// Yield from training units
+		for (int iI = 0; iI < NUM_YIELD_TYPES; iI++)
```
azum4roll, line 42264: Next time you can use `GC.getNUM_YIELD_TYPES()` instead of `NUM_YIELD_TYPES` and skip the simplification block.

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -42285,6 +42260,19 @@ void CvPlayer::processPolicies(PolicyTypes ePolicy, int iChange)
...
+		{
+			YieldTypes eYield = (YieldTypes)iI;
+			if (eYield == NO_YIELD)
```
azum4roll, line 42267: This cannot happen (NO_YIELD is -1) so you can omit this.

PR #12387, opened 2026-01-04
New Accomplishment tracking wars won

Hokath:
There's a proposal that wants this, and I think it would be good to have for modmods so I'm giving it a go


Hokath:
This works in the game 
(I swapped the accomplishment for Fornix and then delivered the beat-down to my neighbor with an army of IGE lancers until they gave up 80+ score)

Can someone check it isn't missing some kind of caching/serialization? I figured adding it to this enum is enough but maybe not?

PR #12392, opened 2026-01-05
Optimize FLuaVector functions and loading

vaderkos:
Avoid repeated loading of `FLuaVector` by caching in `MapModData`, ensuring vector helpers are initialized only once per game session.

`include('FLuaVector')` was being executed many times during game load, recreating the same vector functions repeatedly and generating unnecessary garbage. In a typical load, this happened at least **~118** times, despite the module being logically static.

The optimized non-EUI implementation (identical to EUI) is measurably faster in microbenchmarks:
- old: `~0.50 µs` per VecAdd
- new: `~0.35 µs` per VecAdd

This difference is small and should not be user-visible; the primary benefit is reduced garbage creation and cleaner initialization behavior.

### DIFF
--- a/(3a) EUI Compatibility Files/VP EUI Compatibility.civ5proj
+++ b/(3a) EUI Compatibility Files/VP EUI Compatibility.civ5proj
@@ -167,6 +167,10 @@
       <SubType>Lua</SubType>
       <ImportIntoVFS>True</ImportIntoVFS>
     </Content>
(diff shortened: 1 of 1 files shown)


axatin:
Also need to add the new file in the civproj

vaderkos:
> Also need to add the new file in the civproj

done

PR #12408, opened 2026-01-09
5/6UC Compatible UI -- needs flexible switching

Hokath:
Here are the override versions of UI files needed to display 5/6 UC.
I merged in the 5/6 UC overrides into the CivilopediaScreen.lua as it has been updated since I made the first mod release.

Currently does 2 things:
1) Adds queries to get unique improvements that don't depend on CivilizationType. The 5/6UC improvements are added using Trait_UnitClassBuilds, and the (immaterial?) OrderPriority value is set in Builds to the unused value of 90 to separate them from the table's other entries. This is a bit of a hack to get civs to share a UI.
2) This hard-switches the UI to 6 components rather than 4, including changing the .xml file to get different spacings for icons and text blocks.

The second part needs to become flexible so that it only activates if 5/6 UC is active otherwise there will be empty circles in the UI (empty=Pyramids icon, lol). So cannot merge yet.

### DIFF
--- a/(2) Vox Populi/Core Files/Overrides/UniqueBonuses.lua
+++ b/(2) Vox Populi/Core Files/Overrides/UniqueBonuses.lua
@@ -145,7 +145,7 @@ end
 ------------------------------------------------------------------
 function PopulateUniqueBonuses( controlTable, civ, _, extendedTooltip, noTooltip)
 
-	local maxSmallButtons = 4;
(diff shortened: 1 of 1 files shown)


axatin:
Fixed the remaining UI issues here: https://github.com/LoneGazebo/Community-Patch-DLL/pull/12421

PR #12414, opened 2026-01-10
Unit promotions yield from scouting x100

Hokath:
Integer entries in this table are very hard to balance, so it would be better to have x100 format for like 1 yield per X tiles. 
(still can get rounded down, no fractional instant yield).

I didn't rename any tables, do you think it should have "Times100" in the name as well?


axatin:
Yeah better to make it consistent and add Times100, also in the DLL

Hokath:
These were all the instances I could find
surprisingly never gets called in tactical, movement, or mission unit AI so I was worried I missed somewhere?

Narrator: he missed somewhere

PR #12421, opened 2026-01-11
5/6 UI Compatible UI

axatin:


axatin:
@Hokath Change NUM_UNIQUE_COMPONENTS to 6 in your mod

azum4roll:
Can't we use stacks instead of having a fixed number of icons in the xml?

axatin:
That's what I tried first, but I wasn't able to get the layout right, it's difficult with those two rows of items.

azum4roll:
Should be doable with 2 stacks, one with odd indices and one with even.

PR #12423, opened 2026-01-12
fix: missing SAFE_DELETE_ARRAY calls

qingyin3545:
I reviewed **CvUnitEntry**, **CvPromotionEntry**, **CvPolicyEntry**, **CvBuildingEntry**, **CvTraitEntry**, and **CvBeliefEntry**, and added missing SAFE_DELETE_ARRAY calls for some 1D array pointers.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvBeliefClasses.cpp
+++ b/CvGameCoreDLL_Expansion2/CvBeliefClasses.cpp
@@ -179,6 +179,71 @@ CvBeliefEntry::CvBeliefEntry() :
 /// Destructor
 CvBeliefEntry::~CvBeliefEntry()
 {
+	SAFE_DELETE_ARRAY(m_paiCityYieldChange);
+	SAFE_DELETE_ARRAY(m_paiHolyCityYieldChange);
+	SAFE_DELETE_ARRAY(m_paiYieldChangePerForeignCity);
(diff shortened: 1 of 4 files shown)


Hokath:
Nice.
Could any of these have led to memory leaks or are they only called once in a standard run?

qingyin3545:
> Nice. Could any of these have led to memory leaks or are they only called once in a standard run?

In debug testing, breakpoints show that the destructors are triggered each time “Exit to Main Menu” is clicked, but in normal gameplay this is rarely done, so it usually does not cause noticeable issues.

qingyin3545:
> > Nice. Could any of these have led to memory leaks or are they only called once in a standard run?
> 
> In debug testing, breakpoints show that the destructors are triggered each time “Exit to Main Menu” is clicked, but in normal gameplay this is rarely done, so it usually does not cause noticeable issues.

image

Hokath:
That's interesting. In testing you sometimes cycle Exit To Main Menu -> Mods load -> etc.
and I've noticed after a lot of these it eventually CTDs just like an OOM crash. Maybe this was the reason!

azum4roll:
Mod loader's a mess, like when mods are unloaded, they attempt to load vanilla maps again *using modded Lua files*.

PR #12428, opened 2026-01-12
SeeThrough for canEverRangeStrikeAt->canSeePlot usage

Hokath:
An attempt to make a promotion column that elevates (or, I guess, declines) the level of the unit that has it when they do a ranged attack, allowing them to shoot over *some* obstacles.

Most advanced thing attempted so far.
Will now build the DLL and see if it works.


Hokath:
Need to test fix to Bombardment.lua, but looks like it is working
image

Hokath:
If you were on a hill, you could shoot over the 1-taller mountain tile.
So to stop this, the mountain needs to be height level 3 instead of 2.
I didn't change the SeeFrom of the mountain to 3 because that would lead to giant vision for scouts. This seems fine.
Also updated natural wonders that are mountains to also be 3.

Hokath:
Can I ask the reviewer if they can see in the military AI anywhere the use of canSeePlot instead of canEverRangeStrikeAt that might need changing?

PR #12442, opened 2026-01-14
CS Friend and Ally tables now x100 for future balancing

Hokath:
> Changing the table to Times100 is very simple, it just needs to be renamed and a "* 100" needs to be removed in two or three lines in the DLL
image

Nah it was fine. First time really reading the new automatic tooltips functions. I think I understood that these *100 tables are done in this way?

Hopefully I got all instances?

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvCityStrategyAI.cpp
+++ b/CvGameCoreDLL_Expansion2/CvCityStrategyAI.cpp
@@ -3786,11 +3786,11 @@ int CityStrategyAIHelpers::GetBuildingYieldValue(CvCity *pCity, BuildingTypes eB
 	}
 	if (pkBuildingInfo->GetYieldPerAllyTimes100(eYield) > 0)
 	{
-		iFlatYield += (pkBuildingInfo->GetYieldPerAllyTimes100(eYield) * max(GC.getGame().GetNumMinorCivsAlive() / 400, kPlayer.GetNumCSAllies()));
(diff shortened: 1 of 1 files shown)


CvGameCoreDLL_Expansion2/CvCityStrategyAI.cpp
```diff
@@ -3784,13 +3784,13 @@ int CityStrategyAIHelpers::GetBuildingYieldValue(CvCity *pCity, BuildingTypes eB
 			}
 		}
 	}
-	if (pkBuildingInfo->GetYieldPerAlly(eYield) > 0)
+	if (pkBuildingInfo->GetYieldPerAllyTimes100(eYield) > 0)
 	{
-		iFlatYield += (pkBuildingInfo->GetYieldPerAlly(eYield) * max(GC.getGame().GetNumMinorCivsAlive() / 4, kPlayer.GetNumCSAllies()));
+		iFlatYield += (pkBuildingInfo->GetYieldPerAllyTimes100(eYield) * max(GC.getGame().GetNumMinorCivsAlive() / 400, kPlayer.GetNumCSAllies()));
```
axatin: ` GC.getGame().GetNumMinorCivsAlive() / 400 ` will always be rounded down to zero,  use
`pkBuildingInfo->GetYieldPerAllyTimes100(eYield) * max(GC.getGame().GetNumMinorCivsAlive() / 4, kPlayer.GetNumCSAllies()) / 100` instead

azum4roll:
If the number of times100 tables keeps increasing, we'll be needing an ExtractSimpleYieldTableTimes100 function soon.

PR #12443, opened 2026-01-15
Fix Tersane Description

ron-murhammer:
Fix Tersane Description


Hokath:
Thanks.

I noticed there's no Great Admiral Points in the original proposals, even though there was in the 4UC mod.
It might have been an oversight to drop them because there's no mention of removing it in the proposals.

azum4roll:
There's also no mention of it in component description in pdan's original proposal.

azum4roll:
It's also not "to the adjacent city", but "to the city owning this tile".

PR #12460, opened 2026-01-17
Fix Type column constraints for several tables.

vaderkos:
Add `Database Migrations` directory.
Move db indexes to migrations.


RecursiveVision:
1) Why date the file names?
2) Why the separate subfolder instead of adding it somewhere natural in the database?

vaderkos:
> 1. Why date the file names?
> 
>     2. Why the separate subfolder instead of adding it somewhere natural in the database?

@RecursiveVision 
1. Because migrations should run in a strict, deterministic order. A dateprefix guarantees correct ordering or at least implies it in our case. It also makes it trivial to know which migrations are newer or already applied, without having to look into proj files or metadata.
2. I couldn't find a natural place for it. And as such changes are rather timeline..

And date-prefixed migration files are standard across many ecosystems.

RecursiveVision:
I dislike the stylistic clash with the rest of the mod's files. Not enough to do anything beyond inquire, though. :P

vaderkos:
> I dislike the stylistic clash with the rest of the mod's files. Not enough to do anything beyond inquire, though. :P

¯\_(ツ)_/¯

vaderkos:
@axatin Could you merge it or there are some arguments against?

PR #12463, opened 2026-01-18
Extend unused Incan UA column to Mountains in VP

Hokath:
I'd like to propose a new Inca next time and it involves reverting the UA somewhat. Since we can now do roads on Mountains, the UA should cover that (was just Hills) when MOD_BALANCE_VP is active.


RecursiveVision:
Not sure if it's appropriate here but note the ALTERNATE_CIV_TRAIT CustomModOptions may be of some use to you.

PR #12478, opened 2026-01-20
(9-69a) Incan UA Movement Nerf (River Only)

RecursiveVision:
https://forums.civfanatics.com/threads/9-69a-incan-ua-movement-nerf-river-only.701731

Also decided to decouple the Iroquois's movement change in VP from the general BALANCE_VP option while I was at it.


RecursiveVision:
Please rebase, do not squash.

Issue #12172, opened 2025-11-05
Denounce


thecanidcoder:
### 1. Mod version

5.0-alpha.03

### 2. Installed Components

Vox Populi with EUI

### 3. Describe the Issue

Denounce

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/c2427163-1e0a-4256-b5af-3bb1116108c0" />

### 4. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[Denounce.zip](https://github.com/user-attachments/files/23372382/Denounce.zip)

thecanidcoder:
Logs won't upload; don't imagine they're needed for this one, but if they are, let me know and I'll get it uploaded to MEGA or similar.

Vordikk:
This is actually all what u need in denouncing. Really doesn't matter, what exactly they disliked in your actions. =)

PR #12484, opened 2026-01-20
Fixes

axatin:


(1) Community Patch/Database Changes/Text/en_US/UI/CoreNewUIText.xml
```diff
@@ -2314,6 +2314,10 @@
 			<Text>Schedule a forced resync</Text>
 		</Row>
 
+		<Row Tag="TXT_KEY_CBP_DENOUNCE_HUMAN">
```
azum4roll: Just use TXT_KEY_DENOUNCE_HUMAN instead of this?

axatin: you're right, the text exists already

axatin: Wait a minute, I remember now that there was a problem with that text key. It starts with "TXT_KEY_DENOUNCE", which is used as a generic text key in Diplomacy_Responses, so it ended up being shown in the diplo messages: https://github.com/LoneGazebo/Community-Patch-DLL/issues/12172

I thought I'd fixed that. Seems something went wrong

axatin: Fixed.

PR #12485, opened 2026-01-20
Various CPK utilities

vaderkos:
**This PR doesn't do anything on its own**
Small set of `CPK` utility helpers, focused on improving diagnostics, formatting, and performance measurement.

1. `CPK.FP.Void` - A no-op function useful as a safe default callback or placeholder.
2. `CPK.String.Escape` - Escapes strings for safe output / diagnostics.
3. `CPK.String.PadStart` / `CPK.String.PadEnd` - Simple string padding helpers used for table formatting and aligned output.
4. `CPK.Util.Format.ByteSize` - Formats byte counts into human-readable units (`B`, `KiB`, `MiB`, `GiB`).
5. `CPK.Util.Format.Duration` - Formats durations (in seconds) into human-readable time units (`ns`, `us`, `ms`, `s`).
6. Adds explicit `traceback` types to `Global.d.lua`.
7. Adjusted assertion severity levels in `CPK.Assert.Error`
8. Added `CPK.Util.Benchmark` - utility to measure and compare time and memory in various case, more information below

For example:
```lua
local my_bench = Benchmark {
	print = true,
	title = 'My Benchmark',
	times = 1000,
	loops = 10,
	cases = {
		{ name = '1K', make = function() return 1024 end },
		{ name = '4K', make = function() return 4096 end },
	},
	tests = {
		{
			name = 'table.fill(0, N)',
			make = function(N)
				local fill = table.fill
				return function()
					return fill(0, N)
				end
			end
		},
		{
			name = 'for i = 1, N do arr[i] = 0 end',
			make = function(N)
				return function()
					local arr = {}
					for i = 1, N do arr[i] = 0 end
					return arr
				end
			end
		}
	}
}

my_bench()
```

Will print this if `print = true`, and table with statistics.
image

### DIFF
--- a/(1) Community Patch/Kit/String/CPK.String.PadEnd.lua
+++ b/(1) Community Patch/Kit/String/CPK.String.PadEnd.lua
@@ -1,9 +1,15 @@
 local lua_string_rep = string.rep
 
+local AssertIsString = CPK.Assert.IsString
+local AssertIsNumber = CPK.Assert.IsNumber
+
 --- @param str string
 --- @param len number
 --- @param pad? string
 function CPK.String.PadEnd(str, len, pad)
+	AssertIsString(str)
+	AssertIsNumber(len)
+
 	if #str >= len then return str end
 
 	if pad == nil then pad = ' ' end
--- a/(1) Community Patch/Kit/String/CPK.String.PadStart.lua
+++ b/(1) Community Patch/Kit/String/CPK.String.PadStart.lua
@@ -1,10 +1,17 @@
 local lua_string_rep = string.rep
 
+local AssertIsString = CPK.Assert.IsString
+local AssertIsNumber = CPK.Assert.IsNumber
+
 --- @param str string
 --- @param len number
 --- @param pad? string
 function CPK.String.PadStart(str, len, pad)
+	AssertIsString(str)
+	AssertIsNumber(len)
+
 	if #str >= len then return str end
+
 	if pad == nil then pad = ' ' end
 
 	return lua_string_rep(pad, len - #str) .. str
--- a/(1) Community Patch/Kit/Util/CPK.Util.Benchmark.lua
+++ b/(1) Community Patch/Kit/Util/CPK.Util.Benchmark.lua
@@ -21,6 +21,8 @@ local AssertError = CPK.Assert.Error
 local AssertIsTable = CPK.Assert.IsTable
 local AssertIsNumber = CPK.Assert.IsNumber
 
+local TableCopy = CPK.Table.Copy
+
 local StringPadStart = CPK.String.PadStart
 
 local FormatByteSize = CPK.Util.Format.ByteSize
@@ -43,7 +45,7 @@ local gc_count = function()
 end
 
 local gc_run = function(n)
-	n = n or 3
+	n = n or 1
 	for _ = 1, n do
 		lua_collectgarbage('collect')
 	end
@@ -75,6 +77,9 @@ local function stats(arr)
 			min = v,
 		}
 	end
+
+	arr = TableCopy(arr)
+	lua_table_sort(arr)
 
 	local sum = 0
 	for i = 1, len do
@@ -142,9 +147,6 @@ local benchmark = function(times, loops, fn)
 
 	gc_start()
 	live_after = gc_run()
-
-	lua_table_sort(durs)
-	lua_table_sort(mems)
 
 	return {
 		live = live_after - live_prior,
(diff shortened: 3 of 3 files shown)


(1) Community Patch/Kit/String/CPK.String.PadEnd.lua
```diff
@@ -0,0 +1,12 @@
...
+--- @param str string
+--- @param len number
+--- @param pad? string
+function CPK.String.PadEnd(str, len, pad)
+	if #str >= len then return str end
+
+	if pad == nil then pad = ' ' end
+
+	return str .. lua_string_rep(pad, len - #str)
+end
```
schnetziomi5, line 18: Generally CPK is designed to be strict when it comes to checks, right? So I feel `AssertIsString(str)` is missing here. As is the function might return without throwing if a table is passed.

(1) Community Patch/Kit/String/CPK.String.PadStart.lua
```diff
@@ -0,0 +1,11 @@
...
+
+--- @param str string
+--- @param len number
+--- @param pad? string
+function CPK.String.PadStart(str, len, pad)
+	if #str >= len then return str end
+	if pad == nil then pad = ' ' end
+
+	return lua_string_rep(pad, len - #str) .. str
+end
```
schnetziomi5, line 18: Same as `PadEnd`

(1) Community Patch/Kit/Util/CPK.Util.Benchmark.lua
```diff
@@ -0,0 +1,405 @@
...
+local gc_pause = function()
+	lua_collectgarbage('stop')
+end
+
+local gc_count = function()
+	return 1024 * lua_collectgarbage('count')
+end
+
+local gc_run = function(n)
+	n = n or 3
```
schnetziomi5: Any good reason for `n=3` by default? I know I did so in some files I shared with you, but I came to think this might have been bad practice. Unless userdata finalizers or some kind of linked weak table constructions(?) are used somewhere in the tested code one time should be totally sufficient.

That said at worst it just takes a lot of time...

(1) Community Patch/Kit/Util/CPK.Util.Benchmark.lua
```diff
@@ -0,0 +1,405 @@
...
+
+		durs[i] = (time_after - time_prior) / loops
+		mems[i] = (heap_after - heap_prior) / loops
+	end
+
+	gc_start()
+	live_after = gc_run()
+
+	lua_table_sort(durs)
+	lua_table_sort(mems)
```
schnetziomi5: I strongly think this is _wrong_ for two reasons:

First if anyone wanted to do further statistical analysis on the results durs[x] and mems[x] should be from the x'th run. If you sort here there is no way to reconstruct that.

Secondly I would argue that any sorting operation should naturally be placed in the `stats` function, where the statistical post processing takes place.

(1) Community Patch/Kit/Util/CPK.Util.Benchmark.lua
```diff
@@ -0,0 +1,405 @@
...
+			if last_case_name ~= next_case_name then
+				psb:Append(sep)
+				last_case_name = next_case_name
+			end
+
+			psb:Append(sub:Concat(' | '))
+			sub:Clear()
+		end
+
+		lua_print(psb:Append(sep):Append(''):Concat('\n'))
```
schnetziomi5, line 402: You should print a separate single line here about the fact that the actual output will not show up in the Firetuner, and that it can be found in Lua.log ...

schnetziomi5 (commented):
(Didn't test anything, but read twice, focused on understanding the code on first pass, adding comments during the second...)

(1) Community Patch/Kit/Util/CPK.Util.Benchmark.lua
vaderkos: Fixed

(1) Community Patch/Kit/Util/CPK.Util.Benchmark.lua
vaderkos: Fixed

(1) Community Patch/Kit/Util/CPK.Util.Benchmark.lua
vaderkos: Fixed, default 1

(1) Community Patch/Kit/String/CPK.String.PadStart.lua
vaderkos: Fixed

(1) Community Patch/Kit/String/CPK.String.PadEnd.lua
vaderkos: Fixed

vaderkos:
@axatin Could you merge it or there are some arguments against?

PR #12506, opened 2026-01-23
Backend changes to allow new Great Admiral GPTI

Hokath:
The "Voyage of Discovery" function lives in CvUnit and part of the logic is to identify which resource to pick. But really it is independent of the unit that uses it, and depends only on the map and player. Therefore I've moved it (GetFreeLuxury) to CvPlayer and refactored CvUnit to not have duplicated code.

Then, I've added an improvement boolean column (ExoticResourceFromImprovement) that, when set to 1, will allow for an improvement to use the Voyage of Discovery effect to place a resource, similar to ImprovementResource. It also hooks up to ImprovementResourceQuantity and the policy effect that boosts Great Admiral mission. This will allow us to make Voyage of Discovery give monopolies via a GPTI.

Should have no effects on base VP.


azum4roll:
Please learn to use [rebase](https://hackernoon.com/beginners-guide-to-interactive-rebasing-346a3f9c3a6d) so you don't have a gazillion commits (and you can also edit your commit messages there).

Note that force pushing is always safe on your own branch that nobody else has access to.

----

Will the resource disappear if you remove the GPTI?

Hokath:
Unfortunately I haven't found a way to rebase and/or force push on the github online interface. Do you know if there is a way?

It's not intended to be able to remove the Great Admiral resources so atm I didn't code for the resources to disappear. At the moment there should be no way to replace this improvement -- I intend to set the permanent flag just in case.

azum4roll:
Permanent also disables pillage.

RecursiveVision:
> Unfortunately I haven't found a way to rebase and/or force push on the github online interface. Do you know if there is a way?

Use GitHub Desktop.

Hokath:
I'm trying to avoid downloading the DLL onto my tablet.
It seems there's a thing you can activate called a "codespace" that might work

Hokath:
Yes, it gives a cloud environment that allows terminal commands!

azum4roll:
> It also seems that improvements that generate resources actually remove and place them when pillaged/unpillaged.

Does brazilwood camp work like that? Or villa?

Hokath:
I tested and no, that code doesn't seem to get executed for Brazilwood Camp. You pillage it and the resource stays.
Looking again, it requires this comparison getResourceType() != eResourceFromImprovement
So the pillaged improvement somehow got a resource that wasn't the one it usually has? No idea when that would ever happen.
```
			CvImprovementEntry* pImprovementInfo = GC.getImprovementInfo(getImprovementType());

			int iChange = bPillaged ? -1 : 1;

			ResourceTypes eResourceFromImprovement = static_cast<ResourceTypes>(pImprovementInfo->GetResourceFromImprovement());
			if (eResourceFromImprovement != NO_RESOURCE && getResourceType() != NO_RESOURCE && getResourceType() != eResourceFromImprovement)
			{
				int iQuantity = pImprovementInfo->GetResourceQuantityFromImprovement();
				if (iQuantity <= 0)
				{
					iQuantity = 1;
				}
				GET_PLAYER(getOwner()).changeNumResourceTotal(eResourceFromImprovement, iChange * iQuantity, false, true, true);
			}
```

azum4roll:
There are a bunch of copy pasted code in that function and setImprovementType() and some basically do nothing or are never executed.

The condition from your code snippet is copied from setOwner(). I think this is just wrong?

Hokath:
Yeah maybe that code should just be deleted, idk.

Hokath:
I'll open a ticket so we can remember this SetImprovementPillaged block seems to be deprecated.

Otherwise I think the code can be merged?

PR #12529, opened 2026-01-31
Complete implementation of belief column FaithFromDyingUnits

Hokath:
It turns out this column was never implemented. 
It is required for a couple of proposals in the current congress.

It works the same way as faith from killing units: times 100 percentage of their strength, not scaling. Instant death yields go to the capital (unchanged).

image

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlayer.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlayer.cpp
@@ -26137,8 +26137,15 @@ void CvPlayer::doInstantYield(InstantYieldType iType, bool bCityFaith, GreatPers
 					if(pLoopCity->isCapital())
 					{
 						iValue += getYieldFromDeath(eYield);
-						
-						if (eYield == YIELD_FAITH && pReligion && !(GC.getGame().isOption(GAMEOPTION_NO_RELIGION)))
+					}
+					if (eYield == YIELD_FAITH && pReligion)
+					{
+						const ReligionTypes eReligion = GetReligions()->GetOwnedReligion();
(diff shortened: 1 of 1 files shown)


azum4roll:
You mean capital/holy city? This is a belief!

Depending on which belief type you want this to work on, you may also need to check for belief validity (go through CvReligionBeliefs::GetFaithFromDyingUnits)

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -26137,6 +26137,24 @@ void CvPlayer::doInstantYield(InstantYieldType iType, bool bCityFaith, GreatPers
 					if(pLoopCity->isCapital())
 					{
 						iValue += getYieldFromDeath(eYield);
+						
+						if (eYield == YIELD_FAITH && pReligion && !(GC.getGame().isOption(GAMEOPTION_NO_RELIGION)))
```
azum4roll: I don't think No Religion can be on when you have a religion?

Hokath:
Right, it should preferentially choose holy city, though there is no difference in effect atm since it's only for faith (right?).

I based this off the GetYieldFromKills version on L#25939
It does have a city argument, but just passes it as pHolyCity

I guess if you wanted this a follower belief you would need to have something like,
CvGameReligions::GetBeliefYieldForKill
which gets called after GetYieldFromKills in that other block.
But that function just seems to find if any (one) city has a belief that grants yields, and doesn't find the unit's home city or anything like that...
GetBeliefYieldForKill calls specifically GetFaithFromKills. This and GetYieldFromKills don't behave quite the same. The first one has this int iDistance argument I haven't seen used (?)

azum4roll:
Vanilla God of War has a max distance of 4 from a city with the belief.

PR #12530, opened 2026-01-31
5.1.3 Fixes

axatin:


CvGameCoreDLL_Expansion2/CvBuildingClasses.cpp
```diff
@@ -5816,6 +5816,28 @@ void CvCityBuildings::SetNumRealBuildingTimed(BuildingTypes eIndex, int iNewValu
 	{
 		m_pCity->SetBuildingInvestment(buildingClassType, false);
 		m_pCity->GetCityCitizens()->DoRemoveAllSpecialistsFromBuilding(eIndex, CvCity::YIELD_UPDATE_GLOBAL);
+		if (GetNumFreeBuilding(eIndex) == 0)
```
azum4roll, line 5819: Only counting free buildings?

axatin: don't remove the GWs if the building is still there as a free building

PR #12535, opened 2026-01-31
Add more Projects columns

Hokath:
3 things:

1. I'm using a Project in 5/6 UC and I thought it would be handy if the completion of a project could trigger an event.
This would give a way for the player to controllably be presented with the events interface, and, because projects can be repeatable, possibly over and over.
In my example I just need the one-shot on a player event, but I thought it was a cool functionality so I also added the City Event version.

2. I also wanted Military Supply and I thought that might be useful in the future, especially on a repeatable project.

3. Finally there is a new EventChoice column that changes the number of air slots on your Cities. I need this for the 5/6 UC particular event only. I also noticed the units should spawn after the city-level effects are processed, especially in this case where a lack of air slots means planes wouldnt spawn at all!


azum4roll:
Is it possible to fire multiple events/city events on the same turn? What if you want a project to trigger 2 events?

Hokath:
You could do a player and a city event together, I guess.
If you wanted two different events it might be easiest to link the second event to the first, so it fired the next turn.
I don't really see a use case for that atm, did you have something in mind?

azum4roll:
Nah, just spitballing.

PR #12536, opened 2026-01-31
Cultural project UI

Hokath:
We've had the 5/6 UC UI merged into the VP override files, thanks to @axatin 
That was necessary because of shared Improvements.
Now after #12535 we will have a shared Project, so another change is needed.

I thought because PolicyType is on the Project this would be easier than the improvement version, but actually no.
The Improvement version has database queries that only look at tables that exist in VP, whereas the Project version needs a way to refer PolicyType to CivilizationType, and this is done via a modmod table. Therefore all the queries would need to be guarded for the existence of this table (and column), something I couldn't see how to do very easily.

My alternative solution is to add a UI column to the Civilizations table and use that in the modmod. This way the VP UI functions don't need to get complicated when 5/6 UC isn't loaded, as the new column would just return null harmlessly.

Is this a sensible approach?

### DIFF
--- a/(1) Community Patch/Community Patch.civ5proj
+++ b/(1) Community Patch/Community Patch.civ5proj
@@ -187,6 +187,11 @@
       <Action>
         <Set>OnModActivated</Set>
         <Type>UpdateDatabase</Type>
+        <FileName>Database Changes\Civilizations\CivilizationTableChanges.sql</FileName>
+      </Action>
+      <Action>
+        <Set>OnModActivated</Set>
+        <Type>UpdateDatabase</Type>
         <FileName>Database Changes\Civilizations\NewTraitTables.xml</FileName>
       </Action>
       <Action>
@@ -3161,6 +3166,10 @@
       <ImportIntoVFS>False</ImportIntoVFS>
     </Content>
     <Content Include="Database Changes\Civilizations\NewCivilizationTables.xml">
+      <SubType>Lua</SubType>
+      <ImportIntoVFS>False</ImportIntoVFS>
+    </Content>
+    <Content Include="Database Changes\Civilizations\CivilizationTableChanges.sql">
       <SubType>Lua</SubType>
       <ImportIntoVFS>False</ImportIntoVFS>
     </Content>


(1) Community Patch/Database Changes/Civilizations/CivilizationChanges.sql
```diff
@@ -34,3 +34,6 @@ DELETE FROM Civilization_UnitClassOverrides
 WHERE CivilizationType = 'CIVILIZATION_BARBARIAN'
 AND UnitClassType = 'UNITCLASS_CHARIOT_ARCHER'
 AND UnitType IS NULL;
+
+-- Add a policy column that is used for UI purposes
```
azum4roll: CivilizationTableChanges.sql

Hokath: ah yeah. 
It doesn't exist yet. Shall I start it?

(2) Vox Populi/Core Files/Overrides/CivilopediaScreen.lua
```diff
@@ -5508,7 +5508,7 @@ CivilopediaCategory[CategoryCivilizations].SelectArticle = function( rawCivID, s
 				UpdateButtonFrame( buttonAdded, Controls.UniqueImprovementsInnerFrame, Controls.UniqueImprovementsFrame );
 
 				-- get unique improvements that don't use CivilizationRequired. Flag is to set OrderPriority = 90. Can combine with above block perhaps?
-                local CD_condition = "Type IN (SELECT ImprovementType FROM Builds WHERE OrderPriority = 90 AND Type IN (SELECT BuildType FROM Trait_BuildsUnitClasses WHERE TraitType IN (SELECT TraitType FROM Leader_Traits WHERE LeaderType IN (SELECT LeaderheadType FROM Civilization_Leaders WHERE CivilizationType = '" .. thisCiv.Type .. "'))))";
+                		local CD_condition = "Type IN (SELECT ImprovementType FROM Builds WHERE OrderPriority = 90 AND Type IN (SELECT BuildType FROM Trait_BuildsUnitClasses WHERE TraitType IN (SELECT TraitType FROM Leader_Traits WHERE LeaderType IN (SELECT LeaderheadType FROM Civilization_Leaders WHERE CivilizationType = '" .. thisCiv.Type .. "'))))";
```
vaderkos: That's very unoptimized and hard to read, what about

```lua

local CD_condition = [[
Type IN (
  SELECT b.ImprovementType
  FROM Civilization_Leaders cl
  JOIN Leader_Traits lt
    ON lt.LeaderType = cl.LeaderheadType
  JOIN Trait_BuildsUnitClasses tbu
    ON tbu.TraitType = lt.TraitType
  JOIN Builds b
    ON b.Type = tbu.BuildType
  WHERE cl.CivilizationType = ']] .. thisCiv.Type .. [['
    AND b.OrderPriority = 90
)
]]
```

(1) Community Patch/Database Changes/Civilizations/CivilizationChanges.sql
vaderkos: Why is it called `PolicyHack`?

(2) Vox Populi/Core Files/Overrides/CivilopediaScreen.lua
azum4roll: I didn't even know you could use GameInfo like that.

vaderkos: @azum4roll You can put any sql where clause.

under the hood it does
```SELECT * FROM Table WHERE %s``` `%s` is that actual sql you pass into function.

azum4roll: Does it support placeholders?

(1) Community Patch/Database Changes/Civilizations/CivilizationChanges.sql
Hokath: what's a better name? PolicyForUI?

(2) Vox Populi/Core Files/Overrides/CivilopediaScreen.lua
vaderkos: @azum4roll Do you mean string format `%s` or sql parameters like `?` or `:param`. Unfortunately both answer is NO. In my implementation of `Gi` I've added support for parameters, but I do not have time to finish it yet.

(1) Community Patch/Database Changes/Civilizations/CivilizationChanges.sql
vaderkos: @Hokath could you describe what this column does or represents? For me `PolicyForUI` sound like `Policy for User Interface` XD

Hokath: Correct!
It will allow the database query to connect the (dummy) policy to the cultural components (for projects in particular) and then display in the UI (civ select screen and pedia entry)
This needed to assign them because the "CultureType" (or "SubCultureType") that they work off is in a modmod table.

The alternative is for the modmod to override the VP UI files, but then it will become incompatible whenever they're updated

vaderkos: @Hokath So you say that some Civilizations have some custom projects (like public works?) that are unlockable by specific Policies?

Hokath: Yes, the problem is that multiple civilizations have the same project.
This works for units because there is a replacement table Civilization_UnitClassOverrides, and same for buildings. But for improvements and projects there isnt (as these werent designed to be replaced)

So I guess a long term solution if we are going to have unique improvements replace other improvements would be to have versions of these tables (right now we block the old improvement with Trait_NoBuilds + Improvement.CivilizationType)

PR #12541, opened 2026-02-01
Add new minor poland color

Hokath:
In @RecursiveVision's recent minor civ color update, he changed the full-red colors so they wouldn't look identical to Barbarian colors (Austria, Persia), but he missed Poland.
image
 
 Lightened to pale red -- chosen not to clash with the other updated colors (closest is probably england dusty rose, which is a more washed, grey color)
 
image

It was suggested here,
https://forums.civfanatics.com/threads/9-93-change-denmarks-minor-color-to-look-less-like-barbarians.701331
I think given the original logic of Recursive, which makes total sense, we don't need to vote on this.

### DIFF
--- a/(2) Vox Populi/Database Changes/NewColors.xml
+++ b/(2) Vox Populi/Database Changes/NewColors.xml
@@ -90,7 +90,15 @@
 			<Blue>0</Blue>
 			<Alpha>1</Alpha>
 		</Row>
-		<!-- Additionally, Austria and Persia's reds look too similar to Barbarians, and the same goes for brightened England, Mongolia and Morocco, so some manual adjustments were made. -->
+		<!-- Additionally, Austria, Poland, and Persia's reds look too similar to Barbarians, and the same goes for brightened England, Mongolia and Morocco, so some manual adjustments were made. -->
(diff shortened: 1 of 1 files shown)


azum4roll:
Can we not change Denmark's colour then? That looks nowhere close to barbarians.

vaderkos:
Wanted to ask if there is any logic in that colors? What about coloring city-states by their types? Or is this already happening?

Hokath:
The idea is color city states by the background color of the major civ they are blocked by.
Then some colors have to be lightened because you can't see them against the black background OR they are bright red and so look like barbarians.

> Can we not change Denmark's colour then? That looks nowhere close to barbarians.

Well I agree with you, but there is a vote. As I said, I guess some people have different vision

RecursiveVision:
I personally think that Poland also looks fairly different, being a bright red as opposed to the Barbarians' dull red. But I'm not opposed to this change.

azum4roll:
Barbarians only look dull red because of the hostile bold style from EUI. Ethiopia/Warsaw/Barbarians look super similar in non-EUI, or if they're all hostile.

PR #12555, opened 2026-02-05
ubsan fixes

JohnsterID:
Initial run with ubsan. When I have more time, I can look at more.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvCity.cpp
+++ b/CvGameCoreDLL_Expansion2/CvCity.cpp
@@ -7609,7 +7609,7 @@ int CvCity::maxXPValue() const
 		iMaxValue = std::min(iMaxValue, /*-1 in CP, 70 in VP*/ GD_INT_GET(MINOR_MAX_XP_VALUE));
 	}
 
-	if (iMaxValue > 0)
+	if (iMaxValue > 0 && iMaxValue != MAX_INT)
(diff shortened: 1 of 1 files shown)


CvGameCoreDLL_Expansion2/CvTacticalAI.cpp
```diff
@@ -2793,8 +2793,11 @@ void CvTacticalAI::SortTargetListAndDropUselessTargets()
 		{
 			for (vector<CvTacticalTarget>::iterator it2 = reducedTargetList.begin(); it2 != reducedTargetList.end(); ++it2)
 			{
-				//trick: land zones have id > 0, water zones id < 0. so negative product means domain mismatch
-				if (it->GetDominanceZone() * it2->GetDominanceZone() < 0)
```
azum4roll, line 2797: Just use ^ instead of * here?

JohnsterID: XOR has issues: (1) operator precedence means a ^ b < 0 is parsed as a ^ (b < 0) without parentheses, and (2) even with parentheses, 0 ^ negative gives a negative result, which would incorrectly trigger the condition for unrevealed plots (zone ID = 0). The explicit sign check preserves the original behavior while avoiding overflow.

Issue #12475, opened 2026-01-20
Expenses from Diplomacy deficit v2 [CP ONLY]


PauloMad:
### 1. Mod version

5.1.2

### 2. Installed Components

Community Patch only

### 3. Additional mods used

CP Only

### 4. Describe the Issue

@axatin 

As was mentioned in #12193 and #12367 there's a bug with Diplomacy Expenses that generates an artificial deficit for the player. This is still ongoing as of 5.1.2.

Attached you'll find a MPlayer game, with 2 humans, pick Morocco.

1. Turn 188 to 189 with Morocco sees a deficit start with -1 to Other Civs where there's no gold being traded from the civ. A deal ended that turn with Venice, where it provided 5 gold for Sugar.

2. Then it self-corrects eventually but strikes again 229 » 230 with -2, again we see a deal expiring 7 gold from America for Open Borders. 

3. Another jump 235 » 236, now incremental from -2 to -8 gold again on deal expiry, Sugar for Gold with Mongolia.

The deficit just keeps increasing from 236 onward.

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[morocco-deficit.zip](https://github.com/user-attachments/files/24725577/morocco-deficit.zip)

### 6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

I have my logs from Ethiopia side but not my friend's playing Morocco. If needed, let me know.

axatin:
Yeah I noticed that too already. It's a bug with deal renewals and it's present in every game, Community Patch and Vox Populi, SP and MP. I think it can't be fixed without completely rewriting the deal renewal code. Working on it...

PauloMad:
> Yeah I noticed that too already. It's a bug with deal renewals and it's present in every game, Community Patch and Vox Populi, SP and MP. I think it can't be fixed without completely rewriting the deal renewal code. Working on it...

This one is a bitch, it has such a huge impact on gameplay - glad you picked it up and it's identified. Thank you for your work on this.

Issue #12430, opened 2026-01-12
[5.1] Assertion Failed in Line 3347


brsoft1981:
### 1. Mod version

5.1

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

None

### 4. Describe the Issue

In the grey zone, when a deal ends (Horses to Songhai), i try to trade horses to them and I've got "Assertion Failed in Line 3347"

![Image](https://github.com/user-attachments/assets/6a6b2e92-917f-4477-a479-99b89750c8c8)
![Image](https://github.com/user-attachments/assets/e66c00b9-6e54-4022-bdb5-cf40060ef820)

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[SaveGame_Error3347.zip](https://github.com/user-attachments/files/24575375/SaveGame_Error3347.zip)

### 6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

[Logs.zip](https://github.com/user-attachments/files/24575378/Logs.zip)

axatin:
I don't understand, what exactly did you do before getting the error?

brsoft1981:
I tried to trade horses to Songhai. When I select 'horse' in the trade menu, I've got the error.

axatin:
Oh, you're trying to trade before the AI turn, I see now. Yeah that's another bug with deal renewals...

PR #12557, opened 2026-02-06
Deal renewals

### LINKED ISSUE #12487: Diplomacy Bug with Open Borders
### 1. Mod version

5.1.2

### 2. Installed Components

Community Patch only

### 3. Additional mods used

none

### 4. Describe the Issue

Sometimes you can charge AI twice for your Open Borders.
Image
Image
Image

[diplobug.zip](https://github.com/user-attachments/files/24759437/diplobug.zip)

_No response_

axatin:
- Deal renewal offers are now made one turn before the deal expires, not after it has expired
- When a renewal offer is rejected, the deal expires normally after the turn. When it is accepted, the old deal is immediately canceled and replaced with the new one
- This fixes issues with the processing of changed items (in particular GPT; #12475 )
- Also fixes #12487 and #12430
- Fix incorrect leader messages in trade screen
- Fix AI not always being able to equalize deals
- Show "What will make this deal work" button for renewal offers that aren't Acceptable to begin with

### DIFF
--- a/(1) Community Patch/LUA/Includes/TradeLogic.lua
+++ b/(1) Community Patch/LUA/Includes/TradeLogic.lua
@@ -944,6 +944,11 @@ function DoUpdateButtons()
 				elseif (iNumItemsFromUs > 0 and iNumItemsFromThem > 0) then
 					Controls.WhatWillMakeThisWorkButton:SetHide(false);
 				end
+			elseif (g_iDiploUIState == DiploUIStateTypes.DIPLO_UI_STATE_TRADE_AI_MAKES_OFFER) then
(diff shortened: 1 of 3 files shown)


(1) Community Patch/LUA/Includes/TradeLogic.lua
```diff
@@ -89,6 +89,30 @@ local offsetsBetweenFrames = 4;
 local oldCursor = 0;
 
 
+-- Returns the effective GPT of a player, taking into account any GPT in a renew deal
+function GetEffectiveGoldRate(pPlayer, iPlayer, iOtherPlayer)
```
vaderkos, line 93: Writing comment just to double check if parameters specified to `GetRenewDealGoldPerTurn`, are correct.

Issue #12539, opened 2026-02-01
Building_BonusFromAccomplishments being weird


Hokath:
### 1. Mod version

5.1.3

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

https://forums.civfanatics.com/threads/beta-cultural-components-5-6-uc.701034/

### 4. Describe the Issue

I've made a custom building that gives UnitCombatType a UnitProductionModifier from Accomplishments.
This exists on the Pitz Court for the ACCOMPLISHMENT_LONGCOUNT_DIPLOMAT, where it gives a 5% boost to Diplo units -- I have confirmed this works as intended.
For me, I instead used religion accomplishments 
ACCOMPLISHMENT_BELIEF_FOUNDER
ACCOMPLISHMENT_BELIEF_FOLLOWER
ACCOMPLISHMENT_BELIEF_ENHANCER
ACCOMPLISHMENT_BELIEF_REFORMATION
For Melee and Mounted units.

However this doesn't show up in the production tooltip (or appear to have any effect on production rate)
The relevant code seems to all be the block here
https://github.com/LoneGazebo/Community-Patch-DLL/blob/9fb57553c22c81aa69d67b0874aa7fde0794b78a/CvGameCoreDLL_Expansion2/CvCity.cpp#L14903
(note comment needs fixing)
But I can't see why this wouldn't work. Is there somewhere it excludes certain IDs from the loop?

Hokath:
Ok so, looking into this more, the bonus works for MOUNTED by not MELEE
I suspect then that the building can only have 1 unitcombat bonus from an accomplishment, either because there's a break clause or something else with the structure. I will look deeper

PR #12562, opened 2026-02-07
vectorize the BonusFromAccomplishments

Hokath:
addresses #12539
note also I found there was an operator precedence bug in the building production AI (manual says > happens after *)

I got in a bit of a knife-fight with the msvc compiler. Clang was happy but something about the pkBuildingInfo-> reference stopped it getting the right iterator with auto&. My python brain doesn't understand this, but if someone looks and sees I did it in a weird way, please correct it.

In game, it appears to be working fine.


azum4roll:
`auto` doesn't do anything close to what you expect in our version of C++.

qingyin3545:
Using the VS2013 toolchain would enable some C++11 features in the DLL. However, I can't resolve the compilation issues without incorporating VS2013 files into the workflow.

Hokath:
oh wow, ok, so in these older cpp versions auto does *checks notes*... nothing!?

JohnsterID:
Yes.

https://gist.github.com/JohnsterID/2d2441d470ec574e43a35767048d6f84

PR #12563, opened 2026-02-07
City State yield bonus generalization 9-79

Hokath:
https://forums.civfanatics.com/threads/9-79-city-state-bonus-balancing.701629

Instead of introducing like 20 new defines, I thought a better idea was to just create tables.
This way future modders can choose which yields are given by which city state, at both player (e.g. give great general points from militaristic) and city level (e.g. maritime-like effect for production)
I didn't touch the Militaristic CS hardcoding atm, or the Mercantile, but it's primed to be done later (e.g. all city-states could give units)

This turned into a bit of an epic quest, and I am yet to even try to compile it. But I wanted to show the draft in case anyone has any feedback.


azum4roll:
1. There's no reason not to make new tables times100, especially when the maritime defines already do support it.
2. We don't use xml to insert into composite tables. Make a CoreMinorYieldChanges.sql instead.
3. Database stuff is done in `CacheResults`, not `Init`.

Hokath:
Yeah I was a bit lost with CacheResults. 
MinorCivAI doesn't currently have one (because there were no tables for it before). If I add one, in which file/function would I need to call it (or is it called automatically for this type of class?)

azum4roll:
image

Hokath:
That's the per-civ one, these functions currently live under the CvMinorCivAI class defined higher up

azum4roll:
It's similar to the relation between `CvPlayer` and `CvCivilizationInfo`. You can call `GC.getMinorCivInfo(GetMinorCivType())` to get the `CvMinorCivInfo` from `CvMinorCivAI`. No need to check null on the result.

Hokath:
Yeah I get that part. But if I load the table in with the CvMinorCivInfo, doesn't that mean ill be reading and then holding copies of the xml table for every City-State that gets placed on the map?

CvGameCoreDLL_Expansion2/CvMinorCivAI.h
```diff
@@ -985,6 +984,16 @@ class CvMinorCivInfo : public CvBaseInfo
 	// Arrays
 	int* m_piFlavorValue;
 	std::vector<CvString> m_vCityNames;
+
+	// do it this way so querying a particular yield for entries is O(1) not O(n)
```
azum4roll, line 988: More like O(ln #yields * ln #eras)?

(1) Community Patch/Database Changes/Minors/CoreNewMinorTables.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<GameData>
+	<!-- Determines what player yields a city-state type will give when friendly, in what era -->
+	<Table name="MinorCivilizationTraits_FriendYieldBonuses">
+		<Column name="MinorCivTraitsType" type="text" reference="MinorCivTraits(Type)" notnull="true"/>
```
azum4roll: Shouldn't it be `MinorCivTraitType` btw?

Hokath: ye true

CvGameCoreDLL_Expansion2/CvMinorCivAI.h
Hokath: yeah i guess ln for looking
just yields tho (dont bother seeing if we havent reached first valid era yet)

Hokath:
Ok, we are almost there. 
The code compiles and load through the mod menu, populating the database.

However, on map creation we get a CTD. I assume this is being cause therefore by the city-state initialization. Lua log gets to the "placing goodies" phase before it cuts out.
https://www.dropbox.com/scl/fi/1q5b6b4ub81rqnji24bdp/Logs.zip?rlkey=fj6yj1gusr02d48uiij3xyvue&dl=0

I don't have the VSCode setup to open the minidump (there is no other way, is there?)
https://www.dropbox.com/scl/fi/swywmutx5fakfmshirj7w/CvMiniDump_20260226_135653_Tag_aba2f99_Release.zip?rlkey=wd1l7y0g89961u5cwhdc87dwt&dl=0

Could I ask for some help?

(1) Community Patch/Database Changes/Minors/CoreMinorYieldChanges.sql
```diff
@@ -0,0 +1,30 @@
+// gives the most recent era's value (rather than adding all valid eras)
```
vaderkos: Is this correct comment in sql, it should be `--` isn;t it?

axatin:
`m_piYieldPerTurnFromAnnexedMinors` and the other new arrays need to be resized to NUM_YIELD_TYPES in CvPlayer::reset
```
m_piYieldPerTurnFromAnnexedMinors.clear();
m_piYieldPerTurnFromAnnexedMinors.resize(NUM_YIELD_TYPES, 0);
```

CvGameCoreDLL_Expansion2/CvMinorCivAI.cpp
```diff
@@ -13940,76 +13975,77 @@ bool CvMinorCivAI::DoMajorCivEraChange(PlayerTypes ePlayer, EraTypes eNewEra)
...
 				else
 				{
-					if (iOtherDelta != 0)
+					if (iCityDelta != 0)
 					{
-						pLoopCity->ChangeBaseYieldRateFromCSFriendship(YIELD_FOOD, iOtherDelta / 100);
-						//CUSTOMLOG("updated non-capital food in %s by %d/100 for friendship with %s, current value is %d", pLoopCity->getNameKey(), iOtherDelta, m_pPlayer->getNameKey(), pLoopCity->GetBaseYieldRateFromCSFriendship(YIELD_FOOD));
+						if (iInfluenceLevel == 1)
+						{
+							pCity->ChangeBaseYieldRateFromCSFriendship(eYield, iCapDelta);
```
axatin: Needs to be iCityDelta here and in the next block

axatin: And it should be `/ 100`. I'd suggest renaming `GetCityYieldFlatBonus` to `GetCityYieldFlatBonusTimes100`, that'll make it easier to spot such errors. 

While we're at it, how about changing `Change/GetBaseYieldRateFromCSFriendship` to x100 functions?

(1) Community Patch/Database Changes/Minors/CoreMinorYieldChanges.sql
Hokath: true, i think i had a multiline comment /* */ and only partially removed it, whoops!

CvGameCoreDLL_Expansion2/CvMinorCivAI.cpp
Hokath: Yes, I'd like to make everything x100

Hokath:
we've got past the CTD
next issue: no yields coming from city state allies at all! for tomorrow...

Hokath:
Confirmed that if I set CvMinorCivAI::GetYieldFlatBonus to just return int 10, then things seem to work fine.
So for some reason we are exiting as zero when we shouldn't. Looking deeper.

GetAllyYieldBonusValue is definitely called, so it's deeper still

Hokath:
image

we have liftoff

Hokath:
Ok I have x100 all the major functions.
Seems to working in game at first glance. Still, please review since it's a lot.

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -22870,267 +22802,175 @@ void CvPlayer::ChangeConversionModifier(int iChange)
...
-		// Pre-Medieval
-		else
+		for (int iI = 0; iI < NUM_YIELD_TYPES; iI++)
 		{
-			iBonus += GD_INT_GET(FRIENDS_CULTURE_BONUS_AMOUNT_ANCIENT) + GD_INT_GET(ALLIES_CULTURE_BONUS_AMOUNT_ANCIENT);
+			int iBonus = 0;
+			YieldTypes eYield = (YieldTypes)iI;
+			
+			iBonus += GET_PLAYER(eMinor).GetMinorCivAI()->GetCityYieldFlatBonusTimes100(GetID(), eYield, eEra, 2, true);
+			iBonus /= 100;
```
axatin: this needs to be removed

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -22870,267 +22802,175 @@ void CvPlayer::ChangeConversionModifier(int iChange)
...
+			iBonus *= iSign;
+	
+			//update if necessary
+			//bonus yields are tracked on city level
+			getCapitalCity()->ChangeBaseYieldRateFromCSAllianceTimes100(eYield, iBonus);
+			m_piYieldInCapitalFromAnnexedMinors[eYield] += iBonus / 100;
```
axatin: Let's make this a times100 value too

PR #12575, opened 2026-02-10
Don't show starvation notification for small food deficits

axatin:


azum4roll:
What threshold do you start getting unhappiness from famine?

PR #12582, opened 2026-02-11
(9-71) Isibaya (Zulu) Buff

ron-murhammer:
https://forums.civfanatics.com/threads/9-71-isibaya-zulu-buff.701469/


azum4roll:
Unless we use squash and merge and rename the commit to include the proposal number, it'll be lost.

PR #12590, opened 2026-02-11
(9-04) Earlier Great Work Slots

legenciv:
Proposal's thread: https://forums.civfanatics.com/threads/9-04-earlier-great-work-slots.701262/

Implemented with Palace theming set similarly to Broadcast Tower requirements, used as reference in the proposal when mentioning the theming would work with any 3 GWs; the Broadcast Tower's "any GWs" was only feasible with 2 GW slots. This is the closest to what was proposed.

This was done as follows:

- 3 GWs from the same Era
- 3 GWs from different Eras
(Consecutive Eras is contained in 'different Eras', the Broadcast tower uses for an extra flavor text).

Flavor text for these two theming conditions changed to '{1_EraAdjective} Epics' and 'Timeless Epics', respectively; the proposed 'National Epics' is the same theme name on the Amphitheater.

Tested with IGE: slots on Monument, Shrine, Palace. Temple and Castle removal confirmed. Unique versions (Qila, Tetraconch, Stele) following the changes properly. Palace's theming for both conditions.

### DIFF
--- a/(2) Vox Populi/Database Changes/Text/en_US/City/NewBuildingText.xml
+++ b/(2) Vox Populi/Database Changes/Text/en_US/City/NewBuildingText.xml
@@ -1094,7 +1094,7 @@
 			<Text>Timeless Epics</Text>
 		</Row>
 		<Row Tag="TXT_KEY_PALACE_THEMING_BONUS_HELP">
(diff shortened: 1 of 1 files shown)


azum4roll:
Theming doesn't need to have a restriction.

legenciv:
@azum4roll Thanks. Now that you mention, I could see how. I was under the impression that Building_ThemingBonuses would only work if one condition was set to 1.

I'll close this pull request, I accidently added code from another proposal yesterday. I'll open a new one without the mess.

PR #12594, opened 2026-02-12
Add bombardment highlight

schnetziomi5:
Adds plot highlighting showing which plots can be bombarded when [potentially] moving a ranged unit.

The highlight style is a thinner version of the normal ranged attack border, so it still looks decent when it occasionally overlaps with the movement border.

20D850~1
20B356~1

### DIFF
--- a/(1) Community Patch/Core Files/Overrides/Includes/Bombardment.lua
+++ b/(1) Community Patch/Core Files/Overrides/Includes/Bombardment.lua
@@ -320,3 +320,106 @@ function EndNukeAttack()
 	Events.SerialEventMouseOverHex.Remove( DisplayNukeArrow );
 	ClearUnitHexHighlights();
 end
+
+-------------------------------------------------
+-- Ranged Strike Preview when moving...
+do
+	-- this code is somewhat optimized...
+	
+	local x,y ;
+	local range ;
+	local curr_ring ; 
+	
+	local nextStep ;
+	local stepRepeat ;
+	
+	local nextPlot = { 
+		-1,  1;
+		-1,  0; 
+		 0, -1; 
+		 1, -1; 
+		 1,  0; 
+		 0,  1; 
+	};
+	
+	local ret_table = {x=0,y=0} ;
+
+	local ret = function()
+		if nextStep >= 7 then
+			curr_ring = curr_ring + 1 ;
+			if curr_ring > range then return nil end ; 
+			y = y + 1 ;
+			nextStep = 1 ;
+			stepRepeat = curr_ring - 1 ;
+		else
+			
+			x = x + nextPlot[nextStep*2-1] ;
+			y = y + nextPlot[nextStep*2] ;
+			
+			stepRepeat = stepRepeat - 1 ;
+		end
+		
+		if stepRepeat <= 0 then
+			nextStep = nextStep + 1 ;
+			stepRepeat = curr_ring ;
+		end
+		
+		ret_table.x = x ;
+		ret_table.y = y ;
+		
+		return ret_table,x,y ;
+	end
+	
+	iteratePlots = function( base, _range, _curr_ring )
+		
+		range = _range ;
+		curr_ring = _curr_ring or 1 ;
+		if curr_ring == 0 then
+			x,y = base.x , base.y - 1 ;
+			nextStep = 6 ;
+			stepRepeat = 1 ;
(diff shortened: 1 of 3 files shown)


azum4roll:
Can you try moving rapidly between tiles to test if the game can handle high frequency highlighting/unhighlighting?

schnetziomi5:
... so I selected a unit, held the right mouse button and moved the cursor erratically around the screen for half a minute. Game was fine.

On discord you mentioned that frequent tile highlighting would sometimes crash the game - and also that EUI seems to somehow avoid it. I wrote the code for this completely free, that is I didn't look up EUI's version, but apparently it also works. 

Do you know of any examples for tile highlighting code that is suspected or was found to cause crashes?

azum4roll:
The HighlightHoveredBuildingBonuses function in EUI CityView. It has a highlight colour though.

(1) Community Patch/Core Files/Overrides/Includes/Bombardment.lua
```diff
@@ -319,4 +319,107 @@ function EndNukeAttack()
...
+			stepRepeat = curr_ring ;
+		end
+		
+		ret_table.x = x ;
+		ret_table.y = y ;
+		
+		return ret_table,x,y ;
+	end
+	
+	iteratePlots = function( base, _range, _curr_ring )
```
vaderkos, line 373: Is this local or global?

schnetziomi5: intentionally global (for the moment)
(?)

Hokath:
Would this be easily compatible with #12428 ?

schnetziomi5:
Yes - even though I am not a fan of the siege rework proposal of yours this PR would in fact not exist without it and is meant to help playing with it. 

Technically I modified LuaUnit.canEverRangeStrikeAt to enable calling `bool CvUnit::canEverRangeStrikeAt(int iX, int iY, const CvPlot* pSourcePlot, bool bIgnoreVisibility)` with the full set of parameters - crucially enabling to pass a sourcePlot, so any plot that can be shot at according to the .dll will be highlighted by the preview.

I also intend to edit the standard bombard plot highlighting code to use this new functionality, so that we don't need to update the lua code to mirror the c equivalent of it every time we make a change - as you had to in your PR. However neither LuaCity nor CvCity has a `canEverRangeStrikeAt` function right now, so either I need to add those or keep handling the city case more thoroughly in lua. Will do that when I have the time, and either add it to this PR or make a new one.

(1) Community Patch/Core Files/Overrides/Includes/Bombardment.lua
vaderkos: Could it be in style with other styles in other places like PascalCase and could you include lua documentation with types?
And also luals and luals-rust treat variables that start with `_` as optional, so are this parameters intentionally optional?

schnetziomi5: PascalCase Is UnReadAble To Me.
Hmm... more seriously I would like to put it somewhere else entirely, maybe CPK, but there are more pressing things right now. When I do I will see to it having the same style and documentation as elsewhere.
About the _ prefix: Personally I only use notepad so I didn't know that some code checkers interpret them that way - I think the lua manual says you shouldn't use such names normally - which I didn't: I use such names in a few specific cases only, in this case for a construct like
```
local someValue ;
function setSomeValue( _someValue )
      someValue = _someValue ;
end
```
where I want to set someValue without having to invent another name for either the local variable or the function argument. Is there a accepted convention for such cases?

(also note that the iterator code is intentionally written to not create any new object[which would require allocation] upon being called or during iteration, it operates on the stack only.)

vaderkos: Yeah, I've also used it in similar manner :) , but when I noticed it suppresses checks I stopped. Because it doesn't only treat such params as optional, but if you have a variable name starting with `_` it also disables `unused variable` rule on them. I'd really suggest you try `luals`, makes life much easier with intellisense an checks :D. The guide should e in docs.

azum4roll: To me functions should be PascalCase to follow the game core API.

PR #12598, opened 2026-02-12
Let air units inflict (and receive?) plagues

Hokath:
I would like to drop some special munitions.

This looks like quite a minimal piece of code. 
Is there anywhere else (AI?) it needs changing?


azum4roll:
But planes getting boarded, dazed etc. don't make much sense. They should only get plagues inflicted by planes.

PR #12608, opened 2026-02-13
9-86, Unit upgrade costs increased by 25%

Alpakinator:
As per:
https://forums.civfanatics.com/threads/9-86-increase-unit-upgrade-costs.701724/#post-16912596


azum4roll:
The value is an int in the DLL so this won't work.

Alpakinator:
Good catch, my bad. Fixing.

azum4roll:
Now `UNIT_UPGRADE_COST_PRODUCTION_PERCENTAGE` isn't in the Defines table. You can use floats to store the number in the DLL instead.

UI_bc1/Core/EUI_tooltip_library.lua
```diff
@@ -197,7 +197,7 @@ end
 -- How much does it cost to upgrade a Unit to a shiny new eUnit?
 local function unitUpgradePrice( unit, unitUpgrade, unitProductionCost, unitUpgradeProductionCost )
 	local upgradePrice = GameDefines.BASE_UNIT_UPGRADE_COST
-		+ math_max( 0, (unitUpgradeProductionCost or unitUpgrade.Cost or 0) - (unitProductionCost or unit.Cost or 0) ) * GameDefines.UNIT_UPGRADE_COST_PER_PRODUCTION
+		+ (math_max( 0, (unitUpgradeProductionCost or unitUpgrade.Cost or 0) - (unitProductionCost or unit.Cost or 0) ) * GameDefines.UNIT_UPGRADE_COST_PRODUCTION_PERCENTAGE) / 100
```
azum4roll: There's no reason to calculate the upgrade cost in Lua. Just follow non-EUI and call `unit:UpgradePrice()` instead.

Alpakinator: I suspect it's needed if upgrade cost would be displayed with no unit on the map, so in the civ selection menu or in civilopedia. Am i wrong?

Alpakinator:
I added it to defines, is it ok?. I'm not yet sure if its ok to replace lua calc with call to c++ calc, due to InfoTooltip... being run with no unit instance.

UI_bc1/Core/EUI_tooltip_library.lua
azum4roll: Upgrade cost isn't shown anywhere else. Not with the new tooltips.

The number shown on EUI is only the base amount without discounts anyway.

azum4roll: Actually this function should be completely unused now, since apparently UnitPanel doesn't call it.

Alpakinator: Yea

> Actually this function should be completely unused now, since apparently UnitPanel doesn't call it.

Yeah, seems unused. Should i remove it?  If in the future we'll show unit upgr. costs in eg. in main menu, then it'd be useful i guess. And in that case making it call UnitPrice instead of Lua calculation would cause problems.

azum4roll: Depends on if it's desired. I don't think upgrade costs affect your build decisions.

Alpakinator: Yeah, they don't. I'd prefer not to delete/replace this, cause IMO it's outside of the scope of this proposal.

axatin:
Files in UI_bc1 should not be modified. If necessary, modify the copy in (3) Compatibility Files

Alpakinator:
> ... You can use floats to store the number in the DLL instead.

I did that now, and went back to the original variable name, so changes to UI_bc1 or (3) Compatibility Files are no longer needed.

PR #12609, opened 2026-02-13
9-53, SPAD VII +5 RCS, Ace Pilot IV

Alpakinator:
As per:
https://forums.civfanatics.com/threads/9-53-spad-s-vii-buff.701384/#post-16912249


azum4roll:
It shouldn't lose the Air Logistics promotion.

Alpakinator:
My bad, fixed:

PR #12610, opened 2026-02-13
9-47 indirect-fire dread, +1 range bs, mc, yamato

Alpakinator:
As per:
https://forums.civfanatics.com/threads/9-47-battleship-buff.701371/


azum4roll:
What about strategy texts?

Alpakinator:
I'll modify them

PR #12612, opened 2026-02-13
Tidy "max" tables into Belief_YieldPerXFollowersLocal

Hokath:


Hokath:
Coming back to finish this.

It's also noted that AI eval of the new follower yield beliefs seems very high. I have investigated and indeed the way it is evaluated was affected by the cap. The whole evaluation is a bit weird (and in places, wrong)

Hokath:
This is going to have that issue that the MaxYieldPerFollower and FollowerRequiredPerYield are stored at the Religion level, but two beliefs might have different values and overwrite each other. E.g. Thrift and Creativity.

So we need to refactor the whole thing to calculate it in the religion function for yields, at the belief loop stage.

Hokath:
6 followers with Asceticism and Creativity. 3 food and 2 culture. Bingo.

PR #12614, opened 2026-02-13
(9-33) Atoll Map Effects

Hokath:
https://forums.civfanatics.com/threads/9-33-atoll-map-effects.701316/page-2
Based on the investigation we went through in the thread, I believe this is the only part of the DLL that needs to change.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvUnit.cpp
+++ b/CvGameCoreDLL_Expansion2/CvUnit.cpp
@@ -16506,11 +16506,15 @@ int CvUnit::GetMaxDefenseStrength(const CvPlot* pInPlot, const CvUnit* pAttacker
 
 	if(pInPlot != NULL)
 	{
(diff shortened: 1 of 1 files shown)


azum4roll:
Would be better if units only benefit from terrain defense in their native domain (so ships don't benefit from hills and hovering units don't benefit from atolls etc.)

Hokath:
I started to do it and I found that CvUnit::isNativeDomain returns True for any Hovering unit. So that wouldn't work as the condition. I then realized Helicopters have the NoDefensiveBonus type promotion anyway, so that seems to be how it is currently handled.

Alternative condition 
if (noDefensiveBonus() || getDomainType() != pInPlot->getDomain())
will return false for anything that isnt land or sea domain. So no defensive bonuses for Hovering or Air units this way (ever). Seems fine?

azum4roll:
That's fine.

PR #12615, opened 2026-02-14
Add Shoshone 4UC Strategy Text

ron-murhammer:
Feedback welcome.


Alpakinator:
I'd change 'longer as their Combat Strength' to 'longer because their Combat Strength' to not repeat 'as'.

I'd add advice for Yellow Brow stationing in Encampments and fortifications. Promotion, like march, or stalwart since now martch will stack with fortifying healing:
"
Following Shock promotion line helps counterattacking after defending from enemy offensive due to flanking bonuses. Later, March synergizes with healing from fortifying since these healing sources stack. [NEWLINE][NEWLINE] Alternatively, Drill improves resistance to cities and enables Yellow Brows to push near them, fortify and protect Siege units better. Then, Stalwart fits to the defensive theme of this unit greatly.
"

ron-murhammer:
@Alpakinator Updated to reword it a bit and add a bit more detail on how to best use them. I think its better to avoid too much detail around promotions unless the recommendations are very specific to the UU (here I think most of the recommendations for Tercios or even melee/gunpowder line apply). Otherwise we end up with lots of duplicated text that is difficult to maintain.

(2) Vox Populi/Database Changes/Text/en_US/Civilizations/NewCivilizationText.xml
```diff
@@ -1406,7 +1406,7 @@
...
 		</Row>
 		<Row Tag="TXT_KEY_UNIT_YELLOW_BROW_STRATEGY">
-			<Text></Text>
+			<Text>Yellow Brows are defensive oriented units which receive bonus Combat Strength when fortified and become stronger as they take damage. This allows them to be more durable and often stay fortified on the front lines longer because their Combat Strength doesn't decrease as they become damaged like regular units.[NEWLINE][NEWLINE]If you use ranged and mounted units to deal most of the damage then Yellow Brows can stay fortified so they can absorb more enemy attacks and heal.</Text>
```
azum4roll: Need hyphen in defensive-oriented

azum4roll: Their combat strength actually increases as they take damage, may have to mention it instead of "doesn't decrease".

PR #12580, opened 2026-02-10
C9 database proposals

Hokath:
Currently implemented:
https://forums.civfanatics.com/threads/9-80a-ranged-attacks-do-more-damage-to-civillians.701600/
https://forums.civfanatics.com/threads/9-78a-chancery-tweak-alternative.701608/
https://forums.civfanatics.com/threads/9-75a-nerf-third-alternative.701539/
https://forums.civfanatics.com/threads/9-68-songhai-ua-tweak.701455/
For that proposal we also extended the practice of giving units Combat Class designations so they can be picked out by policy/trait/promo tables, this includes Archaeologists, Great People, Missiles, and Nukes. This is a separate commit for clarity, but is required.
https://forums.civfanatics.com/threads/9-62a-integrate-civilian-puppet-purchases.701556/
https://forums.civfanatics.com/threads/9-61-colonists-found-puppets.701425/
https://forums.civfanatics.com/threads/9-56-venetian-liburna-upgrades-to-fusta.701542/
https://forums.civfanatics.com/threads/9-51-belated-4uc-re-proposal-ii-carthage.701402/
https://forums.civfanatics.com/threads/9-48-shrapnel-rounds.701908/
https://forums.civfanatics.com/threads/9-43-altitude-training-change.701352/
https://forums.civfanatics.com/threads/9-41-religion-iii-per-follower-beliefs.701321/
https://forums.civfanatics.com/threads/9-34-marsh-camp.701318/
https://forums.civfanatics.com/threads/9-29c-swap-hwacha-and-chu-ko-nu-abilities-alternative.701514/
https://forums.civfanatics.com/threads/9-24-featureless-desert-yields.701282/
https://forums.civfanatics.com/threads/9-06b-sovereignty-rework-same-as-op-but-higher-bgp-on-fewer-buildings.701404/
https://forums.civfanatics.com/threads/9-14-policy-tweaks-i-ancient-production.701275/
https://forums.civfanatics.com/threads/9-13-policy-tweaks-0-tree-descriptions.701274/
https://forums.civfanatics.com/threads/9-15-policy-tweaks-ii-per-military-unit-removal.701277/
https://forums.civfanatics.com/threads/9-16-policy-tweaks-iii-progress-opener.701468/
https://forums.civfanatics.com/threads/9-17-policy-tweaks-iv-tidy-up-statecraft.701278/
https://forums.civfanatics.com/threads/9-18-policy-tweaks-v-artistry-buff.701280/
[...]

PR #12621, opened 2026-02-14
Try to get Mongolia to make Skirms

Hokath:
This code was commented out by Ilteroi 4 years ago because all the bonuses were like +5 and didn't matter.

Not sure if it's worth caching any of this, but need to see if it works, first.

Edit: In my first game I'm not seeing Mongolia produce Skirms, still (or anyone else, for that matter...)
Turn 207 (early Renaissance) Karakorum makes 2 heavy skirms back-to-back.
I notice they had a lot of Archers->Comp Bows early, is this happening because Archery units are all treated together?

### DIFF
--- a/(1) Community Patch/Database Changes/Civilizations/NewTraitTables.xml
+++ b/(1) Community Patch/Database Changes/Civilizations/NewTraitTables.xml
@@ -436,7 +436,7 @@
 		<Column name="MovesChange" type="integer" notnull="true"/>
 	</Table>
 
-	<!-- Free promotion on specified unit classes, skipping validations -->
+	<!-- Free promotion on specified unit classes, skipping validations. The Production AI will treat these units like UUs. -->
 	<Table name="Trait_FreePromotionUnitClass">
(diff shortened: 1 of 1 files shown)


Hokath:
#8783

Hokath:
Using @nobubblegums's mod, I do see an occasional Chariot Archer/Skirm, though they haven't got above 2 so far.
So I'm not sure this is the solution by itself.

Hokath:
Yeah even at a boost of 500, there's no change in behavior. 
For whatever reason, this weighting is doing nothing.

I guess the IsPromotionValidForUnitCombatType might evaluate to zero here

Hokath:
Success!
image

Hokath:
when the new combat classes are added in #12580 
we can move the English Archaeologist to the UnitCombat table, so they won't get this AI weighting -- although it shouldn't matter too much because they already have cap+high weight.

I have added a comment in the new table .xml to signpost that the AI behaves this way to unit classes in the table.

azum4roll:
Which civs are affected besides Mongolia?

Hokath:
Atm England (Archaeologist)

PR #12625, opened 2026-02-14
Building_ReligionYieldFromFaithPurchasableBuildings

Hokath:
This allows buildings to give yields in a city, or on empire level, to buildings that you have the ability to faith purchase (locally).
So a City that is Hindu with a Cathedral (Hindu belief) and Teocalli (Islamic belief) would only benefit from the Cathedral value.

This naturally extends to faith-purchase buildings like Monasteries and optionally faith-purchasable buildings (both with beliefs like Faith of the Masses, and unique components like old Spanish Mission).

What I thought was going to be an easy table turned into a bit a brain teaser.
The yields are added to "Religion" tally because of how they have to be recalculated when the religion changes.


(1) Community Patch/Database Changes/Text/en_US/UI/CoreNewUIText.xml
```diff
@@ -1536,6 +1536,9 @@
 		<Row Tag="TXT_KEY_PRODUCTION_BUILDING_ITR_BOOST">
 			<Text>{1_YieldBoosts} to [ICON_INTERNATIONAL_TRADE] Internal Trade Routes from this City</Text>
 		</Row>
+		<Row Tag="TXT_KEY_PRODUCTION_BUILDING_FAITH_PURCHASABLE_BUILDING_YIELDS">
+			<Text>{1_YieldBoosts} for each owned Building that would currently be purchasable with [ICON_PEACE] Faith</Text>
```
azum4roll: In this case it should be {1_YieldBoosts} to City since it's not on the building. Weird implementation though.

PR #12626, opened 2026-02-14
9-84, Ship garrisons give 25% of max(CS/RCS) to city

Alpakinator:
As per:
https://forums.civfanatics.com/threads/9-84-naval-units-in-cities-increase-city-strength-by-50-of-combat-strength-compared-to-land-units.701579/
Partial revert of:
https://forums.civfanatics.com/threads/2-57-naval-units-in-cities-dont-increase-city-strength.680937/

Ships give 25% of max(CS, RCS) to the city strength (not counting towards city ranged attack strength). So same as land units but 25%, not 50%.

In community patch both unit types still contribute only 20%.

Added defines to modify both.

Since only the selected garrison unit increases city combat strength and receives damage when the city is attacked, I adapted garrison selection code ([CvPlot.cpp](https://github.com/LoneGazebo/Community-Patch-DLL/compare/master...Alpakinator:Community-Patch-DLL:9-84_ships_25%25CS_to_garrison?expand=1#diff-3bc6de085cd3ffe0e7db9198f97552c1a8711f5b89cc4b26729cdc74add1a624)) to select between land and naval garrison based on the highest effective contribution to city strength. 

Between now and 2-57 this code used to select land units weighed at 100% strength, and ships at 0%, while ships contributed nothing, makes sense. Before 2-57 land at 100% and naval at 25%, while ships contributed 50% of their max(CS, RCS) to garrison. This resulted in Land unit decreasing city strength when entering it, while a stronger ship was already a garrison, because the land unit was chosen as the garrison over the ship:
image

 Now this won't happen:
image
image
image
image

Allowed HomelandAI to consider ships for putting as a garrison for the purpose of increasing city strength to get yield bonuses (Authority).

Decreased AI penalty for putting ships as garrison.


Hokath:
Nice! Although the old version was "bugged" with that interaction, I preferred it before. Now we have the best of both worlds.

Alpakinator:
(Edit) force pushed:
Removed a block preventing TacticalAI from even considering ships as garrisons:
https://github.com/LoneGazebo/Community-Patch-DLL/pull/12626/changes/0f4b3bb2c0fef2c5bd51d1cbdfe00559a1c95daf#diff-61b60d0f48cbb15889ecd29425434b3e73db94629edfaf3ab637c64340598df7R1487

Prevented peace AI (HomelandAI) from not considering garrison city strength and not exceeding 10 combat strength if a civ has both authority (wanting 10 CS) and imperialism (not needing 10CS) garrison bonuses:
https://github.com/LoneGazebo/Community-Patch-DLL/pull/12626/changes/0f4b3bb2c0fef2c5bd51d1cbdfe00559a1c95daf#diff-b1aea56648e429dbcb215a03b7538545ae50ae252d23360d311e7a05cf046e9fR745 
Almost no difference rn. cause when you have imperialism there's no way to have so weak garrisons, but it's future change proof.

Alpakinator:
On a test map without barbarians or neighbours authority civs do garrison <10 strength cities:
image

but so do non-authority ones  (here, progress):
image
there seems to be no preference (edit: more tests and clearly, on average authority ai puts garrisons to get above 10CS  much more than non authority, when at peace. It avoids avoids garrisoning that will only reach 8 or 9 CS)
 
edit: As soon as there are many barbarians AI gives up on garrisoning which is good. With moderate barbs authority still garrisons at peace, good.
 
 And during war sometimes uses ships as garrisons instead of fighting with them. But I saw it only in case of 2 cities in 3 AI playthroughs.

Alpakinator:
Civilopedia has 2 overlapping modified vanilla articles talk about garrisoning in a city, ranged attacks, etc. Removed the one with the less accurate title and improved the article that remains.

azum4roll:
On one hand ships can't attack while garrisoned, on the other hand it protects the ship if it's outnumbered (but the city isn't too exposed). It's probably still a bad idea to garrison a ship if the city is exposed to lots of attacks (the ship dies anyway).

(1) Community Patch/Database Changes/Text/en_US/Concepts/CoreConceptTextChanges.sql
```diff
@@ -5,18 +5,11 @@ UPDATE Language_en_US
...
...t adds [COLOR_YELLOW]50%[ENDCOLOR] of its [ICON_STRENGTH] Combat Strength or [ICON_RANGE_STRENGTH] Ranged Combat Strength, whichever is greater, to the city [ICON_STRENGTH] Combat Strength if it''s a land unit or [COLOR_YELLOW]25%[ENDCOLOR] if it''s a naval unit. This contribution doesn''t increase City [ICON_RANGE_STRENGTH] Ranged Combat Strength.[NEWLINE][NEWLINE]If a city is attacked while a Garrison is in the city, the Garrison will deflect some of the damage onto itself, thus offering the city even more protection. Be careful, however, as a Garrison can be destroyed this way.[NEWLINE][NEWLINE]If both a land and a naval unit are stationed in the city, their contributions do not stack; only the unit with the higher [ICON_STRENGTH] Combat Strength contribution applies it, and is the garrison that absorbs incoming damage.[NEWLINE][NEWLINE]Additional combat units may move through the city, but they cannot end their turn there. (So if you build a combat unit in a city with a garrison, you have to move one of the two units out before you end your turn.)[NEWLINE][NEWLINE]A unit garrisoned in the city may attack surrounding enemy units [COLOR_YELLOW]unless it''s a naval unit[ENDCOLOR]'
```
azum4roll: We should standardize `City [ICON_RANGE_STRENGTH] Ranged Combat Strength` to `[ICON_RANGE_STRENGTH] City Strike Strength`, to avoid inconsistencies when a city defends against ranged attacks.

A new unit produced in the city will appear adjacent to the city if there's already a same domain garrison.

(2) Vox Populi/Database Changes/Text/en_US/Concepts/ConceptTextChanges.sql
```diff
@@ -207,7 +207,7 @@ UPDATE Language_en_US
...
 WHERE Tag = 'TXT_KEY_COMBAT_CITYFIRINGATTACKERS_HEADING3_TITLE';
 UPDATE Language_en_US
-SET Text = 'A city has a Ranged Combat Strength equal to its full Strength at the start of combat, and [COLOR_YELLOW]it has a range of 1. This range increases as the game progresses based on researched technologies (look for the "ranged strike" icon in the tech tree for these technologies).[ENDCOLOR] It may attack any one enemy unit within that range. Note that the city''s Ranged Combat Strength doesn''t decline as the city takes damage; it remains equal to the city''s initial Strength until the city is captured.'
+SET Text = 'A city has a [ICON_RANGE_STRENGTH] Ranged Combat Strength equal to its full [ICON_STRENGTH] Combat Strength minus the garrison contribution to the city''s [ICON_STRENGTH] Combat Strength. Therefore stationed units don''t change the strength of city''s ranged attacks.[NEWLINE][NEWLINE]It has 1 range, increasing to 2 after building Walls and to 3 after building Arsenal.[COLOR_YELLOW]Note that the city''s Ranged Combat Strength doesn''t decline as the city takes damage; it remains equal to the city''s initial Strength until the city is captured.[ENDCOLOR]'
```
azum4roll: Same here. `Therefore` needs to be followed by a comma.

CvGameCoreDLL_Expansion2/CvTacticalAI.cpp
```diff
@@ -4548,9 +4539,9 @@ CvUnit* CvTacticalAI::FindUnitForThisMove(AITacticalMove eMove, CvPlot* pTarget,
 				if (pLoopUnit->getUnitInfo().GetDefaultUnitAIType() == UNITAI_EXPLORE)
 					iExtraScore -= 50;
 
-				//naval garrisons cannot attack inside cities and don't increase city strength...
+				//naval garrisons cannot attack inside cities and contribure only 25% of their strength to city strength
 				if (!pLoopUnit->isNativeDomain(pTarget))
-					continue;
+					iExtraScore -= 35;
```
azum4roll: This needs to scale with the Define values, for both land and naval units.

CvGameCoreDLL_Expansion2/CvTacticalAI.cpp
azum4roll: I think we should remove this block. Fortifying in cities only affects missile damage.

Alpakinator: Yup, removing

CvGameCoreDLL_Expansion2/CvTacticalAI.cpp
Alpakinator: I'll add that in the force push in a moment, perhaps in an overkill way

(1) Community Patch/Database Changes/Text/en_US/Concepts/CoreConceptTextChanges.sql
Alpakinator: yup

Alpakinator:
force pushing the fixes:

Alpakinator:
The point of the longer TacticalAI code is to make the score dependent on effective increase in city strength when adding that unit as garrison. If a unit will double city strength by 100%, it'll increase that units score by 120. if only by +10%, then only +12 score. This accounts for different land and naval divisors of combat strength.

and -50 score if its a ship since they can't attack from garrison.

Seems in a good magnitude relative to the +30 when unit is ranged, or -20 per turn of distance from the city

PR #12629, opened 2026-02-14
Improvements with era tracking

Hokath:
Finally closing my own Feature Requests
It's like a character arc

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlot.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlot.cpp
@@ -14325,7 +14325,7 @@ void CvPlot::AddArchaeologicalRecord(GreatWorkArtifactClass eType, PlayerTypes e
 		return;
 
 	// if an improvement has set an era already, dont change it
-	EraTypes eNewEra = m_eEra;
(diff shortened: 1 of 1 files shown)


Hokath:
Tested in game

azum4roll:
An industrial era artifact sounds bad. Technically it could even be Modern or Atomic if the tech leader avoids Archaeology and there's an improvement for those eras.

Hokath:
IRL Museums are filled with Modern Era Artifacts :)

azum4roll:
And how does it work in game? Does the landmark get negative yields? (Possible scenario if an Industrial civ conquers a city owned by a Modern civ)

Hokath:
In that case it just doesn't do anything. Code is
```
int iYieldChangePerEra = pkImprovementInfo->GetYieldChangePerEra(eYield);
if (iYieldChangePerEra > 0)
{
	int iPlotEra = GetArchaeologicalRecord().m_eEra;
	int iNumEras = kPlayer.GetCurrentEra() - iPlotEra;

	if (iPlotEra != NO_ERA && iNumEras > 0)  // this would evaluate negative
	{
		iYield += (iNumEras * iYieldChangePerEra);
	}
}
```

PR #12630, opened 2026-02-14
(9-57) Hagia Sophia and Angkor Wat Swap

legenciv:
Proposal thread: https://forums.civfanatics.com/threads/9-57-hagia-sophia-and-angkor-wat-swap.701400/

Effects and tech unlock swapped in the code, instead of shuffling the art and names as written by the proposer, so that BUILDING_ANGKOR_WAT and BUILDING_HAGIA_SOPHIA keep referencing the respective wonders in the code.

Tested with IGE: base yields, Engineer rate, Prophet birth, Border Growth rate, gold cost of tile purchase, free faith buildings, missionary strength, for both wonders. As validation, production cost and number of policies requirement confirmed to work as intended. Civilopedia entries seem to be ok.


RecursiveVision:
> Effects and tech unlock swapped in the code, instead of shuffling the art and names as written by the proposer, so that BUILDING_ANGKOR_WAT and BUILDING_HAGIA_SOPHIA keep referencing the respective wonders in the code.

This is definitely the way to do it!

PR #12632, opened 2026-02-14
9-87, Conquistador -1CS, Cibola for Vanguard

Alpakinator:
As per:
https://forums.civfanatics.com/threads/9-87-spanish-conquistador-nerf.701610/

I renumbered promotion order here:[PromotionChanges.sql](https://github.com/LoneGazebo/Community-Patch-DLL/compare/master...Alpakinator:Community-Patch-DLL:9-87_Conquistador_nerf?expand=1#diff-669d9b14c088f724b43aacdd4b104a22db3e7ca20cb90f143ec26fe3fda2ea89) to keep the Cibola before Santa Maria like in other files, cause I noticed that's the standard. But idk if that was a good decision. Also, running out of promotion numbers.


azum4roll:
> Also, running out of promotion numbers.

Oh no! We may have to use repeated numbers for groups of promotions that can never be on the same unit.

PR #12633, opened 2026-02-14
(9-73) Netherlands UA (Yes, Again)

legenciv:
Proposal's thread: https://forums.civfanatics.com/threads/9-73-netherlands-ua-yes-again.701501/

Tested with IGE: yields from trade, yields on luxury turn 0, effect on monopolies on importing.


Hokath:
We'll also need to fix the AI before this one can be merged.

Hokath:
https://github.com/legenciv/Community-Patch-DLL/pull/1
Look correct?

legenciv:
Seems correct at first glance.

CvGameCoreDLL_Expansion2/CvDealAI.cpp
```diff
@@ -1381,7 +1381,7 @@ int CvDealAI::GetLuxuryResourceValue(ResourceTypes eResource, int iNumTurns, boo
 					}
 				}
 
-				iItemValue -= (iYieldBonusFromExport * OneGPTScaled) / 3;
+				iItemValue -= (iYieldBonusFromExport * OneGPTScaled);
 				if (iItemValue <= OneGPT)
 					return OneGPT;
```
azum4roll, line 1363: This condition needs to go too. Luxury value should commonly be *negative* when sold by Netherlands. Need to check whether AI (both sides) can equalize the deal by adding GPT to the Dutch side, as currently no item has negative value.

Hokath: nominate @axatin

axatin: later maybe

azum4roll: Will probably lead to people complaining about the UA again...

PR #12635, opened 2026-02-14
Fix trade route memory exhaustion in late game

JohnsterID:
## Fix trade route memory exhaustion crash in late game

This fixes a crash caused by memory exhaustion when the AI evaluates thousands of potential trade routes in late game scenarios on large maps with many civs.

### Root Cause Analysis

The original PR attempted to limit routes **after** the cache was built, but crash logs showed memory exhaustion occurring **during** cache building in `UpdateTradePathCache()`:

```
Total Path Nodes: 8,304,015,850 (8.3 billion nodes)
Memory Required @ 8 bytes/node: 62.2 GB
```

Win32 32-bit processes are limited to ~2 GB total address space. The trade path cache was storing full `SPath` objects containing vectors of thousands of `SPathNode` structs (8+ bytes each), which is fundamentally incompatible with the memory constraints.

### Solution: Metadata-Only Cache

Instead of storing full paths, we now store only **16 bytes of metadata per path**:

```cpp
struct STradePathInfo {
    int iTotalCost;
    int iNormalizedDistanceRaw;
    int iTotalTurns;
    int iPathLength;  // Just the count, NOT the path itself
};
```

**Memory impact:**
- Before: ~500K paths × ~64KB average = ~32 GB (impossible on Win32)
- After: ~500K paths × 16 bytes = ~8 MB (fits easily)

### Changes

1. **STradePathInfo struct** - Lightweight metadata extracted from SPath during cache building
2. **GetCachedTradePathInfo()** - Returns metadata pointer instead of full path
3. **ComputeTradePath()** - Computes full path on-demand when actually needed
4. **Scoring functions updated** - Use path length estimates for terrain/danger scoring

### Trade-offs

The scoring functions now use **approximations** for path-dependent factors:
- Terrain yield bonuses: Estimated based on path length and territory ratios
- Danger assessment: Estimated based on path length and diplomatic state
- Passing city yields: Estimated based on path length

The **primary scoring factors** (gold, science, culture, religion bonuses) remain **exact** since they depend only on city properties, not path nodes.

### Original Stack Trace (for reference)
```
CvTradeAI::GetAvailableTR(...) Line 5873
CvTradeAI::GetPrioritizedTradeRoutes(...) Line 7057  
CvHomelandAI::ExecuteTradeUnitMoves() Line 6010
```

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvTradeClasses.cpp
+++ b/CvGameCoreDLL_Expansion2/CvTradeClasses.cpp
@@ -207,6 +207,27 @@ bool CvGameTrade::HavePotentialTradePath(bool bWater, CvCity* pOriginCity, CvCit
 	return hasPath;
 }
 
+const SPath* CvGameTrade::GetCachedTradePath(CvCity* pOriginCity, CvCity* pDestCity, DomainTypes eDomain)
+{
+	if (!pOriginCity || !pDestCity)
+		return NULL;
+
+	bool bWater = (eDomain == DOMAIN_SEA);
+	TradePathLookup& cache = GetTradePathsCache(bWater);
+
+	int iOriginPlot = pOriginCity->plot()->GetPlotIndex();
+	int iDestPlot = pDestCity->plot()->GetPlotIndex();
+
+	TradePathLookup::iterator itOrigin = cache.find(iOriginPlot);
+	if (itOrigin != cache.end())
+	{
+		std::map<int, SPath>::iterator itDest = itOrigin->second.find(iDestPlot);
+		if (itDest != itOrigin->second.end())
+			return &(itDest->second);
+	}
+	return NULL;
+}
+
 void CvGameTrade::InvalidateTradePathCache()
 {
 	for (uint uiPlayer1 = 0; uiPlayer1 < MAX_PLAYERS; uiPlayer1++)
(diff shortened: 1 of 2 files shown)


JohnsterID:
There is more to do... https://github.com/JohnsterID/Community-Patch-DLL/commit/62cc852292d44cbb1dc8d2c1483f7ec968318760

Hokath:
So in a big map, assuming there's a period of peace with enough available trade routes, this was basically guaranteed to happen? Is it?

RecursiveVision:
While you're at it, do you mind copying over the invulnerability exceptions to danger to the AI logic for gold internal routes? I think I missed that one when I made the change.

Cyclopia31:
Is it OK for the multiplayer? I remember that there were a lot of desyncs related to trade route in older VP version. That was fixed by seroperson in 2023 if I remember well. Just to be sure !

JohnsterID:
@Hokath - Yes, this was essentially guaranteed on large maps with many civs during peaceful periods. The cache would try to store 500K+ paths, each containing 2000-8000 nodes at 8+ bytes each, requiring 30-60 GB of memory. On a 32-bit Win32 process limited to ~2 GB, this would always crash eventually.

@RecursiveVision - Added the invulnerability exception to `ScoreGoldInternalTR` in commit 0874db2. Now all three scoring functions (International, Internal, and GoldInternal) properly handle corporate franchise invulnerability.

@Cyclopia31 - The fix is deterministic and safe for multiplayer. The scoring now uses estimates based on path length rather than iterating path nodes, but:
- All random-seed-affecting operations remain unchanged
- The approximations use only deterministic factors (path length, war state, territory ratios)
- No floating point or timing-dependent calculations were introduced

vaderkos:
This should make late game multiplayer great again, awesome work @JohnsterID 🥇

JohnsterID:
The trade-offs will need to be reviewed but at least we are figuring out why some cases of memory exhaustion happen.

axatin:
Thank you very much for this! I think we can improve it even more by storing the parts of the score calculations that require exact knowledge of the path in the cache as well, then we don't need the heuristics. I'm working on it.

axatin:
Problem: In IsValidTradeRoutePath, pPath is always non-null, so this introduces a lot of additional pathfinding. Luckily we can fix this easily as IsValidTradeRoutePath only uses the normalized distance, which is already cached

axatin:
https://github.com/LoneGazebo/Community-Patch-DLL/pull/13072

PR #12645, opened 2026-02-16
Add Portugal 4UC Strategy Text

ron-murhammer:
Feedback welcome.


Alpakinator:
Don't forget to use icons. I'd add a bit stricter wording:

“Caçadores gain +1 [ICON_MOVES] Movement and become stronger as they lose HP. Their mobility makes them excellent for advancing through rough terrain, since they can move onto Hills and still fire in the same turn. Before Landships arrive, they're effective on the front line: they are more durable and hit harder while wounded, and can quickly reposition when heavily damaged.”

Alpakinator:
Or maybe "They're effective on the front line: they are more durable and hit harder while wounded, and can quickly reposition when heavily damaged. However, once Landships arrive to the battlefield, they become vulnerable.”"

azum4roll:
I thought we agreed to not use icons in strategy texts outside of yield icons and resource icons?

Alpakinator:
I don't recall that. Why would that be better?

PR #12655, opened 2026-02-17
Add Ottoman 4UC Strategy Text

ron-murhammer:
Feedback welcome.


Alpakinator:
Well written imo.

Hokath:
As a general point, I think it's good to have a description of the ways it has altered stats/requirements/etc. from the base unit (or reference unit, in this case). I would keep that separate from the special abilities, at least not within the same sentence. Here is an extreme version where the general and specific are separate in paragraphs

> Compared to the Cannon, the Great Bombard has very high [ICON_RANGE_STRENGTH] Ranged Combat Strength, a larger bonus against Cities, and a larger penalty against Units. This makes it a glass cannon which is incredible at doing damage to Cities but vulnerable to melee attackers with its fairly low [ICON_STRENGTH] Combat Strength. It synergizes well with Janisarries as they unlock at the same time and can be used to destroy enemy armies while the Great Bombard makes quick work of sieging Cities.[NEWLINE][NEWLINE]You can only have 2 Great Bombards at one time, so they need to be carefully guarded as they cannot Heal. It may be worth using damaged Bombards in dangerous, even suicidal, attacks and then quickly replacing them, making use of your Siege Foundry bonus, rather than keeping around weakened units.

Alpakinator:
True, that's better.

PR #12657, opened 2026-02-17
9-49 Manufactory and Academy Changes

KungCheops:
Add new table Improvement_ResourceExtractionIncrease(ImprovementType, ResourceType, Num) that lets an improvement increase the output of a particular resource (additive).

Add new table Improvement_ResourceExtractionMod(ImprovementType, ResourceType, Modifier) that lets an improvement increase the output of a particular resource (multiplicative).

Remove unused code related to a non-existing improvement column named ResourceExtractionMod.

The additive modifier is applied before all multiplicative modifiers (e.g. Russia's UA).

Academy now counts towards all same-improvement adjacency bonuses (replacing the manufactory).

Manufactory now gives an additional copy of any resource it is built upon.

Add builder AI support for the new improvement properties. This also adds support for Russia's UA, which was previously not considered.

The EUI tooltip will now show the actual resource output of a plot after modifiers, both for potential and current improvements. It will also show correct resource amount deltas when replacing existing improvements.

Fix a related EUI bug.

Includes text updates by Hokath.


(3a) VP - EUI Compatibility Files/LUA/PlotHelpManager.lua
```diff
@@ -1100,11 +1108,15 @@ local PlotToolTips = EUI.PlotToolTips or function( plot, isExtraTips )
...
-							elseif isResourceConnected then
-								buildTip = string_format("%s [COLOR_NEGATIVE_TEXT]%+i[ENDCOLOR]%s", buildTip, -numResource, resource.IconString )
+								newImprovementAmount = plot:GetNumResourcePostModifiers(activePlayerID, buildImprovementID)
+							end
+							if isResourceConnected then
+								oldImprovementAmount = plot:GetNumResourcePostModifiers(activePlayerID, revealedImprovementID)
+							end
+							if newImprovementAmount > oldImprovementAmount then
+								buildTip = string_format("%s [COLOR_POSITIVE_TEXT]%+i[ENDCOLOR]%s", buildTip, newImprovementAmount - oldImprovementAmount, resource.IconString )
+							elseif newImprovementAmount < oldImprovementAmount then
```
vaderkos, line 1118: What if it's equal?

KungCheops: Then we don't want to show a diff. This is the part of the tooltip that tells you what will change if you build another improvement on an existing one.

PR #12663, opened 2026-02-20
Tooltips for disabled actions

axatin:
- Moved tooltip generation for disabled actions to the DLL, preventing inconsistencies between game core logic and UI 
- Added many missing tooltips
- QoL: It is now permitted to build improvements that are constructed immediately while another unit is building a road on the same tile

(I'll squash the commits below when merging)


(1) Community Patch/Community Patch.civ5proj
azum4roll: Remapper must be last in load order. These two should be somewhere up above.

axatin: Fixed

PR #12672, opened 2026-02-22
 9-74 Repair Fleet Replaced With Passive Healing 

KungCheops:
Add new UnitPromotions integer column PassiveAoEHeal. A unit with this promotion will passively heal all units within range (2 + AuraRangeChange) for the set amount each turn.

Units with the "CannotHeal" flag will not be healed, similarly to the repair fleet behavior. This also includes resource shortage checks.

Remove the repair fleet ability from the Great Admiral and replace it with 3 passive AoE healing.

Needs healing changes from 9-30.


(2) Vox Populi/Database Changes/Triggers.sql
```diff
@@ -106,6 +106,20 @@ BEGIN
...
+WHEN (NEW.Class = 'UNITCLASS_GREAT_ADMIRAL')
+BEGIN
+    -- your modified logic here
+    UPDATE Units
+    SET
+        CanRepairFleet = 0,
+        CanChangePort = 1
+    WHERE Type = NEW.Type;
+END;
+
```
azum4roll, line 122: I don't see a need for this trigger tbf. Modders already add unique units based on their base unit.

KungCheops: We already have a trigger, this is needed to replace the one in (1) Community Patch

azum4roll: In that case, we should just delete the trigger but not replace it. The one in (1) is to keep compatibility with BNW mods, but VP modmods should always be based on VP with the split admiral columns.

KungCheops: this is the trigger in (1):

```SQL
CREATE TRIGGER Units_GlobalSeparateGreatAdmiral
AFTER INSERT ON Units
WHEN (NEW.Class = 'UNITCLASS_GREAT_ADMIRAL')
BEGIN
	UPDATE Units
	SET
		CanRepairFleet = 1,
		CanChangePort = 1
	WHERE Type = NEW.Type;
END;
```

(1) Community Patch/Database Changes/UnitPromotions/PromotionTableChanges.sql
```diff
@@ -352,3 +352,6 @@ ALTER TABLE UnitPromotions_Features ADD IgnoreTerrainCostFrom boolean DEFAULT 0;
 
 -- Additonal healing for a flat amount each turn regardless of action taken
 ALTER TABLE UnitPromotions ADD FlatHealRate integer DEFAULT 0;
+
+-- Great Admiral AoE healing ability
```
axatin: Please change the comment to something more descriptive, like "Provide +X passive healing to all friendly units of the same domain within 2 tiles"

PR #12674, opened 2026-02-22
Modernize and fully implement the Belief_YieldPerHeal table

Hokath:
This table appears to have been added for a CP-dependent mod many years ago.

In the event a unit in owned territory heals and the table entry is set to faith (and only faith), it awards yields scaling with era to the holy city/capital, and an on-tile popup similar to when you gain yields from combat experience. This logic is all within the DoHealing function in CvUnit.cpp

Here I have:
1) added a new instant yield enum for yields from healing
2) refactored the yield trigger into DoInstantYield -- this means all yields are possible
3) generalized the code so yields are awarded wherever a unit is by default, and there is a Belief column to restrict them to owned tiles
4) generalized the table to be time 100, so you can give 1 yield per 5 HP, for example.

I'd like to use this is a Celts proposal in the future.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvReligionClasses.cpp
+++ b/CvGameCoreDLL_Expansion2/CvReligionClasses.cpp
@@ -8242,7 +8242,7 @@ int CvReligionAI::ScorePantheonBeliefAtCity(CvBeliefEntry* pEntry, CvCity* pCity
 
 						int iEraNeeded = pkTechInfo->GetEra();
 						int iCurrentEra = m_pPlayer->GetCurrentEra();
-						iAvailabilityModifier = max(0, 4 - (iEraNeeded - iCurrentEra));  // lose remaining value per era we have to wait
+						iAvailabilityModifier = max(0, 3 - (iEraNeeded - iCurrentEra));  // lose remaining value if we have to wait
 					}
 				}
 			}
@@ -8517,7 +8517,7 @@ int CvReligionAI::ScorePantheonBeliefAtCity(CvBeliefEntry* pEntry, CvCity* pCity
 
 							int iEraNeeded = pkTechInfo->GetEra();
 							int iCurrentEra = m_pPlayer->GetCurrentEra();
-							iAvailabilityModifier = max(0, 4 - (iEraNeeded - iCurrentEra));  // lose remaining value per era we have to wait
+							iAvailabilityModifier = max(0, 3 - (iEraNeeded - iCurrentEra));  // lose remaining value if we have to wait
 						}
 						if (!pCity)
 						{


azum4roll:
Why add a Beliefs column instead of one in the table?

Hokath:
I thought in the future it might be used for other things, whereas I don't actually see it ever being used with this table in VP itself (I just wanted to make sure there was reverse compatibility)

azum4roll:
But it's then stuck with the whole belief, not individual abilities.

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -26340,7 +26340,7 @@ void CvPlayer::doInstantYield(InstantYieldType iType, bool bCityFaith, GreatPers
 					{
 						if (pReligion)
 						{
-							int iTempVal = pReligion->m_Beliefs.GetYieldFromFaithPurchase(eYield, GetID(), pLoopCity, true);
+							int iTempVal = pReligion->m_Beliefs.GetYieldFromFaithPurchase(eYield, GetID(), pLoopCity);
```
azum4roll, line 26343: This is definitely used (Indulgences, Gurukulam). Sounds like a bug fix (follower beliefs should work for any city adopting the religion).

azum4roll: Nvm, the flag is completely skipped by follower beliefs. You want this to be on a pantheon and every city with it as majority should benefit from it?

Hokath: That's right. The bHolyCity flag is a bit weird in that it is not applied for all beliefs (as you mention, followers ignore it)

azum4roll: Which is why follower beliefs should never have player-wide abilities.

CvGameCoreDLL_Expansion2/CvReligionClasses.cpp
```diff
@@ -7897,7 +7897,7 @@ int CvReligionAI::GetValidPlotYieldTimes100(CvBeliefEntry* pEntry, CvPlot* pPlot
 	if (eTerrain != NO_TERRAIN)
 	{
 		int iTerrainYieldChangeTimes100 = pEntry->GetTerrainYieldChange(eTerrain, iI) * 100;
-		iTerrainYieldChangeTimes100 += pEntry->GetYieldPerXTerrainTimes100(eTerrain, iI) / 3; // lower value because it's difficult to get many tiles worked
+		iTerrainYieldChangeTimes100 += pEntry->GetYieldPerXTerrainTimes100(eTerrain, iI);
```
axatin, line 7900: wait, no, this was intentional. It was not a rounding estimate (yields are already calculated correctly times100), it was to take into account the fact that cities usually don't work all the tiles of a specific terrain (as opposed to, for example, all tiles with a resource, those tiles are being worked much more often because they are usually the best tiles available). Without this the AI would always take Renewal.

azum4roll: There's probably a better formula for this. But for now, min(pop, tiles) / 2 should be closer to the actual situation than tiles / 3? Not sure how you can code this though.

PR #12680, opened 2026-02-24
Lua Things (part. 1)

schnetziomi5:
Since #12261 didn't catch on, this is the first part of an alternative approach trying to ultimately achieve the same.

In this PR:
- `RegisterScriptDate` is adjusted to restore access to some core Lua functionality that Fireaxis over-zealously hid away when they sand-boxed their scripting engine. While the sandbox is meant to stay intact (e.g. no `io` or `require`), crucially tools are provided so that it will be possible to put things into the global environments of any and all LuaContexts from within Lua. This can later be used to...

> 1) `CPK` and `VPUI` can be made available by default just like table or string 
> 2) fireaxis objects can have their API expanded(`UI`) or be completely replaced(`GameInfo`) 
> 3) Includes can be caught and rerouted to prevent doing the same file IO a hundred times over(`InstanceManager`)
> 4) ...

Also, general Lua enhancements:
- `CvLuaScopedInstance` is improved so that its "subclasses" e.g. `CvLuaCity` A) will share their proto-tables even if they have not been registered when `RegisterScriptData` was called or B) can now be registered independently of any instances existing at that time. Currently for every type (other than `Player` and `Team`), each context will get a unique copy of the corresponding proto table, which is created the first time an object of a given type is pushed, which is quite inefficient.
- `RegisterScriptData` in `CvLuaSupport` is thusly adjusted to register the proto tables for `City`, `Plot`, `Unit` and `TeamTech`, making it possible to add "secondary" Lua API (which already works for `Player` and `Team`) - e.g. 
```
--At some point at UI/Lua load time, possibly before any city exists:
LuaTypes.City.GetOwnerObj = function(city) return Players[city:GetOwner()] end
LuaTypes.City.MakeCapital = function(city) city:GetOwnerObj():SetCapital(city:GetID()) end
--This will now work:
AnyCityObjInAnyLuaContext:MakeCapital()
```

### DIFF
--- a/CvGameCoreDLL_Expansion2/Lua/CvLuaSupport.cpp
+++ b/CvGameCoreDLL_Expansion2/Lua/CvLuaSupport.cpp
@@ -82,7 +82,7 @@ void LuaSupport::RegisterScriptData(lua_State* L)
 	"rawset = G.rawset;"
 	"rawget = G.rawget;"
 	"newproxy = G.newproxy;"
-	"if G.jit then JIT = true; bitops = G.require("bit") end;"
+	"if G.jit then LUAJIT = true; bitops = G.require('bit') end;"
 
 	"FireaxisObjects={"
 		"UI = G.UI;"


CvGameCoreDLL_Expansion2/Lua/CvLuaSupport.cpp
```diff
@@ -44,13 +48,60 @@ bool luaL_optbool(lua_State* L, int idx, bool bdefault)
...
+	CvLuaTeamTech::PushTypeTable(L);
+
+	//-----------------------------------------
+	// Restore access to some core lua functionality hidden away by Fireaxis, but still keep the sandbox intact
+	// Use lua_dostring(), because it's more comfortable to code and easier to see what's happening
+
+	lua_pushvalue(L, LUA_REGISTRYINDEX);
+	lua_setglobal(L, "R");
+
+	const char* luacommand = ""
```
vaderkos, line 77: @schnetziomi5 That's very interesting, could you explain more how it works?
Also there is a question if these objects are available in states that are not in game.

vaderkos: Can we also have jit and bit modules? Only if they're available of course

schnetziomi5: > Can we also have jit and bit modules? Only if they're available of course

Sadly I have come to conclude that the only way to make Civ5 use LuaJIT is to overwrite the lua binary before the game is launched - an option which personally I have mixed feelings about. 

However if we write a set of C-functions for bitwise operations on numbers we could register them here. Hastily written example:
```
int bAnd(lua_State* L)
{
   int a = luaL_checknumber(L, 1);
   int b = luaL_checknumber(L, 2);
   lua_pushnumber(L, a & b ) ;
   return 1 ;
}
```

vaderkos: I meant, if runtime is luajit just make these functions available.
Just `jit = G.jit` or something, and of course otherwise if it can be made in C++ as fallback.

schnetziomi5: We could also add a modified version of dofile/include that offers some basic precompilation functionality. While direct performance gain from this would probably not be that big, it would make the writing of highly performant code easier by a lot.

Generally speaking I think between LuaJIT and making our lua code base more efficient, we should focus on the latter.

vaderkos: Agree :D

schnetziomi5: > @schnetziomi5 That's very interesting, could you explain more how it works? Also there is a question if these objects are available in states that are not in game.

This currently doesn't work for states that aren't ingame, sadly. I am currently contemplating if that's actually necessary and if so what's the best way to achieve that. (There are some).

If you mind telling me, what would you want an explanation on exactly?

schnetziomi5:
@vaderkos
Is it sufficient for you like this - a LUAJIT flag and the bitops module?

CvGameCoreDLL_Expansion2/Lua/CvLuaSupport.cpp
vaderkos: Do I understand this correctly, you've made the metatables `__index` of `City`, `Unit` and etc accessible via `LuaTypes` so it can be extendable globally through lua code if property doesn't exist?

vaderkos:
> @vaderkos Is it sufficient for you like this - a LUAJIT flag and the bitops module?

Looks reasonable, should it be just `bit`, but bitops is also ok

CvGameCoreDLL_Expansion2/Lua/CvLuaSupport.cpp
schnetziomi5: Yes.

Current dll's code in "pseudo" lua:
```
function push<Type>( ptr )
	
	local typetable = ENV.<Type> ; -- e.g. ENV.City
	
	if not typetable then
		typetable = { __instances = setmetatable( {} , { __mode = "v" } ) } ;
		pushMethods<Type>( typetable ) ;
		ENV.<Type> = typetable ;
	end
	
	if not typetable.__instances[ptr] then
		typetable.__instances[ptr] = setmetatable( { __instance = ptr } , { __index = typetable } ) ;
	end
	
	return typetable.__instances[ptr] ;
end
```
I spit this into two functions that work independently:
```
function pushTypeTable<Type>()
	if not ENV.LuaTypes then
		ENV.LuaTypes = {} ;
	end
	
	if not ENV.LuaTypes.<Type> then
		local typetable = {} ;
		typetable.__instance = setmetatable( {} , { __mode = "v" } ) ;
		typetable.__index = typetable ;
		pushMethods<Type>( typetable ) ;
		ENV.LuaTypes.<Type> = typetable ;
	end
		
	return ENV.LuaTypes.<Type> ;
end

function push<Type>( ptr )
	local typetable = pushTypeTable<Type>() ;
	
	if not typetable.__instances[ptr] then
		typetable.__instances[ptr] = setmetatable( { __instance = ptr } , typetable ) ;
	end
	
	return typetable.__instances[ptr] ;
end
``` 
This is so much easier to write and understand in Lua than it is with Lua's stack manipulation C-API, maybe I should have just used `dostring()` for that aswell...

vaderkos: I'd add something preventing overriding methods maybe. Are we expecting them to be defined only once? Because if we allow extensions globally then it can easily break things if multiple files try to do it.

schnetziomi5: The gamecore table is reset and created anew every time a game is started/loaded - so there is no persistence between different saves. Other than that I would rather make it the responsibility of the (mod-)modder to take care when adding/overriding stuff.

vaderkos: I'm rather worried about multiple modmods conflicting because both edit similar things

schnetziomi5: Yeah, it's not that I don't see your point at all ... but realistically, how many unique modmodders are we looking at, that would use this? Are there even five? For overriding stuff there are not that many use cases anyway, and if they  prefix their additional functions, I think the potential for conflict is pretty small?

vaderkos: Yeah, you have the point :D

PR #12681, opened 2026-02-24
 Tactical AI tuneup and fixes 

KungCheops:
AI will now be less likely to abandon border cities even if they aren't under active siege.

AI will now position ranged units such that they can actually use them, rather than just putting them at max range regardless of terrain.

Several fixes for AI healing/fleeing logic:

 * Will now prefer to stay put and heal over moving to an equally good healing plot
 * Will more strongly prefer not to embark while fleeing
 * Will now cancel movement upon discovering a unit when fleeing (and reconsider where to flee to)
 * Should be better at pillaging/performing ranged opportunity attacks while fleeing

AI exploring non-explorer units will no longer stray far away from owned cities.

Add rudimentary support for rend (AoEDamageOnMove) in danger calculations.

Workers should now always prioritize cleaning fallout over everything else for a given plot.

AI will now try to use units that give yield on kill to get kills on enemy units.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvDangerPlots.cpp
+++ b/CvGameCoreDLL_Expansion2/CvDangerPlots.cpp
@@ -874,6 +874,13 @@ int CvDangerPlotContents::GetDanger(const CvUnit* pUnit, const UnitIdContainer& 
 			//todo: consider whether the enemy units would block each other from attacking?
 			int iDamage = TacticalAIHelpers::GetSimulatedDamageFromAttackOnUnit(pUnit, pAttacker, m_pPlot, bOutOfRange ? NULL : pAttacker->plot(), iAttackerDamage, false, iExtraDamage, true);
 
+			//assume units will do AoE damage on move once per turn
+			int iAoEDamage = pAttacker->getAoEDamageOnMove();
+			if (iAoEDamage > 0)
+			{
+				iDamage += iAoEDamage;
+			}
+
 			if (!m_pPlot->IsKnownVisibleToTeam(pAttacker->getTeam()))
 				iDamage = (iDamage * 80) / 100; //there's a chance they won't spot us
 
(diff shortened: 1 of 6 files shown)


CvGameCoreDLL_Expansion2/CvDangerPlots.cpp
```diff
@@ -874,6 +874,13 @@ int CvDangerPlotContents::GetDanger(const CvUnit* pUnit, const UnitIdContainer&
 			//todo: consider whether the enemy units would block each other from attacking?
 			int iDamage = TacticalAIHelpers::GetSimulatedDamageFromAttackOnUnit(pUnit, pAttacker, m_pPlot, bOutOfRange ? NULL : pAttacker->plot(), iAttackerDamage, false, iExtraDamage, true);
 
+			//assume units will do AoE damage on move once per turn
+			int iAoEDamage = pAttacker->getAoEDamageOnMove();
+			if (iAoEDamage > 0)
+			{
+				iDamage += iDamage;
```
axatin: should be `iDamage += iAoEDamage`

KungCheops: good catch

azum4roll:
> AI will now try to use units that give yield on kill to get kills on enemy units.

What about XP on kill?

CvGameCoreDLL_Expansion2/CvTacticalAI.cpp
```diff
@@ -6007,34 +5997,33 @@ CvPlot* TacticalAIHelpers::FindSafestPlotInReach(const CvUnit* pUnit, bool bAllo
 
 			//use city distance as tiebreaker
 			if (pUnit->getDomainType() != DOMAIN_SEA)
-				iScore = iScore * 10 + iCityDistance;
+				iScore = iScore * 10 - iCityDistance;
```
axatin: why? We pick the option with the lowest score, and large distances should be penalized

KungCheops: Hmm you are right. Strange, when I tested it, it seemed to behave correctly.

PR #12687, opened 2026-03-01
(9-07a) Food/Farms/Specialists

materialcontext:
- [x] Specialist food cost now scales with era, beginning at 3 and scaling to 10 (3, 4, 5...)
- [x] Farms now receive a +1 bonus to Food at Mathematics, and +2 at Fertilizer (up from 1)
- [x] Aqueducts now give a 60% production bonus when a city gains population (up from 25%)
- [x] Text entry for specialists updated to reflect the new cost scaling. 

_Implementation Notes_:
- The original specialist cost calculation was entirely hard coded. I've replaced it with an eras table entry for specialist costs that be adjusted easily in the future. This was Recursive's advice and it seemed sensible.
- As far as I can tell the adjustment to Aqueduct is passed on to Harappan Reservoir automatically.
- C++ is not my primary language, so I may have screwed up the null pointer check, but I think this is correct. Let me know if not.

Resolution to: [(9-07a) Food/Farms/Specialists](https://forums.civfanatics.com/threads/9-07a-food-farm-specialist-tweaks-alt.701505/)


axatin:
(1) If a mod adds a new era and doesn't set a value for SpecialistFoodCost, it defaults to 1. That's not ideal. I think it would be better to do the following: If getSpecialistFoodCost is 0 for the current era, use the value from the latest previous era with a non-zero value instead. Resort to 1 only if no such value exists.

(2) Now that the values are defined directly in the database we can remove the cap `iFoodPerSpec = min(iFoodPerSpec, 10) * 100;` 

(3) We can remove the MOD_BALANCE_VP guard and use a fixed value of 2 for all eras in Community Patch

materialcontext:
Will make changes today. I hadn't even considered the modded era situation, good catch.

Hokath:
It will happen in Enlightenment Era.
This latest-previous condition would be good, then I can just do 
UPDATE Eras SET SpecialistFoodCost = SpecialistFoodCost + 1 WHERE ID > 3; 
Or similar

axatin:
continued in #12720

PR #12702, opened 2026-03-07
(9-03c) Teocalli/Order Ability Swaps + Chichen Itza Nerf (?)

Hokath:
Following the existing Global version of this table, made a separate YieldFromVictory and YieldFromVictoryEraScaling containers.
(n.b. the "EraScaling" one is the old functionality)

EraScaling defaults to true for reverse compatibility, but I did update Greece.sql just to show it explicitly.

Updated tooltip generation to handle new table, including manual belief tooltips.

### DIFF
--- a/(2) Vox Populi/Database Changes/City/Buildings/BuildingChanges2.sql
+++ b/(2) Vox Populi/Database Changes/City/Buildings/BuildingChanges2.sql
@@ -204,7 +204,7 @@ VALUES
 INSERT INTO Building_YieldFromVictoryGlobal
 	(BuildingType, YieldType, Yield, IsEraScaling)
 VALUES
-	('BUILDING_TEOCALLI', 'YIELD_FAITH', 5, 0);
+	('BUILDING_TEOCALLI', 'YIELD_FAITH', 2, 0);
 
 INSERT INTO Building_YieldFromVictory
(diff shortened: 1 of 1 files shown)


Hokath:
I note that the local and the global notifications for the Teocalli appear separately.
This is the same thing that happens for Greek Gymnasion

(2) Vox Populi/Database Changes/City/Buildings/BuildingChanges2.sql
```diff
@@ -202,16 +202,21 @@ VALUES
 	('BUILDING_TEOCALLI', 'YIELD_FAITH', 2);
 
 INSERT INTO Building_YieldFromVictoryGlobal
-	(BuildingType, YieldType, Yield)
+	(BuildingType, YieldType, Yield, IsEraScaling)
+VALUES
+	('BUILDING_TEOCALLI', 'YIELD_FAITH', 5, 0);
```
axatin: should be 2, not 5

Hokath:
Good catch. I must have squashed in the wrong order as the value is 2 locally... woops!

azum4roll:
> I note that the local and the global notifications for the Teocalli appear separately. This is the same thing that happens for Greek Gymnasion

Not much we can do. Notifications of instant yields during your turn are shown per function call.

PR #12714, opened 2026-03-10
Congress 9

azum4roll:


(1) Community Patch/Core Files/Overrides/Includes/InfoTooltipInclude.lua
```diff
@@ -121,26 +121,23 @@ local function AppendEraScaling(strTooltip)
...
---- e.g. {"a"} -> "a", {"a", "b"} -> "a and b", {"a", "b", "c"} -> "a, b and c"
+--- e.g. {"a"} -> "a", {"a", "b"} -> "a and b", {"a", "b", "c"} -> "a, b, and c"
 --- @param tItems string[]
 --- @return string
-local strListSeparator = L("TXT_KEY_LIST_SEPARATOR") .. " ";
-local strListAnd = " " .. L("TXT_KEY_LIST_AND") .. " ";
 local function ConcatWithCommaAnd(tItems)
 	local n = #tItems;
-	if n == 1 then
+	if n == 0 then
```
vaderkos, line 129: Is this ever used for long lists? If yes better to use StringBuilder, if not it's fine as wine

azum4roll: I didn't write this. But practically the list shouldn't be too long. Unless someone feels like making a building boost every other building or something.

PR #12715, opened 2026-03-10
Merge 4UC credits

Hokath:
I'm not aware of this information being added to the repo during integration?

### DIFF
--- a/(1) Community Patch/Credits.txt
+++ b/(1) Community Patch/Credits.txt
@@ -108,9 +108,6 @@ Things borrowed from other mods:
 		- KevinJK - idea and picture of Lamassu Gate UB.
 
 	Austria:
-		- JFD - idea of Landwehr UU,
-		- janboruta - picture and flag of Landwehr,
-		- bernie14 - 3D model of Landwehr.
 
 	Aztec:
 		- mikeburnfire - idea, picture and 3D model of Eagle UU,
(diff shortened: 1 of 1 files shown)


(1) Community Patch/Credits.txt
```diff
@@ -128,7 +66,251 @@ Big Wave : Lord on game-icons.net
...
+	Huns:
+		- JFD - idea of Tarkhan UU (Alan Cavalry)
+		- janboruta - picture and flag of Tarkhan,
+		- danrell - 3D model of Tarkhan,
+		- Wodhann - Picture of Alti Cur UB (Yurt).
+		  
+	Inca:
+		- Leugi - idea, picture and flag of Chasqui UU and Coca UR,
+		- JTitan - 3D model of Chasqui,
+
```
azum4roll, line 182: But we have neither of these. And many other original 4UC stuff.

Hokath: We have them in the files (not sure about Chasqui model, but the icons we do)

Hokath:
But yes, this is probably not complete, we should aim to make sure the credits include all assets we've used, even if we have permission in private communications.

(1) Community Patch/Credits.txt
```diff
@@ -284,6 +275,7 @@ Things borrowed from other mods:
 
 	Siam:
 		- snafusmith - 3D model of Seir Morb UU (155 mm gun).
+        - Putmalk - Seir Morb icon
```
azum4roll, line 278: It's called Suea Mop now

azum4roll:
Can you also find the ones added in 2UC? Cuirassiers etc.

vaderkos:
What about having changelog folder, and having credits and changes inside. Something like

```
Changelog/v5.1.4_2026_10_10.md
Changelog/v5.1.4_2026_10_10.md
and etc.
```

PR #12717, opened 2026-03-10
Fix database errors

axatin:
and remove everything related to asamu


Hokath:
@KungCheops told me we've both added the
UPDATE Defines SET Value = 50 WHERE Name = 'NONCOMBAT_UNIT_RANGED_DAMAGE';
in DefineChanges.sql

Could you remove one?

axatin:
> @KungCheops told me we've both added the UPDATE Defines SET Value = 50 WHERE Name = 'NONCOMBAT_UNIT_RANGED_DAMAGE'; in DefineChanges.sql
> 
> Could you remove one?

Done

PR #12731, opened 2026-03-13
Bugfixes

axatin:


CvGameCoreDLL_Expansion2/CvDiplomacyAI.cpp
```diff
@@ -11414,7 +11414,7 @@ int CvDiplomacyAI::ComputeDynamicStrengthModifier(PlayerTypes ePlayer, PlayerTyp
 	if (!GET_PLAYER(ePlayer).isMajorCiv())
 		return 0;
 
-	int iModifier = 0;
+	int iModifier = 100;
```
RecursiveVision, line 11417: @axatin Good catch. The line just above should also be returning iStrength, not 0 (as in ComputeRatingStrengthAdjustment).

axatin: Fixed

azum4roll:
But buttons also aren't always responsive.

axatin:
for some reason it works much better with a button

PR #12741, opened 2026-03-15
Fix non-mercantile civs giving happiness

axatin:


azum4roll:
Was this the case pre-rework?

axatin:
No, the two functions were only used in GetCurrentHappinessFlatBonus, which had the trait guard at the beginning.

Hokath:
Aha I see, so the tooltip generator needed a guard too.
I was confused at first because I wasn't seeing extra happiness in my game

PR #12745, opened 2026-03-16
Possible fix for IsBeliefValid crash when city doesn't have any religious presence

KungCheops:
Maybe there's another way to fix this that solves the root cause, but this at least removes the crash


axatin:
pCity->GetCityReligions().m_pCity should always be pCity. if it doesn't, we need to investigate and fix the root cause, so it's better to let it crash

PR #12762, opened 2026-03-19
Units will no longer stop mid-move if that would break unit stacking rules

KungCheops:
The diff got kinda messed up, but it's just one new if-statement and one level of indentation of the code.

Not sure if this counts as a balance change or bug-fix, but I think it's a bit stupid that you can break the unit stacking rules. This has become more apparent now that exploring units will halt whenever they discover new tiles (almost every move).

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvUnit.cpp
+++ b/CvGameCoreDLL_Expansion2/CvUnit.cpp
@@ -30288,17 +30288,46 @@ int CvUnit::UnitPathTo(int iX, int iY, int iFlags)
 		}
 	}
 
-	//hack ...
-	SetSpottedEnemy(false);
-	SetSpottedRuin(false);
-	SetRevealedPlot(false);
+	// If we are mid move and can't stop, don't reset the flags. We want to still stop when available.
+	if (CountStackingUnitsAtPlot(plot()) < plot()->getUnitLimit())
+	{
+		SetSpottedEnemy(false);
+		SetSpottedRuin(false);
+		SetRevealedPlot(false);
+	}
 
 	//todo: consider movement flags here. especially turn destination, not only path destination
 	bool bMoved = UnitMove(pPathPlot, IsCombatUnit(), NULL, bDone);
 
-	if (HasSpottedEnemy())
+	if (CountStackingUnitsAtPlot(pPathPlot) < pPathPlot->getUnitLimit())
 	{
-		if (iFlags & CvUnit::MOVEFLAG_ABORT_IF_NEW_ENEMY_REVEALED)
+		if (HasSpottedEnemy())
+		{
+			if (iFlags & CvUnit::MOVEFLAG_ABORT_IF_NEW_ENEMY_REVEALED)
+			{
+				ClearPathCache();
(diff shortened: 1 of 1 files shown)


azum4roll:
But you already can't end your turn with stacked units. It's also already guaranteed that the unit has remaining moves (outside of Lua shenanigans) to move off the stack, so no softlock problems.

KungCheops:
> But you already can't end your turn with stacked units. It's also already guaranteed that the unit has remaining moves (outside of Lua shenanigans) to move off the stack, so no softlock problems.

As I wrote it's to prevent cheating, essentially.

azum4roll:
While it makes sense, wouldn't it create the scenario where a unit ends up with no moves and stacked? If the entire path past the first stacked unit is filled with units.

KungCheops:
> While it makes sense, wouldn't it create the scenario where a unit ends up with no moves and stacked? If the entire path past the first stacked unit is filled with units.

The game doesn't let you queue a move where you don't have vision on the first end turn tile.

KungCheops:
I need to think about where the flags should be reset though. This current one forgets to cancel the move later unless you get another cancel trigger.

azum4roll:
> > While it makes sense, wouldn't it create the scenario where a unit ends up with no moves and stacked? If the entire path past the first stacked unit is filled with units.
> 
> The game doesn't let you queue a move where you don't have vision on the first end turn tile.

I guess this only applies to AI/auto movement then.

KungCheops:
> > > While it makes sense, wouldn't it create the scenario where a unit ends up with no moves and stacked? If the entire path past the first stacked unit is filled with units.
> > 
> > 
> > The game doesn't let you queue a move where you don't have vision on the first end turn tile.
> 
> I guess this only applies to AI/auto movement then.

Have you tried it?

KungCheops:
It could be a mod option potentially, but I know we've been trying to cut down on those

RecursiveVision:
> It could be a mod option potentially, but I know we've been trying to cut down on those

Trying to cut down on ones that have no purpose or can't be cleanly toggled on/off. Otherwise it's fine.

azum4roll:
> > > > While it makes sense, wouldn't it create the scenario where a unit ends up with no moves and stacked? If the entire path past the first stacked unit is filled with units.
> > > 
> > > 
> > > The game doesn't let you queue a move where you don't have vision on the first end turn tile.
> > 
> > 
> > I guess this only applies to AI/auto movement then.
> 
> Have you tried it?

Screenshot
Does this count?

KungCheops:
> > > > > While it makes sense, wouldn't it create the scenario where a unit ends up with no moves and stacked? If the entire path past the first stacked unit is filled with units.
> > > > 
> > > > 
> > > > The game doesn't let you queue a move where you don't have vision on the first end turn tile.
> > > 
> > > 
> > > I guess this only applies to AI/auto movement then.
> > 
> > 
> > Have you tried it?
> 
> Screenshot Does this count?

I meant if there's a unit you can't stack with on the tile you're moving through.

KungCheops:
I.e., if there's a unit on the visible desert hill

PR #12764, opened 2026-03-19
Various AI fixes

KungCheops:
Exploring units will not heal and will try not to end their turn in non-friendly city state territory.

Improve AI danger calculations for exploring units.

Performance improvement for explorers, skip an unnecessary pathfinding call.

Fix incorrect distance calculation when evaluating exploration plots.


azum4roll:
Can they still pass through? With rules like early ships going through oceans.

KungCheops:
> Can they still pass through? With rules like early ships going through oceans.

They can

PR #12766, opened 2026-03-20
Let modded Inca still settle on Mountains

Hokath:
I was missing this update of the new civilian-only mountain passing in the SiteEvaluation method that gets called to decide if Founding is an allowed action (this is the one that gets overridden if there is a road on the tile)

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvSiteEvaluationClasses.cpp
+++ b/CvGameCoreDLL_Expansion2/CvSiteEvaluationClasses.cpp
@@ -80,7 +80,7 @@ bool CvCitySiteEvaluator::CanFoundCity(const CvPlot* pPlot, const CvPlayer* pPla
 
 	if(!pPlot->isValidMovePlot(pPlayer ? pPlayer->GetID() : NO_PLAYER ))
 	{
-		if (!pPlot->IsMountain() || !pPlayer->WorkersMountainPass())
+		if (!pPlot->isMountain() || !pPlayer->WorkersMountainPass())
 			return false;
 	}
 


azum4roll:
This is awkward. The existing code doesn't factor in any unit-specific movement when deciding whether one can found on a plot, so this is something that can never be enabled by promotions. The only solution is your hacky WorkersMountainPass but that's linked to workers, not settlers. What if some modder decides to allow settlers to cross mountains but not workers? Or the promotion comes from policies/beliefs/buildings/events/natural wonders etc.?

Hokath:
Yes WorkersMountainPass is a bit hacky. Better would be CivilianMountainPass and best probably UnitClassMountainPass?

azum4roll:
I dunno. Is there a reason we must skip site evaluation on invalid move plots? What if we skip that check and then check it at whatever's calling CvPlayer::getPlotFoundValue() while a unit is present?

A list of functions that need modified:
- CvAIOperationCivilianFoundCity::PerformMission()
- CvAIOperationCivilianFoundCity::FindBestTargetForUnit()
- CvHomelandAI::ExecuteFirstTurnSettlerMoves()
- CvHomelandAI::ExecuteOpportunisticSettlementMoves()
- CvLuaPlayer::lGetRecommendedFoundCityPlots()

There are also a few CvEconomicAI functions that call `CvPlayer::HaveGoodSettlePlot()`, but it's rare that the only good settle plots are mountains so we should be able to safely ignore it.

Hokath:
Let me start a new branch for that.
It would be good to have this fix temporarily for the playtest

Hokath:
- CvAIOperationCivilianFoundCity::PerformMission()
This one calls the unit method
`if(pSettler && pSettler->canFoundCity(pCityPlot) && pSettler->plot() == pCityPlot && pSettler->canMove())`
which only calls the player method (and the site eval via that) if the plot can be moved into
```if (pPlot)
	{
		if (!canMoveInto(*pPlot, CvUnit::MOVEFLAG_DESTINATION))
			return false;

		if (!GET_PLAYER(getOwner()).canFoundCityExt(pPlot->getX(), pPlot->getY(), bIgnoreDistanceToExistingCities, bIgnoreHappiness, toolTipSink))
			return false;
	}
```

So I don't think this one needs changing?

- CvAIOperationCivilianFoundCity::FindBestTargetForUnit
This one gets the plot it will evaluate as 
`CvPlot* pNewTarget = GET_PLAYER(m_eOwner).GetBestSettlePlot(pUnit, this);`
which does a loop where it exits if it hits the condition
`if(pUnit && (!pUnit->canFoundCity(pPlot) || !pUnit->canMoveInto(*pPlot)))`
And so that doesn't need to change either?
(and this check !pUnit->CanMoveInto is redundant?)

Hokath:
Both the HomelandAI methods go through the Unit call as well.

The LuaPlayer one doesn't, it has the check like
```
// Can't actually found here!
if(!pkPlayer->canFoundCity(iPlotX, iPlotY))
    continue;
```
so we need to change that to 
```
// Can't actually found here!
if(!pFoundingUnit->canFoundCity(pPlot))
    continue;
```

Issue #12773, opened 2026-03-21
5.2_Multiplayer_***Plot Unit Desync*** and Passing Null error


Cyclopia31:
### 1. Mod version

5.2

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

None, Modpack: https://drive.google.com/file/d/1KH-owPTl_pjGM2KJOAflABzGOTROXqyK/view?usp=drive_link

### 4. Describe the Issue

I am playing a game with two human players. 
I decided to start the game in medieval era so players start with 2 settlers each.
When a human player or an AI player use one of their settler I am getting these messages:

<img width="260" height="81" alt="Image" src="https://github.com/user-attachments/assets/e5cda84d-f6ad-4aa7-9c4f-5ac719d7219a" />

So it happened the first turn.

On the second turn during the AI turn I got also a lot of messages, some look like the first issue, and also a "Passing NULL":

<img width="253" height="83" alt="Image" src="https://github.com/user-attachments/assets/5baf9893-a9a4-4e50-b859-d9f2346304c1" />

You can just play some turns to get them if you want to repeat the errors.

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

Here the logs and saves from both players: 

[5.2_MultiplayerTest.zip](https://github.com/user-attachments/files/26157635/5.2_MultiplayerTest.zip)

Thanks a lot !

Cyclopia31:
I continued playing for a while, and it seems the error "*** PLOT UNIT DESYNC *** removeUnit: unit" consistently triggers whenever a unit is consumed or killed. I encountered this error while using settlers, workboats, killing units, deleting (and presumably losing) units.

Cyclopia31:
Were these issues "real" or false positives? I wonder if they are related to this one: https://github.com/LoneGazebo/Community-Patch-DLL/issues/12595 !

axatin:
I added the logging for issues like the one you linked. The messages you reported here were harmless, though.

Cyclopia31:
@axatin The problem is that players are spammed every turn with these "removeUnit: unit" messages, and I'm afraid that "real" errors will be drowned out by these messages (the error can fire like 60 times per AI turn). Do you think it is possible to hide only the error relative to "removeUnit: unit" while keeping the new logging for other potential errors ? Still the case in 5.2.1.

axatin:
Oh, I forgot to include the fix that removed them

[...]

Issue #12759, opened 2026-03-19
Hard Game CTD


Stalker012345:
### 1. Mod version

5.2

### 2. Installed Components

Vox Populi with EUI

### 4. Describe the Issue

Hard crash at the end of the turn.

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[Save_Game.zip](https://github.com/user-attachments/files/26105905/Save_Game.zip)

### 6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

[Logs.zip](https://github.com/user-attachments/files/26105906/Logs.zip)

Stalker012345:
I tried rerunning one more turn behind, and this time generated this error:

<img width="697" height="547" alt="Image" src="https://github.com/user-attachments/assets/4d646135-3b78-4d9a-b3b5-736d820f058f" />

schnetziomi5:
Aww. I found the problem too, but you were faster making a fix, Axatin :)
The string one I either can't reproduce from any of the posted savegames or, worse, the post savegames crash during load, which I interpret as a malformed save possibly caused by corrupted memory??

Issue #12758, opened 2026-03-18
Heap Corruption


ArefSadeghi:
### 1. Mod version

5.2

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

Milae's WASD

### 4. Describe the Issue

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/f9c04fa4-1f27-430b-bbb4-97a8894dabfb" />

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[AutoSave_Post_0072 BC-2920.zip](https://github.com/user-attachments/files/26101163/AutoSave_Post_0072.BC-2920.zip)

### 6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

[Logs.zip](https://github.com/user-attachments/files/26101190/Logs.zip)

axatin:
I can't reproduce this. @ArefSadeghi Can you try if you get the error again when loading the save file?

m4dr0b0t:
i have the same exact issue, do i need to creat a "new issue"post ?

axatin:
Not necessary, I assume it's always the same cause

Issue #12755, opened 2026-03-18
Assertion failed, no crash


Iexhu:
### 1. Mod version

5.2

### 2. Installed Components

Vox Populi with EUI (43 Civ version)

### 3. Additional mods used

Extra mods; quick turns and ethnic units

### 4. Describe the Issue

![Image](https://github.com/user-attachments/assets/dc79d7af-7a74-40b3-8c8c-b6bc091e8871)

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[Was.zip](https://github.com/user-attachments/files/26083857/Was.zip)

### 6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

[Logs.zip](https://github.com/user-attachments/files/26083943/Logs.zip)

bsesar:
Already reported in #12751.

PR #12775, opened 2026-03-22
Fix malformed CvString declaration ('passing NULL to std:string' assert)

schnetziomi5:
The compiler used to generate the release .dlls interpreted `CvString strName = NULL` as `CvString strName = CvString((char*)NULL)` for the 5.1.4 version, for some reason breaking with the previous versions interpretation as `CvString strName = CvString((int)NULL)`. Github's builds and potentially local ones were unaffected.

`CvString strName = NULL` is not properly defined and should just be `CvString strName;`

I used notepads file search function with the regular expression search mask: 
`CvString\s+([a-zA-Z0-9_]*)\s*=\s*NULL`
but it seems CvUnit:616 is the only location where such a construct was used.

Should fix the "passing NULL to std:string" assert popup (but nothing else) reported in these issues:
#12751 
#12773 
#12759 
#12758 
#12755


schnetziomi5:
Regretfully I forgot to mention the crucial help of @Matrx50 on discord, who patiently followed my instructions to generate a debug call stack. Many thanks!

PR #12777, opened 2026-03-22
New policy column FranchisePressure

Hokath:
This was an old idea for an ideology tenet, mirroring the Inquisitor/Underground Sects effect but for Corporations.
Will try to get it working and into Devotion :D


Hokath:
image

success!

PR #12779, opened 2026-03-22
Inca DLL continued

Hokath:
continued from #12766


Hokath:
First thing is making TERRAIN_HILL valid for the Building_YieldPerXTerrainTimes100 table.
image

Second is implementing the change suggested by Azum to skip the isValidMovePlot check in the SiteEvaluation that we changed in the previous commit. Instead, we make sure site evaluation always gets called from the settler class unit itself, which does the pathfinder check. 
This avoids the Hack, and I guess is a microoptimization too because we don't have to check the plot twice.

CvGameCoreDLL_Expansion2/CvCityCitizens.cpp
```diff
@@ -2316,7 +2316,11 @@ void CvCityCitizens::SetWorkingPlot(CvPlot* pPlot, bool bNewValue, CvCity::eUpda
 				if (pPlot->getTerrainType() != NO_TERRAIN)
 				{
 					GetCity()->ChangeNumTerrainWorked(pPlot->getTerrainType(), 1);
-					if (pPlot->getFeatureType() == NO_FEATURE && !pPlot->isHills())
+					if (pPlot->isHills())
+					{
+						GetCity()->ChangeNumTerrainWorked(TERRAIN_HILL, 1);
+					}
+					else if (pPlot->getFeatureType() == NO_FEATURE)
```
azum4roll, line 2323: Hills are still featureless.

azum4roll:
So where are the functions other than CvLuaPlayer?

EDIT: Ok I read your comment on the other PR and decide to trust you on that.

Hokath:
I'd now like to get the city connection/trade route bonus working on hills to give an alternative to 0 maintenance.
Atm the table I need to use also has a really weird era scaling implemented.
While looking, I found the IsOddEraScaler is hardcoded and doesn't actually work with Enlightenment Era, so need to fix that too.

azum4roll:
You have to edit the comment for the trait column too if you changed it.

Hokath:
Yes will update the xml
I am wondering if something like Traits->GetEraScaling(pPlayer) would be a good refactoring

CvGameCoreDLL_Expansion2/CvTraitClasses.cpp
```diff
@@ -5759,17 +5759,17 @@ int CvPlayerTraits::CurrentEraScalingModifier() const
 {
 	int iModifier = 100;
 	
-	if (!m_pTraits->IsEraScaling())
+	if (!IsEraScaling())
```
azum4roll, line 5762: It IS weird that these scaling properties are on player traits which can be multiple traits.

Hokath:
Not quite working as intended, but in testing I am also reminded of the bug where this trait table makes all tiles appear to have the bonus if they are unclaimed. Should fix that too while we are here.

Hokath:
@RecursiveVision you won the prize of implementing the new Inca UB
Here, the food/hills effect is implemented for you to use

azum4roll:
@RecursiveVision also needs to turn the Chasqui Training promotion into a trait bonus or set LostOnGifting to true (even though none of the units can be gifted).

Issue #12796, opened 2026-03-25
Royal Guardhouse describtion: which is right?


Timur512:
### 1. Mod version

5.2.1

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

no

### 4. Describe the Issue

this
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/57748f89-03b3-4eea-a178-76b72068d0a8" />

or this

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/954c0e84-22d9-4087-a991-a6a8eefc22ed" />

What it really gives? +20% Military supply from population in Capital, or +10% in all cities?

Hokath:
The tooltip on the city screen is what the building actually does (it is generated by reading the database).

But whether that's the intended effect?
It seems it was supposed to change with the supply rework:
https://forums.civfanatics.com/threads/7-vt-supply-cap-rework.689244/

But for some reason it didn't. Will fix.

Timur512:
ok thx

PR #12798, opened 2026-03-26
Fix Royal Guardhouse supply effect

Hokath:
https://forums.civfanatics.com/threads/7-vt-supply-cap-rework.689244/ Policy text was updated but xml fell behind.

Note: I don't think this will actually accomplish what the OP says in that thread. 20% in capital is weaker than 10% in all cities, except very early. So this will be a buff to Tradition supply in late game (?)


Hokath:
Fix #12796

PR #12800, opened 2026-03-26
Increasing the impact of flavor values on production AI

CIVITAS-John:
This pull request starts from a discussion in the Discord, where @redrum68 pointed to me that flavor numbers currently have less impact to production AI than it should have. 

After thinking about it (and working on Claude Code), my solution is to change the `sqrt(flavor * xml value)` to `xml value * sqrt(flavor)`. I think this change largely keep the original logic intact, while enabling flavors to steer VPAI at a reasonable level. 

Another change is to enable negative flavor numbers having an impact in production AI. Many built-in strategies have negative numbers set, it would have made no sense to consider them only an on/off switch as the original behavior.

I am testing those changes in Vox Deorum.


CIVITAS-John:
Added a fix to prevent crashes due to negative weight. Still testing...

CIVITAS-John:
It is working. I will run a few dozen of auto-games and send out the balance report.

CvGameCoreDLL_Expansion2/CvProcessProductionAI.cpp
```diff
@@ -79,6 +79,10 @@ void CvProcessProductionAI::AddFlavorWeights(FlavorTypes eFlavor, int iWeight)
 	if (iWeight==0)
 		return;
 
+	// Apply sqrti per-accumulator before XML multiplication to preserve XML flavor differentiation.
+	// sqrti uses abs() internally so we must preserve sign explicitly.
+	int iModifiedWeight = (iWeight > 0 ? 1 : -1) * sqrti(10 * abs(iWeight));
```
azum4roll, line 84: Then you don't need to abs the iWeight?

KungCheops: > Then you don't need to abs the iWeight?

sqrt is only defined for positive numbers

azum4roll: But you're using sqrti there

CIVITAS-John: Yes, @azum4roll is correct. Is there anything else I need to change?

RecursiveVision:
Heads up @CIVITAS-John, I've just pushed a PR that clamps personality flavors between 1 and 10. (Not counting any of the grand strategy code or whatnot that modifies flavors - just the base personality value.)

However, there is compatibility in the flavor manager for up to 100, or more if you change the max value (it's a bit arbitrary, but I doubt anyone would need more granularity than a 0-100% scale at the personality level), and changing it will no longer affect diplomacy negatively.

CIVITAS-John:
@RecursiveVision Thanks for your heads up! VD's personality value is cramped at 0-10 already. Can you point me to the commit so I can double check?

RecursiveVision:
@CIVITAS-John https://github.com/LoneGazebo/Community-Patch-DLL/commit/546c3f2b69948240e651afd0ace699cc54617dce

axatin:
Increasing the impact of building flavors is exactly the opposite of what I think we should do. That flavor table is nothing but a crude mapping of the building abilities to flavor values, with a lot of information being lost in the process. It's also a constant source of errors: building abilities change but flavors aren't updated

Going forward, I would like this table to be emptied and production AI to be changed to evaluate building abilities directly. The table (and its impact on production AI) should stay only for effects from modmods coded with lua, which can't be evaluated in the DLL.

CIVITAS-John:
Hi!

I think the problem is two-fold; we need a micro-level replacement of
flavor, but the macro level should persist, since the production AI should
have a way to understand what’s the global priority instead of always
optimizing at the local level.

axatin:
I agree, global priorities should be taken into account

CIVITAS-John:
So I wonder if this pull request should be integrated or any changes are
needed

Alpakinator:
Fwiw I think it shouldn't even as a short term thing before a flavorless solution is figured out, because this PR will introduce new ai production problems. It's cause this pr isnt extensively tested on tens of ai only games, so we don't know its consequences on ai decisionmaking. assuming that building flavors are more sane than the modofications that c++ does to these flavors is not a good idea.

axatin:
Yeah I tend to agree with that. The impact of flavors was intentionally reduced because the flavor/strategy system didn't work well. Before making flavors more impactful again we'd have to look into city/player strategies, and we'd also need to test all of that very well. But then again, production AI needs to be completely reworked anyway, so it's better to do it all at once as part of a large overhaul and move away from using flavors in the process.

CIVITAS-John:
Didn't realize that was intentional :)
I did run a lot of tests with this PR, either VPAI self-play or with LLMs. Any data that could help you make the decision?

RecursiveVision:
I'm in favor of keeping the player-level flavors only as hints to the AI about how much they should value various yields or game mechanics. The non-diplo flavor randomization per game should likely be removed.

I think flavors on anything else should be removed (except for scoring modmod abilities, as axatin said).

CIVITAS-John:
That does make sense. I also think it makes sense for AI to recognize if
they need an offensive army, a defensive one, etc, and how much effort to
spend building them. Also how much they should be expanding or
consolidating…

PR #12804, opened 2026-03-28
Clarify ETR boost tooltips

Hokath:
We talked about this in 5.0 but we've had another confused post https://forums.civfanatics.com/threads/custom-houses-markets-and-others.702514/

I think this leaves no room for misunderstanding, but perhaps there is a better wording (for one or both)

### DIFF
--- a/(1) Community Patch/Database Changes/Text/en_US/UI/CoreNewUIText.xml
+++ b/(1) Community Patch/Database Changes/Text/en_US/UI/CoreNewUIText.xml
@@ -1067,10 +1067,10 @@
 			<Text>{2_Sign}{1_Num} [ICON_GOLD] Gold to [ICON_CARGO_SHIP] Sea International Trade Routes originated from this City</Text>
 		</Row>
 		<Row Tag="TXT_KEY_PRODUCTION_BUILDING_ETR_DESTINATION_OUR_GOLD">
-			<Text>{2_Sign}{1_Num} [ICON_GOLD] Gold to Destination City for [ICON_INTERNATIONAL_TRADE] International Trade Routes targeting this City</Text>
+			<Text>{2_Sign}{1_Num} base [ICON_GOLD] Gold for the Destination City in [ICON_INTERNATIONAL_TRADE] International Trade Routes targeting this City</Text>
 		</Row>
 		<Row Tag="TXT_KEY_PRODUCTION_BUILDING_ETR_DESTINATION_THEIR_GOLD">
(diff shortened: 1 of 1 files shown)


(1) Community Patch/Database Changes/Text/en_US/UI/CoreNewUIText.xml
```diff
@@ -1067,10 +1067,10 @@
 			<Text>{2_Sign}{1_Num} [ICON_GOLD] Gold to [ICON_CARGO_SHIP] Sea International Trade Routes originated from this City</Text>
 		</Row>
 		<Row Tag="TXT_KEY_PRODUCTION_BUILDING_ETR_DESTINATION_OUR_GOLD">
-			<Text>{2_Sign}{1_Num} [ICON_GOLD] Gold to [ICON_INTERNATIONAL_TRADE] International Trade Routes targeting this City</Text>
+			<Text>{2_Sign}{1_Num} [ICON_GOLD] Gold to Destination City for [ICON_INTERNATIONAL_TRADE] International Trade Routes targeting this City</Text>
 		</Row>
 		<Row Tag="TXT_KEY_PRODUCTION_BUILDING_ETR_DESTINATION_THEIR_GOLD">
-			<Text>{2_Sign}{1_Num} [ICON_GOLD] Gold to Origin City of [ICON_INTERNATIONAL_TRADE] International Trade Routes targeting this City</Text>
+			<Text>{2_Sign}{1_Num} [ICON_GOLD] Gold to Origin City for [ICON_INTERNATIONAL_TRADE] International Trade Routes targeting this City</Text>
 		</Row>
```
vaderkos, line 1074: For me it's still confusing, what about maybe?

```
{2_Sign}{1_Num} [ICON_GOLD] Gold in this city for each incoming [ICON_INTERNATIONAL_TRADE] International Trade Route
{2_Sign}{1_Num} [ICON_GOLD] Gold in the city that sends [ICON_INTERNATIONAL_TRADE] International Trade Route here
```

Hokath: Unfortunately that's not how it works. The gold goes "to the Trade Route" as part of the calculation. So you don't actually get {1_Num}, but something affected by modifiers

Hokath:
Ok so "to Origin City" is also misleading because it sounds like yields go to the City directly.

azum4roll:
I'm fine with that

PR #12808, opened 2026-03-28
Update CvDiplomacyAI.cpp

maprezentalok:
Shouldnt there be sign checks (maybe with a helper returning -1 or 1 to multiply the bonuses with) for all vApproachScores ? Or using abs values for right value vApproachScores would be clearer.

Like edge case: ressurect a 1 city civ, then next turn lose their capital again, and in next turn resurrect again? Current logic I think would make them really pissed off and unfriendly.


maprezentalok:
ah its vApproachBias sorry

PR #12811, opened 2026-03-28
Sealift

Hokath:
In short, this is a clone of Airlift (minus any improvement hooks, at the moment) with an additional logic to check if the two cities are directly connected by a water route. 
(I envisage the column on Harbor or Seaport)

There's a lot of places Airlift appears and the lua stuff is especially hard to understand so this might take a while.


(1) Community Patch/Database Changes/Misc/NewInterfaceModes.sql
azum4roll, line 1: Ew. XML please.

Hokath:
Ok so atm this has demoted ranged attack to... airstrike
image

Must be how I edited CvEnums.h, that this is called by its hardcoded value?

Hokath:
Here's the current status

Sealift = 1 on the Harbor. I make Cities 2 and 3 have Lighthouse + Harbor and make an Archer. 
You _don't_ see the Sealift button appear on the unit. 

However the pathfinder DOES work between the Cities,
image
(Although it labels the action as Airlift)

Btw fun observation: airlift using the pathfinder (rather than the action button on the interface) causes the unit graphic to SPRINT directly across the face of the map. Same for change port. Looks hilarious! :D

(2) Vox Populi/Database Changes/Text/en_US/UI/NewUIText.xml
```diff
@@ -1129,6 +1129,9 @@
 		<Row Tag="TXT_KEY_CITY_HAS_AIRPORT">
 			<Text>Allows [COLOR_POSITIVE_TEXT]Airlifts[ENDCOLOR] to or from this City. Increased Air Unit Capacity of the City. +10 [ICON_STRENGTH] Damage to Enemy Air Units during Air Strikes on the City.</Text>
 		</Row>
+		<Row Tag="TXT_KEY_CITY_HAS_SEALIFT">
+			<Text>Allows [COLOR_POSITIVE_TEXT]Transportation[ENDCOLOR] to or from this City.</Text>
+		</Row>
```
azum4roll, line 1134: Where is this used?

Hokath:
Ok so I ran 2 AI games with this + sealift=1 on the harbor and everything seems to have worked fine. (Gandhi won Culture victory both times...?!)

Last thing to do is add the city banner icon, i will use the cargo ship and try to copy what kungcheops did for industrial connections

Hokath:
Ok that's all working too.

I'm not sure what the files in (2) are for, since they don't seem to be part of the installation? Anyway I updated them too.

CvGameCoreDLL_Expansion2/Lua/CvLuaCity.cpp
```diff
@@ -2612,6 +2614,36 @@ int CvLuaCity::lIsHasBuilding(lua_State* L)
...
+	{
+		lua_pushboolean(L, true);
+	}
+	else
+	{
+		lua_pushboolean(L, false);
+	}
+	return 1;
+}
+//------------------------------------------------------------------------------
```
azum4roll, line 2617: You can simplify the entire thing to just
```cpp
int CvLuaCity::lCanAirlift(lua_State* L)
{
	return BasicLuaMethod(L, &CvCity::CanAirlift);
}
int CvLuaCity::lCanSealift(lua_State* L)
{
	return BasicLuaMethod(L, &CvCity::CanSealift);
}
```
or even
```cpp
LUAAPIIMPL(City, CanAirlift)
LUAAPIIMPL(City, CanSeaLift)
```

Praise whoward!

Hokath:
ok so here's an interesting case
image

zoom of the minimap
image

It seems I'm connected by water to my Cities on the right side, though I myself do not know the route.
I.e. the Cities of Babylon and Ottomans do know the route around, and I'm connected through them, I guess?

But I imagined that the check for
```
bool bWaterConnection = kPlayer.GetCityConnections()->AreCitiesDirectlyConnected(pCurrentCity, pTargetCity, CvCityConnections::CONNECTION_ANY_WATER);
```
would not evaluate true here?

Unless it somehow goes through the Ice at the top?
image

CvGameCoreDLL_Expansion2/Lua/CvLuaCity.cpp
Hokath: ok i see the header has a special thing too along lines of
LUAAPIEXTN(CountTerrain, int, iTerrain);

Hokath:
Thoughts about airlift rules:
1) you can airlift to allied city-states, should you also be allowed to go to defensive pact partners?
2) should civilians have reduced requirements, e.g. allowed to go to open borders partners
(problem here with diplo units. should not be allowed to just go to any friendly city-state)
3) you can go to an allied city-state, but you can't get back. should travel be reciprocal?

PR #12816, opened 2026-03-30
Parameterize SQL queries and add post-mod ANALYZE

JohnsterID:
Replace sprintf_s string interpolation with parameter binding (?/Bind) for all SQL value parameters in C++ and Lua code.

C++ (6 files, 11 queries):
- CvInfos.cpp: HandicapInfo_Goodies, GameSpeed_Turns (count+select), Resource_QuantityTypes, BuildFeatures
- CvProjectClasses.cpp: Project_VictoryThresholds
- CvPromotionClasses.cpp: UnitPromotions_UnitName
- CvDllDatabaseUtility.cpp: PostDefines ROWID lookup, DatabaseRemapper UPDATE (converted to prepare-once with Bind/Execute/Reset per row)
- CustomMods.cpp: CustomModPostDefines ROWID lookup

GameSpeed_Turns count query changed from SingleResult to Results because SingleResult auto-steps in SetCommand() before Bind() can be called.

Lua (2 files, 3 queries):
- CivilopediaScreen.lua (x2): Leader_Traits JOIN query
- VP - EUI Compatibility Files/ImprovedTopPanel/TopPanel.lua: Leader_Traits JOIN query

Add DB.Analyze() call in CvDllDatabaseUtility::CacheGameDatabaseData() after ClearCountCache() to refresh query planner statistics for the 60+ indexes added by mod SQL files.

### DIFF
--- a/(2) Vox Populi/Core Files/Overrides/CivilopediaScreen.lua
+++ b/(2) Vox Populi/Core Files/Overrides/CivilopediaScreen.lua
@@ -2774,7 +2774,7 @@ CivilopediaCategory[CategoryTech].SelectArticle = function( techID, shouldAddToL
 			thisPlayer = Players[Game.GetActivePlayer()];
 			if thisPlayer ~= nil then
 				leaderID = thisPlayer:GetLeaderType();
-				for leaderTraits in DB.Query( "SELECT TraitType FROM Leader_Traits INNER JOIN Leaders on Leaders.Type = LeaderType WHERE Leaders.ID = " .. leaderID ) do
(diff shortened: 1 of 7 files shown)


azum4roll:
Don't modify base EUI; bc1 won't like that.

vaderkos:
It'd be nice if we could also found tables that lack indexes. But overall until we make `GameInfo` orm itself fast, the difference might be neglegible

JohnsterID:
Fixed up and conflicts resolved.

Issue #12364, opened 2025-12-30
CTD on spreading religion to barbarian-owned city


simpleisbliss:
### 1. Mod version

5.0.1

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

n/a

### 4. Describe the Issue

CTD when spreading religion to barbarian-owned city. Error screen shows `ePlayer < MAX_MAJOR_CIVS` in CvCultureClasses.cpp line 3589, invalid player index.

azum4roll:
I don't think Evangelism is supposed to give tourism towards the target city owner only, based on its description and strength. (It used to work that way though, when the numbers were huge.) I'll be changing it to give tourism towards all major players.

It also currently gives tourism if the target city is not the holy city of the missionary owner's primary religion. I'll be changing it to only give tourism if the target city is not owned by the missionary owner ("if the City is foreign").

PR #12818, opened 2026-03-30
Curious case of Belief_YieldFromSpread

Hokath:
This table is not meant to check if the City is Foreign, we have Belief_YieldFromForeignSpread for that.

I found that the behavior was changed in commit 430e3b0
If I cross reference the commit text, it seems there was some confusion?
Azum talks in #12364 about changing the Tourism, but that part is the Belief_YieldFromForeignSpread
So, looking at the second part of that change, should here
```
CvPlayer &kPlayer = GET_PLAYER(m_eOwner);
if (pCity->getOwner() != m_eOwner)
{
  int iOtherFollowers = pCity->GetCityReligions()->GetFollowersOtherReligions(eReligion, true);
  kPlayer.doInstantYield(INSTANT_YIELD_TYPE_F_SPREAD, false, NO_GREATPERSON, NO_BUILDING, iOtherFollowers, false, pCity->getOwner(), plot());
}
```

instead this pCity->GetOwner() be replaced with NO_PLAYER? It will also need
`if(ePlayer != NO_PLAYER && ePlayer != GetID())` 
guard in the DoInstantYield to be removed (and the ePlayer != GetID() isn't necessary because it is checked in CpUnit.cpp)

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlayer.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlayer.cpp
@@ -26175,10 +26175,7 @@ void CvPlayer::doInstantYield(InstantYieldType iType, bool bCityFaith, GreatPers
 						}
 						if(pCity != NULL)
 						{
-							if (pCity->getOwner() != GetID())
-							{
-								iValue += pReligion->m_Beliefs.GetYieldFromSpread(eYield, GetID(), pLoopCity, true) * max(1, iPassYield+1);
-							}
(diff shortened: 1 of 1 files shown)


azum4roll:
Change `if(GET_PLAYER(ePlayer).isMinorCiv() && eYield == YIELD_TOURISM)` to `if (!GET_PLAYER(ePlayer).isMajorCiv() && eYield == YIELD_TOURISM)` in the INSTANT_YIELD_TYPE_F_SPREAD part. It shouldn't matter since you can never missionary spread to a barb city but it's clearer that way.

PR #12822, opened 2026-03-31
Corporation load order problem

Hokath:
Change to the modinfo was lost by the rebuild, needed to change the project file too.

This means offices aren't giving yields to resources locally in current version.


azum4roll:
Maybe put it before BuildingChanges.sql... Something between that and BuildingChanges2.sql doesn't look good.

vaderkos:
That's why better to do it migration style, so the order is determined by date of the changes

azum4roll:
I can't imagine trying to figure out what the mod does 3 years later if it's done that way.

vaderkos:
> I can't imagine trying to figure out what the mod does 3 years later if it's done that way.

That's a common pattern, but it's easily done, because then you just have descriptive history of changes.

PR #12825, opened 2026-04-01
Lua Things pt. 2 (dll embedded lua script - preparation for pt. 3)

schnetziomi5:
This is a follow up for the changes made to the scripting system in #12680 and preparation for part 3.

Key things:
* Backward compatibility: `Player` and `Team` are once again globally available, fixing some modmod bugs
* A lua script embedded in c++ code is now run after the database has been loaded cached by the .dll. The script runs once, will hook into the script system and automatically remove the hooks again when the .dll is unloaded.

The script has two functions:
* It resets the `GameCore` table before loading into a game. Currently if you load another save while already in game, anything put there will persist (vanilla behavior)
* It spawns a lua context 'VPUI_loader' that is populated with additional API, exposing A) some standard lua functionality or modified, sandbox-compliant variants thereof and B) functionality to change the objects other lua contexts are populated with.

In part 3 (which is entirely .lua files and basically finished, save for some code cleanup) the exposed functionality will be used to replace the standard `include` with a more efficient version that can directly reuse (some selected) previously created objects instead of potentially loading the same file over a hundred times from disk per game (re-)load (e.g. `VPUI_core` or `EUI_utilities`!). This will reduce disk-io, compilation work and memory consumption. (And will not require any changes to the files calling `include`)


vaderkos:
@schnetziomi5 What about using R macro? That'll make everything much more readable, https://en.cppreference.com/w/cpp/language/string_literal.html.

```cpp
const char* luaCommand = R"(
local G = REGISTRY._LOADED._G ;
local Threads = G.Threads ;
local IncludeFileList = G.IncludeFileList ;
)";
...etc.
```

CvGameCoreDLL_Expansion2/Lua/CvLuaSupport.cpp
```diff
@@ -61,48 +64,292 @@ void LuaSupport::RegisterScriptData(lua_State* L)
...
+"local setmetatable = setmetatable ;\n"
+"local type = type ;\n"
+"local _ENV = getfenv(0) ;\n"
+"\n"
+"Threads[coroutine.running()] = {} ; -- Prevent my global env from being cleared out by the engine!\n"
+"\n"
+"local function sandboxedCall( func, ... ) setfenv( 0, getfenv(0), setfenv( 0, getfenv(func) ), pcall( func, ... ) ) end;\n"
+"local newThreadCallback = nil ;\n"
+"local function setNewThreadCallback( func ) newThreadCallback = ( type(func) == 'function' and func ) or error('Not a function!') end;\n"
+"local function setfenv2( func, env, ... ) local ret = getfenv(func) ; setfenv(func, env) ; return ret, ... end;\n"
```
vaderkos, line 123: I'd rather call it swapfenv or replacefenv :D. What's the reason of passing ... here, it doesn't look like it's used 🤔

schnetziomi5:
oh yeah, this isn't too pretty but I couldn't figure out something better ... problem is R literals were introduced with c++11

Also Sorry ... My intention was to upload the script alongside the PR but Github wouldn't let me because .lua is a not a supported file format apparently ...
So with a .txt ending:
[VP_LuaFramework.txt](https://github.com/user-attachments/files/26436061/VP_LuaFramework.txt)

vaderkos:
@schnetziomi5 I think we are actually at c++11 or higher lvl. @JohnsterID, @axatin, @azum4roll Do you know?

vaderkos:
@schnetziomi5 I think if you're fixing include that should make CPK loader obsolete :D

CvGameCoreDLL_Expansion2/Lua/CvLuaSupport.cpp
```diff
@@ -61,48 +64,292 @@ void LuaSupport::RegisterScriptData(lua_State* L)
...
+"		to[k] = v ;\n"
+"	end\n"
+"end;\n"
+"\n"
+"local extraApi = {\n"
+"	setNewThreadCallback = setNewThreadCallback ;\n"
+"	rawset = rawset ;\n"
+"	rawget = rawget ;\n"
+"	newproxy = G.newproxy ;\n"
+"	settenv = function( env,... ) local ret = getfenv(0); setfenv(0, env); return ret,... end ;\n"
```
vaderkos, line 163: What's the difference between set**t**env and set**f**env?

CvGameCoreDLL_Expansion2/Lua/CvLuaSupport.cpp
schnetziomi5:
```
local function setfenv2( func, env, ... ) local ret = getfenv(func) ; setfenv(func, env) ; return ret, ... end;
local function callWithEnv( func, env, ... )
	assert( type(func) == 'function', 'First argument to callWithEnv must be a function!' ) ;
	assert( type(env) == 'table', 'Second argument to callWithEnv must be a table!' ) ;
	return select( 2, assert( select( 2, setfenv2( func, setfenv2( func, env ), pcall(func, ...) ) ) ) ) ;
end;
```
`setfenv2` is just a helper function for `callWithEnv`. If it didn't return any extra arguments `callWithEnv` would need to create(=allocate) a short lived table just to capture all results returned by pcall(func,...). This way the function stays on the stack and it's only a single line of code...

vaderkos:
> oh yeah, this isn't too pretty but I couldn't figure out something better ... problem is R literals were introduced with c++11
> 
> Also Sorry ... My intention was to upload the script alongside the PR but Github wouldn't let me because .lua is a not a supported file format apparently ... So with a .txt ending: [VP_LuaFramework.txt](https://github.com/user-attachments/files/26436061/VP_LuaFramework.txt)

@schnetziomi5 What about something like this? I'm not a cpp expert but that could work...
```cpp
#define LUA_SCRIPT R"lua(
#include "script.lua"
)lua"

const char* lua = LUA_SCRIPT;
```

CvGameCoreDLL_Expansion2/Lua/CvLuaSupport.cpp
schnetziomi5: setfenv is the standard lua function that can set the global environment of any function or the calling thread (by calling setfenv(0,env)). settenv is a sandbox compliant version, that can only strictly set the active thread's environment. Reason for passing extra arguments are the same as above.

Now I am actually not sure if standard setfenv would also be sandbox compliant. Possibly. 
But that's the rationale behind callWithEnv - it changes the environment of a function, calls it (using pcall) and then restores the previous environment and only then passes any exception. Thus you can't permanently change the environment of a function with it...

CvGameCoreDLL_Expansion2/Lua/CvLuaSupport.cpp
vaderkos: Neat :D

schnetziomi5:
```
#define LUA_SCRIPT R"lua(
#include "script.lua"
)lua"

const char* lua = LUA_SCRIPT;
```
yes, if our compilers support 'R' raw string literals the script can be included like this and it would be my preferred solution :)

axatin:
> @schnetziomi5 I think we are actually at c++11 or higher lvl. @JohnsterID, @axatin, @azum4roll Do you know?


We're on C++03

azum4roll:
More accurately MSVC v90. It somehow also supports things like `map<int, vector<int>>` without space.

PR #12827, opened 2026-04-03
Update FollowerChanges.sql [Fix]

jarcast:
Thrift not working with bonus resources added by modmods. Fixed by adding a SQL trigger.

### DIFF
--- a/(2) Vox Populi/Database Changes/Beliefs/FollowerChanges.sql
+++ b/(2) Vox Populi/Database Changes/Beliefs/FollowerChanges.sql
@@ -137,14 +137,6 @@ SELECT
 FROM Resources
 WHERE ResourceClassType = 'RESOURCECLASS_BONUS';
 
-CREATE TRIGGER IF NOT EXISTS BeliefFWNewBonusResources AFTER INSERT ON Resources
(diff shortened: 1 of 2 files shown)


Hokath:
In the long run, what do you think is better:
a) Modmods copy existing tables' bonuses
b) Modmods copy no tabels' bonuses, and they are all added in VP triggers

jarcast:
For things not related with the main entry: triggers. Like here the main entry is beliefs and then they refer to resources. You may want to include all of them.
For changes aimed to be global: triggers.
For the rest (like unique components): modmods pick and copy what's needed for their design.

azum4roll:
Triggers don't belong in this file. Go to Triggers.sql.

Will need to notify all modders so they don't get duplicated entries.

Also need triggers on UPDATE and DELETE (which is why I won't bother - there will be hundreds of triggers).

Hokath:
On UPDATE and DELETE it would be the modders own fault, and problem, right?

azum4roll:
> On UPDATE and DELETE it would be the modders own fault, and problem, right?

As much as it is for INSERT, really.

PR #12830, opened 2026-04-03
Policy city yield per monopoly

Hokath:
Policy equivalent of the Netherlands UB effect.

Like other city yield policies, this goes to the plot of the city itself.


Hokath:
It work

azum4roll:
> Like other city yield policies, this goes to the plot of the city itself.

But we should move away from that...

azum4roll:
Missing policy AI changes

Hokath:
> But we should move away from that...

is there any existing example to follow? I didn't see one
so would need to make some new source of yields with tooltip etc., is it?

axatin:
> > Like other city yield policies, this goes to the plot of the city itself.
> 
> But we should move away from that...

I'll update all of them when I get around to it. We need a new source of yields in cities for that

Issue #12838, opened 2026-04-05
Zealotry belief bug


Timur512:
### 1. Mod version

5.2.1

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

no

### 4. Describe the Issue

Zealotry belief suppose, that player can buy LAND units with faith, but what I see in the current version, that I can build air units and even atomic bombs with faith. Further more, atomic bomb cost 1800 faith and have 5250 production to build, in compare, for example, to bazooka, that cost 2100 faith and 2700 production to build
Obvious bug

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[Harun al-Rashid.zip](https://github.com/user-attachments/files/26489742/Harun.al-Rashid.zip)

Hokath:
line 30557 of CvCity.cpp doesn't check anything except Archaeologist.
I didn't see any commits in the search so I assume this has always been the case...?

So maybe Air units weren't given Faith costs in the past and now they are?

Not clear to me which is the better fix (change the DLL to specify land units or remove the faith cost of planes)

@Timur512 could you buy ships too?

Timur512:
I watch this "bug" for several patches/versions, may be it was later, I don't always get to older eras in my plays

Ships no, belief doesn't allow to buy ships with faith

azum4roll:
I guess it's simpler to just remove faith costs from air units.

Hokath:
Ah yes I see there is a special check for the naval units first
`if (pkUnitInfo->GetDomainType() == DOMAIN_SEA && pkUnitInfo->GetSpecialUnitType() == NO_SPECIALUNIT && !GET_PLAYER(getOwner()).GetPlayerTraits()->IsCanPurchaseNavalUnitsFaith())
				return false;`

People might want to buy air units with faith with other mechanisms, so I might try to do the more fundamental fix of skipping air in the calc. (I do something like that in Devotion mod, but because of how it works with dummies it isn't actually affected by this)

PR #12849, opened 2026-04-09
Zealotry fix

Hokath:
Fundamental fix for #12838

Will allow modmods to have faith purchasable unique naval units with RequiresFaithPurchaseEnabled=0

Part of the logic looks a bit redundant--the archaeologist check--but that (the first check) is for CP-only.
(This way you could do like a belief that lets you buy Cargo Ships and it wouldn't get unlocked by Zealotry)

### DIFF
--- a/(2) Vox Populi/Database Changes/Units/UnitSweeps.sql
+++ b/(2) Vox Populi/Database Changes/Units/UnitSweeps.sql
@@ -105,14 +105,14 @@ SET HurryCostModifier = 25
 WHERE SpaceshipProject IS NOT NULL;
 
 -----------------------------------------------------------------
--- RequiresFaithPurchaseEnabled: only applicable to land/air units
+-- RequiresFaithPurchaseEnabled
 -- This blocks the unit from being always faith purchasable as long as it has a faith cost
 -----------------------------------------------------------------
 
 -- All non-naval military units except mercenary units
 UPDATE Units
 SET RequiresFaithPurchaseEnabled = 1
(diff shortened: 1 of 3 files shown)


(2) Vox Populi/Database Changes/Units/UnitSweeps.sql
```diff
@@ -105,14 +105,14 @@ SET HurryCostModifier = 25
 WHERE SpaceshipProject IS NOT NULL;
 
 -----------------------------------------------------------------
--- RequiresFaithPurchaseEnabled: only applicable to land/air units
+-- RequiresFaithPurchaseEnabled
 -- This blocks the unit from being always faith purchasable as long as it has a faith cost
 -----------------------------------------------------------------
 
 -- All non-naval military units except mercenary units
```
azum4roll: You missed a comment

azum4roll:
But now you just blanket ban air units anyway?

Hokath:
Yes, from the Zealotry table evaluation.

What I learned was that there is a way to make faith purchasable planes currently without changes:
you specify each plane in the specific unit faith purchase table, and give them a faith cost.
The function checks if they are in this table and, if they are, it blocks them from being evaluated in the zealotry part.
(in the absence of this you need to set faith cost to zero on planes -- current behavior)

New behavior you do exactly the same thing, you can just give faith cost to planes even without said belief and not worry about it. 

The main improvement of this PR is for UU naval units, ironically.

I will test the functionality today.

Hokath:
image
correctly doesnt let me buy plane with zealotry even with a faith cost, so that's good :D
Nor a Galley.

But as Spain...
image
turn 1 screen. well that's not good

Hokath:
Ok that all seems to be working, now.

PR #12851, opened 2026-04-09
Rewrite and optimize PlotMouseOverInclude

vaderkos:
Flow is now easier to follow.
Caches only necessary things.
Does not produce a magnitude of intermediate strings anymore.

Still minor optimization till the times when whole plot hover text is reworked.


azum4roll:
Anything that shouldn't be in (1)? We should move it there.

PR #12853, opened 2026-04-10
Border Settle policy column for modmod

Hokath:
Allows you to bypass the border check and culture bomb with Settlers.
Before, bypassing the check would steal the tiles but not cause any diplomatic incident, by using culture bomb code we trigger all the related consequences. 
Rudimentary AI diplo consideration to stop them using the effect on people they wouldn't culture bomb.

Since I was looking at tile acquisition, this also includes a swap around of ProcessBuilding logic to support Legen's proposal


Hokath:
@axatin the last commit here [0605981](https://github.com/LoneGazebo/Community-Patch-DLL/pull/12853/commits/0605981e548bb0a085188ae23fc7cf9b2c4bf260) is needed to make Legen's Tatara work as intended (place iron, then grab it)

PR #12856, opened 2026-04-12
5.2.2 fixes

axatin:


azum4roll:
Seems all these Civilopedia additions should belong in (1). Will there be any issue if we just move CivilopediaScreen.lua and CivilopediaScreen.xml to (1)?

axatin:
> Seems all these Civilopedia additions should belong in (1). Will there be any issue if we just move CivilopediaScreen.lua and CivilopediaScreen.xml to (1)?

It should be fine I think, will test it later and move the files if it works.

Hokath:
Nice additions!

Issue #11728, opened 2025-03-21
Changing world age in Communitu also changes resource settings


Zeodex:
Changing the world age in Communitu automatically switches several other resource settings, and switching back the resource settings also reverses the world age. This happens immediately when changing the setting in multiplayer, but also seems to occur in singleplayer, as updating the game pace also switches the resources and starting a game with a different world age changes the default resource setting for the next game.

https://github.com/user-attachments/assets/2af542da-4642-454d-b369-d34d8e23da72

https://github.com/user-attachments/assets/27c3f527-8d78-4c37-a2c8-7f22af7340d6

Zeodex:
Hi @axatin, this issue still persists for me with the latest map script.

axatin:
It works fine for me. Are you sure you're not using an older copy of the map (Are you using a modpack, and did you manually copy the map to the Maps folder)?

I just realized I didn't update the version number when I fixed the bug. To be sure it's actually the newest version, use the file below, the version number in the map selection screen must be 3.1.1.

[Communitu_79a.zip](https://github.com/user-attachments/files/21209678/Communitu_79a.zip)

Zeodex:
Yes, I am still having the bug with the file above.  The settings don't seem to immediately change anymore when changing the world age, but changing settings like the map size or game pace after changing the world age results in the same bug.

PR #12869, opened 2026-04-16
Fix strncmp prefix-matching bug in mapscript CustomOption

Alpakinator:
Issue:
image
image

GetMapOption/SetMapOption and GetGameOption/SetGameOption used strncmp(storedName, searchKey, strlen(storedName)), which made short option names like "1" falsely match longer keys like "10"-"17".

This caused Map.GetCustomOption(17) to return option 1's value (2), affecting default values of these >=10 options if they were intended to have default value that is not 2 (option 1 has default 2 usually so in this case the error doesn't manifest.)

Replaced strncmp with strcmp for exact string matching. 

Tested and it works.


axatin:
oh, nice. this should fix #11728

PR #12871, opened 2026-04-16
New Diplo Unit Art

Hokath:
Gazebo's CSD came very early on in Civ V modding and used recycled models for the Units (Emissary=Settler, Envoy=Missionary, Ambassador=Archaeologist).

Since the community has many more models now, it was suggested by Jarcast to add distinct models for these units, and update (and freshen up) the icons to match.

Current progress:
image
We are still searching for an Ambassador model.

The icons:
image
the Diplomat and Ambassador are from artists whose permission we have not yet obtained, so we are writing to them.
Hinin helped provide the ancient Great Diplomat icon, and Gwennog has also given advice.


Hokath:
We've received permission from RawSasquatch for the Ambassador icon.
With Jarcast's extraction expertise I've recolored the Archaeologist that is used to have a black suit and team color waistcoat 
image
We've also got a new model for the modern great diplomat with ceremonial robe matching the icon (or, best as I could manage :D)

I think the Envoy above is a good icon match for the Renaissance Great Diplomat. We can use the old modern great diplomat instead for the common unit. Maybe we can do another recoloring but atm should be ok.

Hokath:
we kept a female unit in the diplo, ymmv on the historicity 
image
the white highlights on the dress, belt, and cuffs are team color

with icon by me and jarcast
image

Hokath:
old envoy got a model made, but we've decided against using it. 
This is what we had to match the second icon above:
image

image

Not uploading the model for size reasons, but the icon is kept in the atlas in the empty slot on the row, for now

Hokath:
finally, the icon for the man in the powder wig, is the renaissance/industrial great diplomat
image

PR #12887, opened 2026-04-23
Remove faith purchase check in the city UI

### LINKED ISSUE #12884: [5.2.3] Cannot purchase civilian units in puppet cities with EUI
### 1. Mod version

5.2.3

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

_No response_

### 4. Describe the Issue

In EUI, when you enter the city screen of a puppet _(and you're not Venice)_, there are no buttons to purchase civilian units, a feature implemented in version 5.2.
Image

In non-EUI, the Purchase button will appear on puppet city screens, with an according menu. **Purchasing a civilian unit works fine.** The problem is no UI elements to perform this in EUI.
Image
Image

[Save with puppet cities.zip](https://github.com/user-attachments/files/26946423/Save.with.puppet.cities.zip)
Example 5.2.3 save with puppet cities in it.

_No response_

_No response_

Hokath:
Tested with a puppet in current VP and it makes Worker purchasable correctly

Fixes #12884

### DIFF
--- a/(3a) VP - EUI Compatibility Files/LUA/CityView.lua
+++ b/(3a) VP - EUI Compatibility Files/LUA/CityView.lua
@@ -1500,44 +1500,61 @@ end)
 	-- Update Selection List
 	-------------------------------------------
 
-	local unitSelectList = table()
-	local gpUnitSelectList = table()
-	local religiousUnitSelectList = table()
-	local landUnitSelectList = table()
-	local seaUnitSelectList = table()
-	local airUnitSelectList = table()
-	local militaryUnitSelectList = table()
-	local buildingSelectList = table()
-	local projectSelectList = table()
-	local nationalWonderSelectList = table()
-	local wonderSelectList = table()
-	local processSelectList = table()
+	local isActivePlayerCity = cityOwnerID == Game.GetActivePlayer()
+	local isCityCaptureViewingMode = UI.IsPopupTypeOpen(ButtonPopupTypes.BUTTONPOPUP_CITY_CAPTURED)
+	local isSelectionList = isActivePlayerCity and not isCityCaptureViewingMode
 
-	if g_isAdvisor then
-		Game.SetAdvisorRecommenderCity( city )
-	end
-	-- Buildings & Wonders
-	local orderID = OrderTypes.ORDER_CONSTRUCT
-	local code = orderID / 64
-	for item in GameInfo.Buildings() do
-		local buildingClass = GameInfo.BuildingClasses[item.BuildingClass]
-		local isWonder = buildingClass and (buildingClass.MaxGlobalInstances > 0 or buildingClass.MaxPlayerInstances == 1 or buildingClass.MaxTeamInstances > 0)
-		local isWorldWonder = buildingClass and (buildingClass.MaxGlobalInstances > 0 or buildingClass.MaxTeamInstances > 0)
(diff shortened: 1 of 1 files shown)


azum4roll:
Make sure faith buildings are not purchasable in puppets with the changes. But really it's better to just call city:IsCanPurchase for all that...

Overall the EUI city view has too much bloat with all the GNK/BNW and Civ5BE checks. If we have to keep maintaining it we should at least remove the bloat.

(3a) VP - EUI Compatibility Files/LUA/CityView.lua
```diff
@@ -1511,58 +1500,44 @@ end)
 	-- Update Selection List
 	-------------------------------------------
 
-	local isSelectionList = not g_isViewingMode or isVeniceException or g_isDebugMode or g_isFaithPurchaseBuildingsInPuppetsMod
```
axatin, line 1514: I think removing the `g_isViewingMode` check will make it possible to invest into building in foreign cities using spies.

Hokath: ~~ah interesting :D i will fix~~
Wait, wouldn't the "or" work for Venice in that case?

Hokath: ok so IsVeniceException is `isActivePlayerCity and not isCityCaptureViewingMode and civ5_mode and bnw_mode and cityOwner:MayNotAnnex() and city:IsPuppet()`
so it checks for active player city already

axatin:
> Make sure faith buildings are not purchasable in puppets with the changes. But really it's better to just call city:IsCanPurchase for all that...
> 
> Overall the EUI city view has too much bloat with all the GNK/BNW and Civ5BE checks. If we have to keep maintaining it we should at least remove the bloat.

For all individual items IsCanPurchase is called to check if they're available. The changed code only controls when the selection menu is shown

(3a) VP - EUI Compatibility Files/LUA/CityView.lua
axatin: yeah, but isViewingMode is used for both puppets and foreign cities, and the exclusion of foreign cities needs to stay
```
g_isViewingMode = city:IsPuppet() or not isActivePlayerCity or isCityCaptureViewingMode
```

Hokath: i just set isSelectionList = true to see what happens in a city with a spy
image
the buttons are here but i cannot click them. also, the active/greyed out depends on the Babylonian treasury apparently

Hokath: `isSelectionList = isActivePlayerCity and not isCityCaptureViewingMode`
hides the list for all cities for some reason...

Hokath: was a scope issue.

so im not sure if line 1457 above actually works?

axatin: should be `(not isActivePlayerCity or isCityCaptureViewingMode)` in 1457, shouldn't it?

axatin: and yes, you're right, isActivePlayerCity is not in scope, so the condition is always false. we can use `(not goldCostPQ or not isVeniceException)` directly instead

Issue #12786, opened 2026-03-23
5.1.4 Hot Seat stuck/freeze on AI turn


thetrollingstone:
### 1. Mod version

5.1.4

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

I don't run any additional mods, the modpack I use is 5.1.4 version from here:
https://drive.google.com/drive/folders/1Qq5mWnGGv2uuEPA62azseUhw-pE4--3O

I did update Database Changes from Vox Populi to disable the barbarians quest (my GF hates it :D )
So table Defines still has 1 for 'QUEST_DISABLED_HORDE'

### 4. Describe the Issue

I'm playing HotSeat with my girlfriend, the issue is after we end turn, the game gets stuck on Denmarks turn (no crash, but it's just stuck and not responding)

CPU usage is stuck at 3% more or less, RAM is like 1.5Gb very slowly rising by .1 Mb a minute or so it seemed (didn't have the time to keep the game stuck for longer to double-check)

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[Latest.zip](https://github.com/user-attachments/files/26170714/Latest.zip)

### 6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

[Logs.zip](https://github.com/user-attachments/files/26170697/Logs.zip)

### 7. Crash dump (ALWAYS ATTACH IF REPORTING A CRASH)

There is no dump, since the game doesn't crash.

azum4roll:
Should be the Squads lock, already fixed in 5.2

Issue #12721, opened 2026-03-12
Game Frozen when hovering land tile with semi-embarked squad


AdamFipke:
### 1. Mod version

5.1.4

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

I am using the 5.1.4_VP-EUI_SSEM modpack located here https://forums.civfanatics.com/threads/voxpopuli-modpacks-designed-to-be-used-primarily-in-multiplayer-last-update-5-1-4.685164/

I also have Wonder Race enabled, but I doubt that has any affect.

### 4. Describe the Issue

I have a save where I have a squad of units. Some embarked in the water, some not. If I select this squad and then attempt to move it onto land, the game freezes. More specifically, it freezes when I hover a land tile with my squad selected.

Note that since the game freezes, I don't think I get a crash log message. I need to use task manager to close the game.

Steps to reproduce:

1. Open the save
2. Double tap `1` to select squad 1
3. Move your cursor from a water tile to a land tile. The game will freeze.

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[QuickSave.zip](https://github.com/user-attachments/files/25926272/QuickSave.zip)

### 6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

[Logs.zip](https://github.com/user-attachments/files/25926287/Logs.zip)

### 7. Crash dump (ALWAYS ATTACH IF REPORTING A CRASH)

None available since the game freezes instead of crash.

axatin:
Probably the same as #12679 , so it should be fixed in the next version

PR #12898, opened 2026-04-27
Fix Squads OOB Read Error

### LINKED ISSUE #12840: Squads freeze with 5.2.2
### 1. Mod version

5.2.2

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

none, but I am using the modpack version of VP

### 4. Describe the Issue

Squads freezes the game in 5.2.2 for me when I try to move the squad. Manage squad is fine, but move freezes. Tested on 2 computers with Windows 11, both times trying to move ships. This didn't happen in 4.2.2. With 5.2.2, it freezes and not even Alt-F4 works, nor does it let me alt-tab to task manager. Start menu works so I have to reboot.

_No response_

_No response_

_No response_

artemigkh:
Finally got to the bottom of suspicious hangs occurring in Squads in optimized releaase but not debug builds. Problem was that when the number of units in a squad is too big, or the plot assignment request had no valid solutions, it would try increasingly larger `RING_PLOTS` radiuses to expand the candidate search space. However, this array only has 6 entries, so if it failed to find a solution in the first 5 rings the next read would be to an out of bounds address past the `RING_PLOTS` array. On debug builds this failed silently because the compiler initialized data memory to 0s, so the value at the next out of bound address it read would return a 0, hence the program finished the loop and exited cleanly.

However, on optimized builds where adjacent memory could be holding anything, this loop can go on for a potentially extremely long time. This is almost certainly the same root cause behind
#12721 and #12786 as well, this explains why it didn't just crash but once managed to recover after 10s of hours of wating.

This MR also fixes a related problem where port cities weren't considered water traversable for the purpose of `DOMAIN_SEA` units, resulting in an empty list of destination plots matching the destination type. This fixes Squads movement previews for squads including ships previewing group movement to a coastal city.

Fixes #12840

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvUnit.cpp
+++ b/CvGameCoreDLL_Expansion2/CvUnit.cpp
@@ -15140,13 +15140,20 @@ void CvUnit::DoSquadPlotAssignmentsByDomain(std::vector<CvUnit*> eligibleUnits, 
 	// ring when there are still insufficient plots for the current selection
 	std::vector<CvPlot*> eligiblePlots;
 	int currRingEndIdx = 0;
-	for (int targetPlotIdx = 0; targetPlotIdx < RING_PLOTS[currRingEndIdx]; targetPlotIdx++)
+	for (int targetPlotIdx = 0; 
+		currRingEndIdx < 5 /* Ring array size */ && targetPlotIdx < RING_PLOTS[currRingEndIdx]; 
+		targetPlotIdx++)
 	{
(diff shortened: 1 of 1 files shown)


azum4roll:
Can squads use the "sealift" or "airlift" abilities?

artemigkh:
> Can squads use the "sealift" or "airlift" abilities?

It can do whatever the `MISSION_MOVE_TO` mission can do. If airlifts were taken into account by pathfinding AStar and used by the regular move unit mission, a unit moved through squads should just follow the same mission.

PR #12902, opened 2026-04-29
More Fixes

axatin:


CvGameCoreDLL_Expansion2/CvGameCoreUtils.h
```diff
@@ -496,6 +496,8 @@ class fraction
 	inline bool operator<=(const fraction &rhs) const { return operator<(rhs) || operator==(rhs); };
 	inline bool operator>=(const fraction &rhs) const { return !operator<(rhs); };
 	inline bool operator>(const fraction &rhs) const { return !operator<(rhs) && !operator==(rhs); };
+	inline int getDen() const { return den; };
+	inline int getNum() const { return num; };
```
azum4roll, line 501: I think it's better to have a tostring function for fractions instead of exposing the member variables.

axatin: yes I agree

PR #12904, opened 2026-04-29
Feature request from Hinin

Hokath:
For a new mesoamerican espionage civ.

Spy Point based on Won Wars. Implement using the Accomplishments table.
Also add a new Accomplishments table for yield modifiers as well as flat yields.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlayer.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlayer.cpp
@@ -37848,7 +37848,7 @@ void CvPlayer::CompleteAccomplishment(AccomplishmentTypes eAccomplishment)
 					        const AccomplishmentBonusInfo& bonusInfo = vBonuses[iBonus];
 					        // apply the bonuses
 					        pLoopCity->ChangeBaseHappinessFromBuildings(bonusInfo.iHappiness);
-			
+
 							CreateSpies(bonusInfo.ExtraSpies);
 							
 					        if (bonusInfo.eDomainType != NO_DOMAIN)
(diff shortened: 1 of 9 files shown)


CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -37829,7 +37848,7 @@ void CvPlayer::CompleteAccomplishment(AccomplishmentTypes eAccomplishment)
 					        const AccomplishmentBonusInfo& bonusInfo = vBonuses[iBonus];
 					        // apply the bonuses
 					        pLoopCity->ChangeBaseHappinessFromBuildings(bonusInfo.iHappiness);
-
+
```
azum4roll: Literally unplayable

Hokath: does you IDE make the screen flash when there's an extra tab?

Hokath:
unfortunately my inserting a new column into the middle of this thing is not reading correctly, hitting errors like
image
which suggests the read-in is scrambled somehow. i don't see it atm so going to have to debug

CvGameCoreDLL_Expansion2/CvPlayer.cpp
azum4roll: No, but it's obvious when you read the diff.

For SQL though the file name does turn red indicating error.

Hokath:
The problem has mysteriously vanished on rebase and rebuild.
I think maybe it was triggering the accomplishment with lua, possibly.

PR #12940, opened 2026-05-04
Optimize YieldIconManager

vaderkos:
### Overview

The old YieldIconManager relied on the engine's `ShowHexYield` event to decide
which plots to show icons for. This event proved unreliable -- it fired for every
plot on the map when yields were toggled (not just visible ones), caused redundant
rebuilds on `HexYieldMightHaveChanged`, and gave no control over when or which
plots were updated. There was also no viewport culling: the engine decided what
was shown with no way to tune or predict it.

This solution replaces that approach with deterministic camera matrix projection.
A new `CameraView` context tracks which plots are currently within the camera's
viewport using exact view-space math, fires enter/leave events per plot, and
`YieldIconManager` responds to those events. The result is predictable frame-accurate
icon visibility, a structured pool that reuses UI instances instead of reparenting
to `Controls.Scrap`, and a separation of concerns that lets any other context
reuse `CameraView` events without duplicating the camera math.

---

### New: `CameraView` context

Fires `LuaEvents.PlotEnterView(idx)` and `LuaEvents.PlotLeaveView(idx)` as plots
enter and leave the camera's visible area. Subscribers manage UI without touching
any camera math.

**How it works**

A timer runs 10 times per second. `CameraViewChanged` updates the stored camera
matrix. `VisibilityUpdated` and `StrategicViewStateChanged` flag the next tick for
a recompute. On a flagged tick the timer builds a fresh set of visible plot indices
and diffs it against the previous set. Leave events fire before enter events so
recycled UI instances are available in the pools before new ones are requested.

- **3D view** -- Back-projects the four screen corners (TL, TR, BR, BL) through
  the view matrix onto the ground plane (wz = 0) to get four grid-space corner
  hexes. The frustum projects as a trapezoid - wide at the top (far edge) and
  narrow at the bottom (near edge). A **scanline** sweep iterates each row between the
  min and max corner rows; each quad edge is intersected once per row to find the
  left/right x bounds, then every tile in that range is marked visible. Wrapping
  and out-of-bounds rejection are handled per-tile inside `grid_to_index`.

- **Strategic view** -- The camera is flat (top-down), so the projection math is
  degenerate and back-projecting corners gives no useful frustum shape. Instead the
  screen center is raycast to find the center hex and a fixed `STRATEGIC_*` hex
  rectangle around it is marked visible.

**Public API**

`PlotEnterView(idx: integer)` -- plot entered the viewport  
`PlotLeaveView(idx: integer)` -- plot left the viewport  
`RequestViewPlots(caller: string | number)` -- request full in-view set (queued, next tick)  
`RespondViewPlots(caller: string | number, plots: table<integer, true>)` -- response with full set  
`RequestViewPlot(caller: string | number, idx: integer)` -- check if a single plot is in view (sync)  
`RespondViewPlot(caller: string | number, idx: integer, isInView: boolean)` -- response for single plot check  

**Tuning knobs**

* `ADJUST_L` / `ADJUST_R` / `ADJUST_T` / `ADJUST_B` -- 3D viewport fraction per edge
* `STRATEGIC_L` / `STRATEGIC_R` / `STRATEGIC_T` / `STRATEGIC_B` -- hex radius per edge

---

### Rewritten: `YieldIconManager`

The manager subscribes to `CameraView` events.

- `PlotEnterView` / `PlotLeaveView` handle ongoing camera movement.
- `RequestViewPlots` is called on initial load, on yield display re-enable, and
  after active player change.
- `HexFOWStateChanged` handles hexes that become revealed while already in the
  viewport, using a `RequestViewPlot` / `RespondViewPlot` round-trip to
  `CameraView` to confirm the plot is currently in view before rendering.
- UI instance pool for anchors, stacks, icons and labels. Pool grows as new unique
  hexes are visited and stabilises once all reachable hexes have been seen. The
  total active + pooled count does not grow after stabilisation.
- `PrintYieldIconStats()` function is available for testing/debug purposes in `YieldIconManager` context.

---

### New: `CPK.Util.Timer`

Lightweight repeating task scheduler. Attaches to a LuaContext's `SetUpdate`
heartbeat, `SetPostInit`, `SetShutdown` and supports multiple named tasks with
independent intervals, lifespans and debounce options. Used by `CameraView`.

---

### New: `CPK.UI.Control.Instance.Move`

Helper to reparent a pooled control instance to a new parent without destroying
and recreating it. Used by YieldIconManager's pool recycling.

---

### Changed: `CPK.Util.Stack`

- `Pop` now reads and nils by index instead of calling `table.remove`, avoiding
  an O(n) array shift on every pool dequeue.
- `Empty` fixed from `self.items[1] == nil` to `self.count <= 0`, which was
  incorrect after any sequence of push/pop operations.

---

### Known Behaviours

**Icon popping near the viewport edge is expected.**  
The `ADJUST_*` constants control how far beyond the logical screen edge icons are
preloaded. At the default values some pop-in at the very edge of the screen is
intentional to keep the active UI control count low. Increase `ADJUST_*` above
`1.0` to widen the preload margin.

**Minimap click does not immediately reposition icons.**  
Clicking the minimap teleports the camera but `VisibilityUpdated` does not fire
until the camera starts moving. Icons will update correctly as soon as any camera
movement occurs (pan, zoom or drag), which triggers a recompute on the next tick.


azum4roll:
Is it the same case for resource icons, unit flags, and city banners?

vaderkos:
> Is it the same case for resource icons, unit flags, and city banners?

@azum4roll Theoretically same logic is applicable to all of them, the simplest that could be done this way is resource icon manager. Unit flags and city banners are complicated, I'm not sure what happens here.

schnetziomi5:
My notes:
- forget about grid coordinates - stick to the hex and world ones. Those are much nicer in a mathematical sense.(They can be converted back and forth using a simple matrix transformation.) I abhor the grid coordinates, they are pain. They kind of translate like `world_x = grid_x + 1/2 * [sin( pi/2 * grid_y )]^2`. We should purge them from lua and abstract them away in the dll.
- The trapezoid thing is too much hardcoding. I think it will break on larger screens? You could use the matrix passed by CameraViewChanged to calculate the plots (= their hex coordinates) in the corners of the screen. They should sit at screen coordinates `( +/-0.5 | +/- 0.5 )`. Doing this every time the camera view changes might be too expensive computationally, so you might want to cache the resulting trapezoid for a given zoom level(or range). The zoom level is in fact just the Z component of the translation vector (which I don't think is the "correct" way to do it, but firaxis' devs did what they did)
- `Events.ShowHexYield` is missing. This is definitely a problem because it's used to display yield icons on selected plots when e.g. a settler, worker or city is selected.
- Some (edge) cases to consider: 
~ Very small maps for which the same plot might be visible twice
~ Strategic View?
~ Maps that wrap vertically

schnetziomi5:
There is also ` UIManager:GetGameGridTopLeft` and `UI:ManagerGetGameGridBottomRight` have you tried using those?

vaderkos:
@schnetziomi5 
1. What's the difference between hex and grid coordinates then... Although, I'm not sure how to use the switch to different units...
2. Trapeziod is in grid units, so it's always the same not matter what screen size is. Or I do not know the case when it would break. Other problem is that there are maps with vertical wrap 0_o, that requires different handling, for cylindrical map this one works... Is there a way to know which kind of a map is loaded @azum4roll, @axatin? Also, @schnetziomi5 somehow there are multiple values encoded in some numbers of the matrix, and separate them was sure a lot of pain, but I still do not get how it does that, I had multiple tries making it work...

Strategic View has it's own rendering as far as I know, it's decoupled from YieldIconManager.
_Very small maps for which the same plot might be visible twice_ -> it checks the bounds, but I see what you mean, but that means that there are maps that are the size that camera renders on full zoom out...

schnetziomi5:
image

axatin:
> 2\. . Other problem is that there are maps with vertical wrap 0_o, that requires different handling, for cylindrical map this one works... Is there a way to know which kind of a map is loaded @azum4roll, @axatin?

`Map:IsWrapX()`and `Map:IsWrapY()`

schnetziomi5:
image

```
changefunc = function( cam )
	if lock then return end ;
	Events.ClearHexHighlightStyle("RangedAttackPreview"); 
	for i = 0 , Map.GetNumPlots()-1 do
	
		local plot = Map.GetPlotByIndex(i) ;
		
		local gx,gy = GridToWorld( plot:GetX() , plot:GetY() ) ;
		local gz = 0 ;
		
		local x2 = cam[1][1] * gx + cam[2][1] * gy + cam[3][1] * gz + cam[4][1] ;
		local y2 = cam[1][2] * gx + cam[2][2] * gy + cam[3][2] * gz + cam[4][2] ;
		local z2 = cam[1][3] * gx + cam[2][3] * gy + cam[3][3] * gz + cam[4][3] ;
		
		local screen_x = x2 / z2 ;
		local screen_y = y2 / z2 ;
		
		
		if -0.45 < screen_x and screen_x < 0.45 and -0.25 < screen_y and screen_y < 0.25 then
			local hex = ToHexFromGrid{ x=plot:GetX() , y=plot:GetY() } ;
			Events.SerialEventHexHighlight(hex, true, nil, "RangedAttackPreview");
		end
	end
end

Events.CameraViewChanged.Add( changefunc ) ;
```

This showcases how the transformation works. The division by `z2` makes inversion tricky. But it's possible to do it analytically for any given matrix. Just somewhat tedious...

vaderkos:
@azum4roll, @axatin, @schnetziomi5, @Alpakinator -> Ready for the next review round. Spent a lot of time handling all the edge cases. I'd really appreciate if someone except me tests that.

vaderkos:
> Is it the same case for resource icons, unit flags, and city banners?

@azum4roll Now you can do it easily with `PlotEnterView` and `PlotLeaveView`

(1) Community Patch/Core Files/New UI/CameraView.lua
```diff
@@ -0,0 +1,437 @@
...
+		end
+	end
+end
+
+--- Fills vis with plot indices visible in 3D view using exact camera projection.
+--- Back-projects screen corners to get a candidate bounding box, then tests each
+--- hex with is_visible. Handles cylindrical and toroidal map wrapping.
+--- @param cam number[][] Civ V view matrix
+--- @param vis table<integer, boolean> Set to write visible plot indices into
+local update_3d_view = function(cam, vis)
```
schnetziomi5, line 295: This seems somewhat inefficient still - I would suggest to abstract grid wrapping away and just assume a rectangle rather than try to fit the trapezoid. The same function might then work for both 3d and SV:
```

function validIndexFromCoords( x, y )
	-- apply wrapping if applicable and return the plot index. 
	-- return nil if a non wrapping coordinate is out of bounds
end

function updateView(cam, vis)
	local min_gx, max_gx, min_gy, max_gy = get_bounds(cam) ;
	for x = min_gx , max_gx do
		for y = min_gy , max_gy do
			local idx = validIndexFromCoords(x,y) ;
			if idx then
				vis[idx] = true ;
			end
		end
	end
end
```

If you want to keep the trapezoid it should be possible to interpolate its boundaries using the full set of screen-cornering plots calculated in `get_bounds`. ~If you want to be totally generic(support any perspective) you might form two triangles from the four corner plots, and for each plot in the outer rectangle check whether it's positioned in one of the rectangles - should be slightly(?) faster than the current is_visible check and especially doesn't need any extra treatment for wrapping maps or around their edges.~ I guess their is an even better way: For each Y(=row of plots) calculate directly or find iteratively(starting from the sides), x_min and x_max corresponding to the X coordinates of the outermost visible plots for that row. All plots in between are visible too.

(1) Community Patch/Core Files/New UI/CameraView.lua
```diff
@@ -0,0 +1,437 @@
...
+
+--------------------------------------------------------------------------------
+
+-- 3D view configuration
+
+-- Finely tuned numbers :D
+local ADJUST_L = 0.90 -- left edge
+local ADJUST_T = 0.50 -- top edge  (far from camera)
+local ADJUST_R = 0.90 -- right edge
+local ADJUST_B = 0.55 -- bottom edge (near camera)
```
schnetziomi5, line 108: Hmmm. These numbers are different to mine. Is there a hidden factor 2 somewhere? Either way having these hardcoded is not good - which means we should try to find their formula...

schnetziomi5 (commented):
Sorry for the delay... also I couldn't yet and can't test this ingame right, so now my review is just based on reading...

(1) Community Patch/Core Files/New UI/CameraView.lua
vaderkos: @schnetziomi5 I tested these values to exclude as many anchors outside the viewport as possible.
Using these values resulted in around 613 possible anchors at maximum zoom-out.

(1) Community Patch/Core Files/New UI/CameraView.lua
vaderkos: @schnetziomi5 So what you actually mean, is that calculating if plot is visible costs more in performance than we save on memory..

schnetziomi5: We are tight on memory and have CPU cycles to spare... but my point is that it's an unnecessary waste of CPU time to call `is_visible` for a plot corresponding to world coordinates `(b,y)` if it's already known that both `(a,y)` and `(c,y)` are visible on screen, and `a<b<c` - in that case `(b,y)` will always be visible...

vaderkos: I'll try to adjust the solution, scanlines is a nice idea :D

vaderkos:
@schnetziomi5 ready for the next code review round. Updated with a scanline implementation - visual popping is almost gone, anchor count looks reduced, and FPS during camera movement seems improved as well.

schnetziomi5:
Did a quick reading, looks good!

Hopefully I have time today or tomorrow to test it live...

It might maybe be reasonable to just hide anchors while the camera is moving and/or only update them when it stopped?

vaderkos:
> Did a quick reading, looks good!
> 
> Hopefully I have time today or tomorrow to test it live...
> 
> It might maybe be reasonable to just hide anchors while the camera is moving and/or only update them when it stopped?

Might be too invasive for the ux.

Issue #12917, opened 2026-05-01
Claude - 5.2.4 Performance Review


Stalker012345:
### 1. Mod version

5.2.4

### 2. Installed Components

Vox Populi with EUI

### 4. Describe the Issue

Here are some optimization recommendations for both dll and lua

[5.2.4_performance reviews.zip](https://github.com/user-attachments/files/27288092/5.2.4_performance.reviews.zip)

Cyclopia31:
Could be nice to also have it on memory usage.

JohnsterID:
https://gist.github.com/JohnsterID/12dfaa8378944bf94dc5d708096c3eb1

Most of the claims aren't real positives.

Stalker012345:
[5.2.4_performance reviews_markdown.zip](https://github.com/user-attachments/files/27305135/5.2.4_performance.reviews_markdown.zip)

here are the files converted to markdown as requested

PR #12943, opened 2026-05-05
Some bug fixes for claude audit

### LINKED ISSUE #12913: Claude - 5.2.4 Bug Review and Audit
### 1. Mod version

5.2.4

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

_No response_

### 4. Describe the Issue

[Claude Bug Reports.zip](https://github.com/user-attachments/files/27280840/Claude.Bug.Reports.zip)

This contains a list of bug reviews done on each of the main dll/cpp files within VP. It also includes a project wide audit that noted some consistent issues across many files.

_No response_

_No response_

_No response_

JohnsterID:
Closes https://github.com/LoneGazebo/Community-Patch-DLL/issues/12913
Only fixes C++ issues in https://github.com/LoneGazebo/Community-Patch-DLL/issues/12917, I haven't looked at the Lua.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvGame.cpp
+++ b/CvGameCoreDLL_Expansion2/CvGame.cpp
@@ -4600,9 +4600,9 @@ void CvGame::UpdateGameEra()
 			iCount++;
 		}
 	}
-	if (iCount >= 0)
+	if (iCount > 0)
 	{
-		int iRoundedEra = int(fEra / (max(1, iCount)) + 0.5f);
+		int iRoundedEra = int(fEra / iCount + 0.5f);
 		m_eGameEra = (EraTypes)iRoundedEra;
 	}
 	else
--- a/CvGameCoreDLL_Expansion2/CvVotingClasses.cpp
+++ b/CvGameCoreDLL_Expansion2/CvVotingClasses.cpp
@@ -207,6 +207,7 @@ bool LeagueHelpers::IsResolutionForTriggerActive(ResolutionTypes eType)
 	if(eType != NO_RESOLUTION)
 	{
 		CvLeague* pLeague = GC.getGame().GetGameLeagues()->GetActiveLeague();
+		if (!pLeague) return NO_RESOLUTION;
 		const ActiveResolutionList& vActiveResolutions = pLeague->GetActiveResolutions();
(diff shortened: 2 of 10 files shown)


CvGameCoreDLL_Expansion2/CvGame.cpp
azum4roll, line 4610: Is it really possible that there's no alive major civ?

CvGameCoreDLL_Expansion2/CvVotingClasses.cpp
```diff
@@ -207,6 +207,7 @@ ResolutionTypes LeagueHelpers::IsResolutionForTriggerActive(ResolutionTypes eTyp
 	if(eType != NO_RESOLUTION)
 	{
 		CvLeague* pLeague = GC.getGame().GetGameLeagues()->GetActiveLeague();
+		if (!pLeague) return NO_RESOLUTION;
```
azum4roll: This function is weirdly coded; it can only return NO_RESOLUTION or eType. Should be converted into a boolean method under CvLeague instead, especially given the name starts with `Is`.

The return value isn't being used either other than the comparison to NO_RESOLUTION.

CvGameCoreDLL_Expansion2/CvGame.cpp
JohnsterID: Probably not in normal gameplay, but the code already had an else branch for exactly that case, it was just unreachable due to `>= 0`. The fix doesn't change behaviour, it just un-deadcodes the safety branch.

CvGameCoreDLL_Expansion2/CvVotingClasses.cpp
JohnsterID: Yes. Refactored to `bool CvLeague::IsResolutionForTriggerActive(ResolutionTypes eType) const`. Both callers (CvVotingClasses.cpp and CvDiplomacyAI.cpp) already had a league pointer and only compared the result to NO_RESOLUTION.

axatin:
Please fix the merge conflicts. I also refactored GetResolutionForTriggerActive, not noticing you already did it here

JohnsterID:
Conflicts resolved.

PR #12945, opened 2026-05-05
(10-041a) Hero Worship Rework Alternative

Hokath:
https://forums.civfanatics.com/threads/10-041a-hero-worship-rework-alternative.702993/

Adds GG and GAd increase directly the change XP method (multiplicative) using a new table. For other GP it works just as a local bonus

Adds the yields on GP Birth no era scaling table, with separate city limit structure instead of old religion-level city limiter implementation. Also has flexibility for a city-level yield boost if used in future for a follower/reform/etc., which differs from existing GP Use and GP Expend tables.


azum4roll:
Not going through IsBeliefValid (or at least a GetOwnedReligion() check) makes it work on non-founders. You'll likely want to use the former method so these tables work on other types of beliefs correctly (founder only for founder beliefs and enhancer beliefs; majority religion AND state religion for reformation beliefs; and majority religion for pantheons and follower beliefs). GGP and GAdP should either pass in the capital or the origin city of the combat unit.

Hokath:
What isn't going through IsBeliefValid?

GetGreatPersonRateModifier calls IsBeliefValid
therefore, for a Founder, it will only work if you own the Holy City, no?

azum4roll:
Hmm I must've misread. Though it still doesn't work on non-state pantheons and followers. Get the majority religion of the origin city of pFromUnit instead of the state religion?

Btw, this looks scary.
image

Hokath:
I don't think we can use the religion of the unit's origin city, rather than the player's state religion.

If we did that, and the origin city didn't follow the state religion, your founder bonus wouldn't work on that unit.
I don't think that's the behavior we want?

azum4roll:
> If we did that, and the origin city didn't follow the state religion, your founder bonus wouldn't work on that unit.

This is the limitation of the religion code I guess. It simply must have different behaviours across belief types. It's the same with God of War currently - your holy city/capital must follow the state religion that has the belief to benefit from it.

Do you (or your binary-speaking friend) have any idea how to modify the code to achieve this?

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -32381,6 +32408,28 @@ void CvPlayer::setNavalCombatExperienceTimes100(int iExperienceTimes100, CvUnit*
...
+		const CvReligion* pReligion = GC.getGame().GetGameReligions()->GetReligion(eStateReligion, GetID());
+		if (pReligion)
+		{
+			CvCity* pOriginCity = NULL;
+			if (pFromUnit)
+			{
+				pOriginCity = pFromUnit->getOriginCity();
+			}
+			const GreatPersonTypes eAdmiral = static_cast<GreatPersonTypes>(GC.getInfoTypeForString("GREATPERSON_ADMIRAL"));
+			int iModifier = pReligion->m_Beliefs.GetGreatPersonRateModifier(eAdmiral, GET_ID(), pOriginCity);
```
azum4roll: What's GET_ID()?

Hokath:
> Do you (or your binary-speaking friend) have any idea how to modify the code to achieve this?

Best idea is that when that point is reached, you check ahead to see if it's a follower belief, if so you give it the origin city religion instead

Hokath:
Everything seems to be working in my test

PR #12949, opened 2026-05-06
Refactor Improvement Era Scaling to avoid Landmark overwriting its own age

### LINKED ISSUE #12946: Issue with culture yields from Landmarks
### 1. Mod version

5.2.5

### 2. Installed Components

Community Patch only

### 3. Additional mods used

IGE+47

### 4. Describe the Issue

In BNW, Landmarks are supposed to produce +1 culture for each era of the artifact prior to the current era. Instead, they produce culture based on each era prior to the construction of the Landmark itself.
Here you can see that the artifact is from the Ancient Era:
Image
However, after constructing the Landmark, it provides 0 culture (the current era is the Modern Era).
Image
It only started providing +1 culture after I advanced to the next era.
Image

[landmark.zip](https://github.com/user-attachments/files/27438006/landmark.zip)
To reproduce the issue, just build a Landmark in the provided save file.

_No response_

_No response_

Hokath:
Fix https://github.com/LoneGazebo/Community-Patch-DLL/issues/12946

Bug was obvious -- Landmark qualifies as an Era scaling improvement, so overwrites its own age! Needs to check for Landmark explicitly.

Notice the existing code to check for Landmark is a bit indirect. Works at the moment but the new Torii would count as a Landmark @azum4roll

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlot.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlot.cpp
@@ -8382,17 +8382,6 @@ void CvPlot::setImprovementType(ImprovementTypes eNewValue, PlayerTypes eBuilder
 					ClearArchaeologicalRecord();
 				}
 			}
-
-			// if the improvement has yield changes by era, it needs the historical record set. 
-			// in the rare case there is also an invisible artifact record on the tile, its age is altered (player has no way to know this)
-			for (int iK = 0; iK < NUM_YIELD_TYPES; ++iK)
-			{
-				if (newImprovementEntry.GetYieldChangePerEra(iK) != 0)
-				{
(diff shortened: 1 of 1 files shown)


Hokath:
image
Ancient Era Landmark working correctly now

RecursiveVision:
Several places in the code check directly for the Landmark, such as Barb Encampment or dig site placement code.

Hokath:
Ok nice, I will look.

To be clear, I didn't write that check, it was already there.

Hokath:
Ok so looking into it more, this GetNumLandmarksBuilt is used just for the Happiness part, not e.g. the voting code.
Indeed even in this block we can see it's incremented by the happiness value, not by +1.
So it's maybe a bit confusingly named?

I'm going to leave that alone here and just have a separate block that checks for "not landmark"

PR #12955, opened 2026-05-07
Change religion territory bonus handling in CvUnit.cpp

### LINKED ISSUE #12953: Crusader Spirit and Defender of the Faith's combat boosts dont't work for non founders
### 1. Mod version

5.2.5

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

_No response_

### 4. Describe the Issue

I'm just gonna report it here before I forget:

https://forums.civfanatics.com/threads/10-102-new-reformation-belief-work-ethic.702781/#post-16941552

This report is because of this comment. I did not check if it is true. I'm sorry if this is not the way to do these things. Thanks.

_No response_

_No response_

_No response_

Hokath:
Fix #12953

1) Check state religion not owned religion, since its a reformation bonus other players should get it 
2) remove holy city argument -- this just results in checking if the holy city follows its own religion, but I don't see why this bonus should stop working in that case (Founders can't even check for that)


axatin:
Will fix the merge conflicts in my branch

PR #12960, opened 2026-05-07
Tactical AI fixes and tweaks

KungCheops:
Fix wrong median unit XP calculation, leading to tactical AI treating all its units as very important.

Make AI more willing to send melee ships in to attack.

Fix AI not wanting to swap units in and out.

Fix GGs and GAs suiciding because they don't know where friendly units are.

Fix units standing still in bad positions.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvTacticalAI.cpp
+++ b/CvGameCoreDLL_Expansion2/CvTacticalAI.cpp
@@ -914,23 +914,30 @@ void CvTacticalAI::ExecuteCaptureCityMoves()
 			{
 				int iRequiredDamage = pCity->GetMaxHitPoints() - pCity->getDamage();
 				int iExpectedDamagePerTurn = ComputeTotalExpectedDamage(*pTarget);
-				//actual siege will typically be longer because not all units actually attack the city each turn
-				int iMaxSiegeTurns = 13; //do we even need a limit here or is this handled through the tactical posture?
 
-				int iCityHealRate = !pCity->IsBlockadedWaterAndLand() ? GD_INT_GET(CITY_HIT_POINTS_HEALED_PER_TURN) : 0;
-
-				//assume the city heals each turn ...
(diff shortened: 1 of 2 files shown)


RecursiveVision:
if MOD_BALANCE_VP is enabled, I'd add getPopulation() to the city's estimated heal rate. Should be a cheap call but an improvement in accuracy.

KungCheops:
> if MOD_BALANCE_VP is enabled, I'd add getPopulation() to the city's estimated heal rate. Should be a cheap call but an improvement in accuracy.

This code isn't in a hot path, I think it's called like maybe a handful of times per turn, so it doesn't have to be cheap. I'm a bit afraid that making it more accurate risks making the AI more passive.

KungCheops:
Although it might also make it smarter about not sending useless attacks towards cities that can never succeed.

RecursiveVision:
> I'm a bit afraid that making it more accurate risks making the AI more passive.

Long-term we should probably be addressing the disease, not the symptoms. :P

KungCheops:
> > I'm a bit afraid that making it more accurate risks making the AI more passive.
> 
> Long-term we should probably be addressing the disease, not the symptoms. :P

You may notice that I already changed it :)

CvGameCoreDLL_Expansion2/CvTacticalAI.cpp
```diff
@@ -7115,12 +7127,12 @@ size_t CvBasePosition::getFirstInterestingAssignment() const
 
 void CvBasePosition::UpdateScore(const STacticalAssignment& assignment)
 {
-	UpdateScore(assignment.iUnitID, assignment.GetPlotScore(), assignment.GetBonusScore(), assignment.GetDamageDelta());
+	UpdateScore(assignment.iUnitID, assignment.GetPlotScore(), assignment.GetOldPlotScore(), assignment.GetBonusScore(), assignment.GetDamageDelta());
```
axatin: DamageDelta and BonusScore are in the wrong order

KungCheops: Good catch

PR #12970, opened 2026-05-09
Congress 10 work ethic

Hokath:
Need to test


Hokath:
Farm goes from 7 turns to 6 turns, success

PR #12972, opened 2026-05-10
Cleanup/assert correctness and formatting

JohnsterID:
Better to place if/else/for/while bodies on separate lines from the condition to allow precise debugger breakpoint placement.


azum4roll:
Note that we may turn off PRECONDITIONs and ASSERTs for stable versions. In that case, those do nothing.

I'm against turning them off though.

JohnsterID:
I would be against blanket disabling. ASSERT (always active in debug/release when CVASSERT_ENABLE is defined so we could switch off in release builds) but PRECONDITIONs should never be turned off.

JohnsterID:
Conflicts resolved.

FirePlace/include/FireWorks/FVariableSystem.h
```diff
@@ -136,17 +136,28 @@ class FVariable
...
+			if ( !_stricmp(szType, "UInt"))
+				return FVARTYPE_UINT;
+			if ( !_stricmp(szType, "Float"))
+				return FVARTYPE_FLOAT;
+			if ( !_stricmp(szType, "Float"))
+				return FVARTYPE_DOUBLE;
+			if ( !_stricmp(szType, "String"))
+				return FVARTYPE_STRING;
+			if ( !_stricmp(szType, "String"))
+				return FVARTYPE_WSTRING;
```
vaderkos, line 160: What's the difference between those?

PR #12985, opened 2026-05-13
(10-130a) Make Open Borders Available At Philosophy

legenciv:
Link to proposal:
https://forums.civfanatics.com/threads/10-130a-make-open-borders-available-at-philosophy.702992/

Should Philosophy receive an AI diplomacy flavor for its Open Borders? I don't know if 'FLAVOR_DIPLOMACY' is meant to include interactions with other major civs.


Hokath:
Azum said that Tech flavors are only used for free techs, so I wouldn't worry about that too much

PR #12988, opened 2026-05-13
vanilla missile text except can be intercepted

Alpakinator:
the original is

```
		<Row Tag="TXT_KEY_AIRPOWER_MISSILES_HEADING2_BODY">
			<Text>Missiles are one-shot weapons. They perform a single air strike mission against a target, and then, win or lose, they are destroyed. Unlike normal aircraft, missiles cannot be intercepted. </Text>
		</Row>
```


azum4roll:
Syntax error

azum4roll:
Also I think this is the case in CP as well?

Alpakinator:
My bad, added missing ; and moved to (1).

PR #12994, opened 2026-05-14
(10-037) (10-041) (10-083) (10-084) (10-129) (10-136) (10-140) (10-A) VP Congress Session #10 Proposals

RecursiveVision:
Please do not squash.

- Implement https://forums.civfanatics.com/threads/10-037-x-turns-open-borders-after-peace.702615
  * Added to Community Patch Only as well.
  * I did the bare minimum as far as notifying the player of the game mechanic (it's now in the peace notification).
  * Lua function Team:GetRemainingDisengagementTurns(eTeam) has been added if anyone wants to make it more visible, since this isn't my area of expertise.

- Implement https://forums.civfanatics.com/threads/10-041-reduce-warmonger-penalty-for-city-capture-if-ai-lost-a-war-to-the-victim.702622

- Implement https://forums.civfanatics.com/threads/10-083-culture-bombing-with-a-citadel-is-treated-as-warmongering.702970
  * Also add warmongering values for stealing via Manifest Destiny or Homestead
  * Warmongering is not applied when revenge stealing (see just below)

- Implement https://forums.civfanatics.com/threads/10-084-reduce-warmonger-amount-for-retaliating-against-culture-bombs.702813
  * Casus Belli is applied to the city when it steals tiles from another major civ, and is cleared when the city is acquired by the team whose tiles were stolen
  * Casus Belli is not applied when revenge stealing (tit-for-tat per city)
  * Casus Belli is not applied if the stolen tile is within 3 tiles of the thief's capital (those are 'their' tiles)
  * Casus Belli is ignored by the civ that owns the city and their teammates/masters/vassals
  * Casus Belli is displayed in the city's warmonger tooltip if the reduction is active for any civ who is present in the tooltip

- Implement https://forums.civfanatics.com/threads/10-129-reduce-minimum-turns-needed-to-make-peace-keep-war-weariness-timing-unchanged.702909

- Implement https://forums.civfanatics.com/threads/10-136-restore-food-to-ancient-classical-era-maritime-city-state-bonuses.702916

- Implement https://forums.civfanatics.com/threads/10-140-tea-nerf.702929

- Implement https://forums.civfanatics.com/threads/10-a-remove-re-vassalization-cooldown-after-rebellion.702907

- Update war damage when making peace, since cities can be traded as part of the peace deal, and that would increase war score for the winner
  * This is done before war scores are stored for (10-041) purposes
  * The post-city exchange value is used for (10-041) purposes, but the highest of the pre- and post- city exchange values is used for determining whether a war has been won

- Move and rescale GetWarScore()
  * Should really be in CvPlayer since the underlying mechanics work for City-States too
  * Now stores the value in 2x units, allowing odd war scores to exist

- Religion now only affects warmongering penalties if the civ is a founder and it's their religion that's being spread/hindered (also tweaked the values a bit)

- War victory bonuses now work properly for teammates
  * All teammates of a winning player get the war victory bonuses, even if they were losing themselves
  * Winning a war against a team of multiple players still only counts as one war

- Fix an advanced start bug where the AI would not register recent peace treaties in many parts of the code

- Fix CvGameDeals::ActivateDeal() to *actually* always process Peace Treaties last, like the comment says is supposed to be the case

- Fixes for warmongering code
  * All warmonger triggers now increment the 'previously done this action' counter *after* applying the warmonger penalty, not before
  * Adjusted initial weights accordingly
  * The warmongering penalty for using nukes is now 40 for all players (was 40 for victim, 20 for other players), plus 20 for each previous time witnessed
  * Using nukes now applies increasing warmongering penalties for every player, not just when the same player is nuked repeatedly
  * Using nukes on City-States now also counts for warmongering, and reduces your Influence with the City-State to -60
  * Vassals of the nuking player and dead players also no longer apply warmongering penalties
  * Exposed warmonger trigger defines to the database

- Citadels, America UA and Homestead can no longer steal tiles from teammates


azum4roll:
eOwner can be NO_PLAYER here. Same in CvUnit.cpp and CvBuilderTaskingAI.cpp.

RecursiveVision:
Fixed, thanks.

RecursiveVision:
Fixed the merge conflicts.

PR #12997, opened 2026-05-15
Add escort missions for building improvements

KungCheops:
AI now has the ability to create escort missions for Great Generals, Archaeologists and regular Workers. Previously they could only do it for Settlers, Great Merchants and diplomat units. Future consideration for Great Prophets or perhaps all missionary units.

AI will now escort Great Generals when they are building citadels. As well as when they are moving between cities.

AI can now escort Archaeologists and Workers when they are building outside of friendly territory.

AI can now escort workers moving between different non-contiguous parts of their empire.

Consolidate archaeology logic into the regular worker logic.

Worker AI:

Add support for IsIgnoreOwnership and IsOnlyCityStateTerritory builds.

Add evaluation of archaeology digs in builder code.


Hokath:
Escort on Great General building Citadel is very nice

PR #12975, opened 2026-05-10
Congress 10 Subterfuge

Hokath:

PR #13000, opened 2026-05-16
Modernize culture from specialists

Hokath:
Azum's suggestion (https://github.com/LoneGazebo/Community-Patch-DLL/pull/12975#issuecomment-4446722421) to move the column CulturePerTurn on Specialists into the SpecialistYields table, as was done in some other places like Buildings

Actually this wasn't the issue in the end, but this does seem to work and I guess it's nicer/cleaner to have?


azum4roll:
Function declaration still exists in CvInfos.h

axatin:
then we can also remove the member variable m_iCulturePerTurn

PR #13001, opened 2026-05-16
Add contender info to mouse over

Hokath:
Request from Gwennog
Otherwise you have to open the panel to see this information, is tiresome when there are many CS


axatin:
Please use text keys so it can be localized and not hard-coded English lines of text in the code file

axatin:
I'll fix

PR #13005, opened 2026-05-16
Remove rivers and lakes from civilopedia features list

### LINKED ISSUE #12832: Civilopedia: Rivers and Lakes are in Features list, but not features
### 1. Mod version

5.2.2

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

_No response_

### 4. Describe the Issue

In the civiliopedia, Rivers and Lakes are noted under the "features" list,. However, they are not features in game terms and should be removed.

_No response_

_No response_

_No response_

jtpavlock:
Fixes #12832 

For reference, the only "FakeFeatures" are rivers and lakes.

Before and after pics:
image
image

### DIFF
--- a/(2) Vox Populi/Core Files/Overrides/CivilopediaScreen.lua
+++ b/(2) Vox Populi/Core Files/Overrides/CivilopediaScreen.lua
@@ -1434,28 +1434,6 @@ CivilopediaCategory[CategoryTerrain].PopulateList = function()
 		categorizedList[(CategoryTerrain * absurdlyLargeNumTopicsInCategory) + row.ID + 1000] = article; -- add a fudge factor
 	end
 
-	-- now for the fake features (river and lake)
-	for row in GameInfo.FakeFeatures() do
-		local article = {};
-		local name = Locale.ConvertTextKey( row.Description )
-		article.entryName = name;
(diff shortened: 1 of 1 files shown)


azum4roll:
There still needs to be something that shows TXT_KEY_CIV5_FEATURES_RIVER_TEXT and TXT_KEY_CIV5_FEATURES_LAKE_TEXT.

Hokath:
Here, I think a Lake tile really fits the definition of a Terrain (even if the code is a fake feature)

On this topic, Hills are also given a fake allocation as Terrain
Maybe Hill and River should go in a new section called like "Types" or "Modifiers" or something like this (fancy: morphology/geomorphology)

azum4roll:
Hill is a plot type just like mountain/flatland/water. We can add a separate category for them.

Lake is just coast that's reclassified by the game.

River is its own thing.

Hokath:
If we did show Flat/Hill/Mountain/Water plot types to the player, we could take the opportunity to explain plot height
which i think is also explained nowhere, currently

axatin:
We shouldn't remove them unless they are added somewhere else.

jtpavlock:
If a solution is determined in #12832 I can try to submit another PR.

PR #13010, opened 2026-05-17
Add CPK.UI.Control.Instance.Manager with Queue and Inspect utilities

vaderkos:
## CPK Instance Manager

Adds a managed lifecycle replacement for the base game `InstanceManager`, alongside two supporting utilities.

### Motivation

The base game `InstanceManager` has no lifecycle hooks, no state tracking, no good error reporting, and locks the parent control at construction time. Every UI screen that uses it ends up with the same boilerplate: manual show/hide, no visibility into what is active or pooled, and silent failures when instances are misused. This PR introduces a CPK replacement that addresses all of that while keeping the migration path straightforward.

### New files

**`CPK.UI.Control.Instance.Manager`** — the main addition. Manages UI control instances through three states: *acquired* (in use), *recycled* (returned to the free list, hidden), and *released* (permanently destroyed).
- `Acquire(parent)` / `Recycle(inst)` / `Release(inst)` — core lifecycle
- `onCreated` fires once on first allocation — good for registering callbacks and static setup
- `onAcquire` / `onRecycle` / `onRelease` hooks for per-use configuration and cleanup
- `mode = 'LIFO'` (default) or `'FIFO'` ordering of the recycled free list — FIFO gives stable sequential ordering without needing `SortChildren`
- `RecycleAcquired()` / `ReleaseRecycled()` for bulk operations
- `HasInstance` / `HasAcquired` / `HasRecycled` / `Size` / `CountAcquired` / `CountRecycled` for inspection
- Error messages with pool state, instance details, current game state and context

**`CPK.Util.Queue`** — two-stack amortised O(1) FIFO queue with the same `Push` / `Pop` / `Peek` / `Size` / `Empty` interface as `Stack`, usable as a drop-in alternative inside the manager for ordered recycling.

### DIFF
--- a/(1) Community Patch/Community Patch.civ5proj
+++ b/(1) Community Patch/Community Patch.civ5proj
@@ -3941,6 +3941,10 @@
       <SubType>Lua</SubType>
       <ImportIntoVFS>True</ImportIntoVFS>
     </Content>
+    <Content Include="Kit\UI\Control\CPK.UI.Control.Inspect.lua">
+      <SubType>Lua</SubType>
+      <ImportIntoVFS>True</ImportIntoVFS>
+    </Content>
     <Content Include="Kit\UI\Control\CPK.UI.Control.IsKind.lua">
       <SubType>Lua</SubType>
       <ImportIntoVFS>True</ImportIntoVFS>
@@ -3973,6 +3977,14 @@
       <SubType>Lua</SubType>
       <ImportIntoVFS>True</ImportIntoVFS>
     </Content>
+    <Content Include="Kit\UI\Control\Instance\CPK.UI.Control.Instance.Manager.lua">
+      <SubType>Lua</SubType>
+      <ImportIntoVFS>True</ImportIntoVFS>
+    </Content>
+    <Content Include="Kit\UI\Control\Instance\CPK.UI.Control.Instance.Move.lua">
+      <SubType>Lua</SubType>
+      <ImportIntoVFS>True</ImportIntoVFS>
+    </Content>
     <Content Include="Kit\Util\CPK.Util.Benchmark.lua">
       <SubType>Lua</SubType>
       <ImportIntoVFS>True</ImportIntoVFS>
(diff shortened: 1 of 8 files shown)


schnetziomi5 (approved):
Got nothing to say, looks good :)

schnetziomi5:
> Register new Kit files in Community Patch.civ5proj

How would you feel if we created a pseudo DLC that contains all the CPK files? (Like `VPUI` - which isn't actually deployed in the `CP only` installation, despite `VPUI_core` already being in the `CP` mod. We could also use such a DLC to override mod-agnostic front end UI and add tile highlights in `CP`[if we ever found a  workaround for the crashes])

vaderkos:
> > Register new Kit files in Community Patch.civ5proj
> 
> How would you feel if we created a pseudo DLC that contains all the CPK files? (Like `VPUI` - which isn't actually deployed in the `CP only` installation, despite `VPUI_core` already being in the `CP` mod. We could also use such a DLC to override mod-agnostic front end UI and add tile highlights in `CP`[if we ever found a  workaround for the crashes])

Then it should be the first mod to load

schnetziomi5:
Pseudo dlcs are loaded at startup - way before any mod could be enabled or any lua is executed..

vaderkos:
> Pseudo dlcs are loaded at startup - way before any mod could be enabled or any lua is executed..

@schnetziomi5 never heard about it, seems like a nice solution :). But do we really need it when you introduce your `include` fixes?

azum4roll:
The DLC approach is only for pre-mod menu UI changes and files that can't be modded.

schnetziomi5:
The point is you can make the engine load all CPK files into the virtual file system by specifying a folder rather than having to tediously register every file. Since CPK isn't overriding anything there is zero potential for conflict and it would ease development...?

schnetziomi5:
> But do we really need it when you introduce your include fixes?

It's long overdue that I finish that code. However, as a security measure, it will still rely on the old include internally as a gatekeeper when loading anything from disk, meaning files have to be known to the virtual file system...

azum4roll:
I don't know how to add new files to the VFS via DLC either. How does MPMPM do it?

vaderkos:
@schnetziomi5 ideally if we could just load all these utilities once globally. Idk if it's possible via yours include rewrite, but very desired :D

PR #13012, opened 2026-05-17
Prevent minor civs from choosing limited units as their UU

### LINKED ISSUE #12967: MaxPlayerInstances units can exceed limit if gifted by City-States
### 1. Mod version

5.2.5

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

.

### 4. Describe the Issue

This is not necessarily a problem, but I'll record it here.
If a unit class has MaxPlayerInstances set, e.g. the Turkish Great Bombard, then the player cannot exceed that number of units. Great.
However, these units can appear as Militaristic City-State Gifts. I have had a user report to me in Cultural Diversity that this gifting ignores the cap (in their case the cap is 1 and they have 3 of the unit).

I checked the MinorCivAI code and it seems we have two flags already to deal with this in 
NoMinorCivUU
NoMinorCivGift

The first prevents the unit being chosen as their unique by the minor civ, and the second prevents you gifting the unit to the minor civ AND the unit being chosen. The second is set on the Conquistador (although it is only working in the "gift-to" capacity because for unique selection there is also a check to exclude Units that can found Cities)

We could set one of these flags on the Turkish UU. Or we could amend the DLL so the minor civ cannot continue gifting units over the cap.
Or do nothing :)

_No response_

_No response_

_No response_

jtpavlock:
This avoids a player from being gifted units that would cause them to exceed the limit.

Fixes #12967

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvGame.cpp
+++ b/CvGameCoreDLL_Expansion2/CvGame.cpp
@@ -8656,6 +8656,15 @@ UnitTypes CvGame::GetRandomUniqueUnitType(bool bIncludeCivsInGame, bool bInclude
 				continue;
 			}
 			
+			// No units that have a cap on instances (player/team/global)
+			// gifting these could push a player over their limit
+			if (pkUnitClassInfo->getMaxPlayerInstances() != -1)
+				continue;
+			if (pkUnitClassInfo->getMaxTeamInstances() != -1)
+				continue;
+			if (pkUnitClassInfo->getMaxGlobalInstances() != -1)
+				continue;
+
 			// We only want unique units that are not in the game already, or are explicitly Minor Civ Gifts
 			if (!pkUnitInfo->IsMinorCivGift() )
 			{


CvGameCoreDLL_Expansion2/CvGame.cpp
```diff
@@ -8656,6 +8656,16 @@ UnitTypes CvGame::GetRandomUniqueUnitType(bool bIncludeCivsInGame, bool bInclude
 				continue;
 			}
 			
+			// No units that have a cap on instances (player/team/global) -
+			// gifting these could push a player over their limit
+			// use -1 as "no limit" sentinel
+			if (pkUnitClassInfo->getMaxPlayerInstances() != -1)
```
RecursiveVision, line 8661: sentinel? I'd just keep the first two lines of the comment. No need for the - in the first line either.

jtpavlock: edited as requested

azum4roll:
There's the `isLimitedUnitClass()` function that checks all 3 plus city limit (which should probably also be blocked from being unique).

axatin:
I'll use isLimitedUnitClass instead

Issue #13021, opened 2026-05-19
5.2.6 - Immediate error on turn end / CS Alliance w/ Rome


santheocles:
### 1. Mod version

5.2.6

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

none

### 4. Describe the Issue

Immediately after ending the turn, the following error message pops up. No issues if accepting it, no crash

<img width="412" height="492" alt="Image" src="https://github.com/user-attachments/assets/bd048ee9-86bc-45e2-9e31-5d6af8289345" />

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[QuickSave.zip](https://github.com/user-attachments/files/28009517/QuickSave.zip)

Hokath:
also #12791

Hokath:
i guess the city of york has just been captured?
(and we are missing the rome UA being updated for it)

PR #13026, opened 2026-05-21
Rome UA Fixes

### LINKED ISSUE #12791: assertion error
### 1. Mod version

5.2.1

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

R.E.D.  + I.g.e.

### 4. Describe the Issue

assertion error apeared once after capturing Copenhagen from Denmark and hitting end turn

Image

[auto.zip](https://github.com/user-attachments/files/26221374/auto.zip)

 too big

too big

Hokath:
1. Update yields on newly acquired cities, fixes #12791 #13021

2. Include friend yields as well as ally, so yields indeed behave as if the annexed CS "were allies"

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvCity.cpp
+++ b/CvGameCoreDLL_Expansion2/CvCity.cpp
@@ -28877,10 +28877,10 @@ void CvCity::UpdateYieldsFromExistingFriendsAndAllies(bool bRemove)
 	//this method should only be called once per city on its first turn or when the capital moves ...
 
 	int iSign = bRemove ? -1 : +1;
-
+	CvPlayer& kPlayer = GET_PLAYER(getOwner());
+	
 	if (isCapital())
 	{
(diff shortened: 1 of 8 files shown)


CvGameCoreDLL_Expansion2/CvCity.cpp
```diff
@@ -28895,6 +28895,21 @@ void CvCity::UpdateYieldsFromExistingFriendsAndAllies(bool bRemove)
 		}
 	}
 
+	// Rome UA
+	if (kPlayer.GetPlayerTraits()->IsAnnexedCityStatesGiveYields())
```
RecursiveVision, line 28899: Would avoid "Rome UA" comments as the Rome UA is changing.

Hokath: yeah idd
can go through and ctrl+R them

PR #13039, opened 2026-05-23
(10-004c) colonialism rebalance (no new generalized bonus)

Hokath:
The "sensible" way to implement this means it will also affect the Indonesia trait. Therefore this is an out-of-congress balance change and needs approval.
Personally, I think that's probably fine gameplay-wise: the unique restriction is easy to miss and the logic about changing the Imperialism policy also applies to the Indonesia trait.
Note however that Indonesia went up a lot in latest AI test, from 18th place to 5th.

For the Golden Age Yield % modifiers, it seems like these already stacked?
They are calculated on-the-fly as a loop over resources, and the monopoly mod bonus is added per-resource
```
// Resource monopolies (not cached, so give an option to skip this check)
    if (MOD_BALANCE_RESOURCE_MONOPOLIES && bCheckMonopolies)
    {
        for (int iResourceLoop = 0; iResourceLoop < GC.getNumResourceInfos(); iResourceLoop++)
        {
            ResourceTypes eResourceLoop = (ResourceTypes)iResourceLoop;
            if (eResourceLoop != NO_RESOURCE)
            {
                CvResourceInfo* pInfo = GC.getResourceInfo(eResourceLoop);
                if (pInfo && pInfo->isMonopoly())
                {
                    if (HasGlobalMonopoly(eResourceLoop) && pInfo->getMonopolyGALength() > 0)
                    {
                        int iTemp = pInfo->getMonopolyGALength();
                        iTemp += GetMonopolyModPercent();
                        iModifier += iTemp;
                    }
                }
            }
        }
    }
```
 
so either this was bugged and no one noticed, or I missed something?

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlayer.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlayer.cpp
@@ -30733,6 +30733,25 @@ int CvPlayer::GetMonopolyModFlat() const
 void CvPlayer::ChangeMonopolyModPercent(int iChange)
 {
 	m_iMonopolyModPercent += iChange;
+
+	// all the yield monopolies you already have need to be updated
+	for (int iResourceLoop = 0; iResourceLoop < GC.getNumResourceInfos(); iResourceLoop++)
+	{
+		ResourceTypes eResource = (ResourceTypes) iResourceLoop;
+		CvResourceInfo* pkResource = GC.getResourceInfo(eResource);
+		if (pkResource && m_pabHasGlobalMonopoly[eResource])
+		{
+			for (int iI = 0; iI < NUM_YIELD_TYPES; iI++)
+			{
+				eYield = (YieldTypes)iI;
+
+				if (pResource->getCityYieldModFromMonopoly(eYield) != 0)
+				{
+					changeCityYieldModFromMonopoly(eYield, iChange);
+				}
+			}
+		}
+	}
 }
 void CvPlayer::SetMonopolyModPercent(int iChange)
 {
@@ -38042,6 +38061,7 @@ void CvPlayer::SetHasGlobalMonopoly(ResourceTypes eResource, bool bNewValue)
 				int iModValue = pResource->getCityYieldModFromMonopoly(eYield);
 				if (iModValue != 0)
 				{
+					iModValue += GetMonopolyModPercent();
 					if (bNewValue)
 					{
 						changeCityYieldModFromMonopoly(eYield, iModValue);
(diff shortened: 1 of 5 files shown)


Hokath:
Updates existing monopolies correctly
image

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -30733,6 +30733,25 @@ int CvPlayer::GetMonopolyModFlat() const
 void CvPlayer::ChangeMonopolyModPercent(int iChange)
 {
 	m_iMonopolyModPercent += iChange;
+
+	// all the yield monopolies you already have need to be updated
+	for (int iResourceLoop = 0; iResourceLoop < GC.getNumResourceInfos(); iResourceLoop++)
```
azum4roll: There's actually a GetGlobalMonopolies() method you can use to get the vector of all monopoly resources. A lot faster.

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -30733,6 +30733,25 @@ int CvPlayer::GetMonopolyModFlat() const
...
+	{
+		ResourceTypes eResource = (ResourceTypes) iResourceLoop;
+		CvResourceInfo* pkResource = GC.getResourceInfo(eResource);
+		if (pkResource && m_pabHasGlobalMonopoly[eResource])
+		{
+			for (int iI = 0; iI < NUM_YIELD_TYPES; iI++)
```
azum4roll, line 30745: GC.getNUM_YIELD_TYPES() to skip unused yields.

Hokath: i copied this from nearby
should we replace it everywhere?

azum4roll: I replace whenever I come across it. Or you can find a time when there are few pull requests to batch update everything. There isn't a chance in the past few months.

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -30735,9 +30735,10 @@ void CvPlayer::ChangeMonopolyModPercent(int iChange)
...
 
 	// all the yield monopolies you already have need to be updated
-	for (int iResourceLoop = 0; iResourceLoop < GC.getNumResourceInfos(); iResourceLoop++)
+	const std::vector<ResourceTypes>& vGlobalMonopolies = GetGlobalMonopolies();
+	for (int iResourceLoop = 0; iResourceLoop < vGlobalMonopolies.size(); iResourceLoop++)
 	{
-		ResourceTypes eResource = (ResourceTypes) iResourceLoop;
+		ResourceTypes eResource = vGlobalMonopolies[iResourceLoop];
 		CvResourceInfo* pkResource = GC.getResourceInfo(eResource);
 		if (pkResource && m_pabHasGlobalMonopoly[eResource])
```
azum4roll, line 30743: Redundant check here. It's already known to be global monopoly!

azum4roll: And getResourceInfo() can never return NULL in practice.

PR #13040, opened 2026-05-23
Tooltip fix: add bullet icon to CS per Wonder and Damage Reduction

Hokath:
These guys use the non-bullet code so they need it in the text key
e.g.
image


Hokath:
renaming branch deletes it apparently

PR #13041, opened 2026-05-23
Congress 10 Art

Hokath:
New PI are in atlas 7, starting row 3 like so:
image
This includes new icons for barbarian promos. The silver chevron will be used to distinguish "ignore terrain cost" for effects that increase movement above base (gold chevron).
There is also a jungle-only icon for future use.
The dune graphic is used instead of the cactus graphic for desert-movement effects.

Wonders are in new building atlas 3
image
The Grand Canal has a new icon
The Menin Gate icon has been remastered
(I'm not in love with the Port Royal icon, can be improved in future)

building atlas 2 now has 15 free slots for new buildings
image
The Dar-e Mehr was also included in the final slot and can be replaced if needed. 
The Order icon has been updated. 
The Chancery and Holy Council have been remastered. 
The Tidal Power Plant has been denoised of compression artifacts.
The Chartarium has a new icon.


azum4roll:
What about the art define for the Lake Salt feature? Would

```xml
		<Row> 
			<StrategicViewType>ART_DEF_FEATURE_LAKE_SALT</StrategicViewType> 
			<TileType>Feature</TileType> 
			<Asset>SV_Salt.dds</Asset> 
		</Row>
```

work?

Hokath:
Gwennog is still working on that (because the base mod it is a bonus resource, it has a different type of icon)

PR #13043, opened 2026-05-23
(10-038) Japan: Dojo Replaced By Tatara (Unique Forge)

legenciv:
Link to proposal:
https://forums.civfanatics.com/threads/10-038-japan-dojo-replaced-by-tatara-unique-forge.702619/

TODO:
- Tatara text (historical info, strategy)
- Icon replacement
- LostOnGifting (when/if "10-143 Reduce UA Promotion Transferability and Clutter" is implemented)

Before I replace the Dojo icon for the Tatara one, what is the policy regarding old icons? Should I leave the current Dojo icon on a particular atlas/file/slot, or just remove it from its current atlas?

Also, what is the intended yield order? The auto generated UI places Production before Science, while the current tooltip for Council of Elders and Divine Inheritance (NW) place Production after Science (and Gold).


azum4roll:
Yield order is however the Yields table is defined. Currently the only intended exception is pantheon description. Other beliefs are all over the place as they haven't been rewritten.

legenciv:
3rd Tatara icon added (with rays behind, see [post](https://forums.civfanatics.com/threads/10-038-10-039-10-040-japan-reworks.703005/page-2#post-16945806)) on VPBuildingAtlas1, where the Dojo icon used to be. Dojo icon preserved on slot 45, in case a reversal is needed.

LostOnGifting left for a later patch, when proposal (10-143) gets implemented. Otherwise, all Database changes finished. 

There's a DLL change desired to Building_ResourceClaim, discussed during the proposal phase (make it happen after Building_ResourcePlotsToPlace, instead of before). @Hokath, can you implement it?

Hokath:
I already did ;)

Also, did you compress the non-xcf atlases with DXT5? The size looks a bit large?

legenciv:
Thanks! And I didn't compress; first time working with images, I wasn't aware of this step. I prepared a compressed version of these files now, and tested in game to confirm they're ok.

azum4roll:
AI isn't going to build it without flavors.

legenciv:
Good reminder about flavors. 

One question: how are the flavors added and used to/from the database? I see the flavors of all UBs added to Building_Flavors in Civ5DebugDatabase.db, but not on Civ5CoreDatabase.db.

And not just for the Tatara, but for UBs of many other civs; the only UB flavors appearing on the latter file are UBs that were already present on vanilla (_e.g._ Arabian Bazaar, Indian Mughal Fort), and with the wrong flavors (_e.g._ Mughal Fort has culture 25, city defense 25, instead of city defense 40, culture 15 wonder 10 set on BuildingFlavorSweeps.sql). It also has UBs from vanilla that aren't present anymore in VP (_e.g._ Chinese Paper Maker)

UBs added during VP/4UC, such as the Ottoman Siege Foundry and the Polynesian Marae are absent from  Civ5DebugDatabase.db.

I'm asking because either I'm missing something, or it may be a bug affecting how the AI handles UBs that don't have an equivalent in vanilla/BNW.

PR #13045, opened 2026-05-24
Modernize CultureFromExperienceDisbandUpgrade

Hokath:
Convert to use doInstantYield, completing TODO
Will also change from boolean to integer, allowing modders to set higher conversion rates

facility is also possible to discriminate between scrap/upgrade and have different values, if desired, using bEvent flag

new TODO: allow any yield


azum4roll:
Isn't it too much culture even if set to 1? Will need a times100 value if you want it to be balanced on early game units.

Probably better to make it two tables for disband and upgrade separately.

Hokath:
If you wanted to give it to a very early unit yes, you would need x100

also yes, in the future disband and upgrade split along with all yields are good objectives

PR #13059, opened 2026-05-30
[Balance] Remove Treasure Ship increased prod cost 

Hokath:
At the moment, this unit unlocks earlier and has higher cost. 
The reason is that the "Revolt China" proposal gave +25% Prod to Units (while Unhappy), and it was argued this would be OP for early unlock Frigate, necessitating the higher cost. 
(In the original 4UC the treasure ship only had early unlock, standard cost.)

However, the implementation of this UA was abandoned.  

Therefore I think it would be consistent if this cost increase also be dropped.


Hokath:
By the way, all the cost changes to UU are in the CostSweeps file.
It feels like it is more natural to have their changes over the base unit in the respective Civilizations/ file? Should we move them?

To do that the sweep would have to go into the changes, earlier in the load order.

azum4roll:
Sweeps are post-uniques by definition.

Also can you remove the other congress 8 changes from China?

(2) Vox Populi/Database Changes/Text/en_US/Civilizations/NewCivilizationText.xml
```diff
@@ -433,7 +433,7 @@
...
 		</Row>
 		<Row Tag="TXT_KEY_UNIT_TREASURE_SHIP_STRATEGY">
-			<Text></Text>
+			<Text>The Chinese {TXT_KEY_UNIT_TREASURE_SHIP} unlocks earlier than the Frigate it replaces. This gives a larger window to use its Kowtow promotion, which is [COLOR_NEGATIVE_TEXT]Lost on Upgrade[ENDCOLOR]; this promotion converts combat experience into [ICON_INFLUENCE] Influence with a nearby [ICON_CITY_STATE] City-State. Therefore a large fleet of active {TXT_KEY_UNIT_TREASURE_SHIP}s can be used to gain control over what might be City-States on the other side of the world. They are aided in this by their Quick Study promotion, which not only accelerates them towards important combat promotions, but also increases the Influence they gain! Floating a large number of ranged vessels far from your borders sounds dangerous, but together with their early unlock the {TXT_KEY_UNIT_TREASURE_SHIP} is stronger than the Frigate, and also has access to the otherwise unavailable Hull III promotion, which makes them extremely sturdy. All you need is somewhere to heal - perhaps the City-State with whom you have gained all that Influence?!</Text>
```
azum4roll: We don't use text keys on strategy texts anymore

Hokath:
> Sweeps are post-uniques by definition.

That's what I wrote. It would have to be "unswept".

> Also can you remove the other congress 8 changes from China?

What do you mean? That would revert them to pre-4UC

azum4roll:
> What do you mean? That would revert them to pre-4UC

The GGP and GAdP during golden age on Siheyuan.

azum4roll:
> That's what I wrote. It would have to be "unswept".

Then it'll have to be UnitCostChanges.sql? While effectively being a sweep since it's done on every single unit based on tech column. Which also creates a desync between the base unit and the unique unit before the civilization files (supposed to be delta only).

Hokath:
Yes it would be like a UnitCostChanges.sql without any unique units, then the delta in the civilization files

> The GGP and GAdP during golden age on Siheyuan.

Oh, right. It's not actually in the pedia at the moment.
Since that's a Golden Age effect and you are meant to stay unhappy a lot of the time, affecting GA frequency, I think you can definitely argue that should be removed also. Yes.

axatin:
more merge conflicts

Hokath:
rebased

azum4roll:
> Yes it would be like a UnitCostChanges.sql without any unique units, then the delta in the civilization files

That won't work unless we do a complex query updating costs based on the column of the prereq tech of the default unit of the unit class of the unit. A Treasure Ship needs to have identical stats as a Frigate before China.sql.

PR #13061, opened 2026-05-30
Project AI valuation changes

Hokath:
1) Flat unhappy reduction had quadratic modifier scaling. Leads to (very) large values when unhappy.
image

Scaling itself is inappropriate since the value does not change no matter how far below the threshold you are, only that each source of unhappiness is present (for Public Works -- note I didn't consider the hypothetical case when it lowers unhappiness by more than 1 and you have less than that, which would slightly reduce valuation)

If valuation were now to be consistently too low, increase the iScaleFactor variable

2) Happiness boolean divided total score even if the project had non-happiness related effects. Refactor to enter into each effect individually

3) The new CitySupplyFlat and Event-starting columns were not scored. The latter needs something heuristic or AI won't know about special effects in modmods.


axatin:
nice!

PR #13062, opened 2026-05-30
Subterfuge AI Changes

Hokath:
Scoring for both missions and counterspy focusses, including taking into account offensive (negative) and personal (positive) effects separately, where the former is affected by DiploAI

For gaining spies:
Convince AI to work civil servants if they have high espionage flavor, just like IsDiplo civs do

Make Great Diplomat's trade mission preferred for espionage focussed players, based on the balance of espionage and diplo flavors of the leader, and whether the player is trying for diplo victory. 

Evaluate buildings and policies that give spies based on espionage flavor so they are desired more by England, etc.

--------------------------
This also includes some fixes:
- Incorrect kill% for infiltrate mission
- Some text errors/omissions
- Spy points not longer generated by embassies


CvGameCoreDLL_Expansion2/CvPolicyAI.cpp
```diff
@@ -3565,14 +3565,13 @@ Firaxis::Array< int, NUM_YIELD_TYPES > CvPolicyAI::WeightPolicyAttributes(CvPlay
...
-		{
-			yield[YIELD_GOLD] += PolicyInfo->GetFreeSpy() * 250;
-		}
-		else
-		{
-			yield[YIELD_GOLD] += PolicyInfo->GetFreeSpy() * 50;
-		}
+		int iEspionageFlavor = GetFlavorManager()->GetPersonalityIndividualFlavor((FlavorTypes)GC.getInfoTypeForString("FLAVOR_ESPIONAGE"));
+		int iDiplomacyFlavor = GetFlavorManager()->GetPersonalityIndividualFlavor((FlavorTypes)GC.getInfoTypeForString("FLAVOR_DIPLOMACY"));
+		yield[YIELD_GOLD] += PolicyInfo->GetFreeSpy() * (iEspionageFlavor * iEspionageFlavor * iDiplomacyFlavor / 2) * GD_INT_GET(ESPIONAGE_SPY_POINT_UNIT) / GC.getGame().GetSpyThreshold();
```
azum4roll, line 3570: Unless espionage flavor actually should count quadratically, I don't agree with squaring it to get close to 250. Flavors also probably should never be multiplied with each other. Use additions and constant multipliers instead.

Hokath: the idea is that flavor 10 needs to be more than twice as important as flavor 5
cant do that without either either a) non-linear function or b) subtracting some value and sending negative values to 0

Hokath:
ok, I think that covers all cases, and a couple unused ones for potential future expansion (building destruction)

now have to decide how to test it...

RecursiveVision:
There will be a merge conflict with my implementation of (10-013).

I'll hold off on implementing (10-068) until this is implemented.

CvGameCoreDLL_Expansion2/CvPolicyAI.cpp
azum4roll: I think all personality flavors are expected to be linear.

I also note that flavors are completely absent from that part of the code. Should we even start using them?

Hokath:
There's a divide by zero somewhere.

edit: I believe we have detained the suspect

Hokath:
Now need to examine logs.
Right now I imagine things like Steal GW or Steal Tech have too-low evaluation

azum4roll:
AI stockpiles too much NP without performing missions. At the end of every observer game I open the winner's espionage menu and can always instantly do a bunch of missions getting a huge load of yields. Even in enemy cities.

Hokath:
Before that last commit I watched a full England game and they were mainly spamming cheap missions and moving around. Many dead spies.

Now going to do it again with the logging open

Hokath:
Could someone with eagle eyes (@azum4roll) check that last commit "Revive the getTotalGreatPeopleRateModifier function" to see if I missed anywhere

azum4roll:
Misclicked... Seems there's nothing wrong at a glance?

CvGameCoreDLL_Expansion2/CvEspionageClasses.cpp
```diff
@@ -8399,11 +8399,11 @@ int CvEspionageAI::GetNegativeMissionScoreForOwner(CvCity* pCity, PlayerTypes eO
 	if (pkMissionInfo->getReligiousUnrestModifier() < 0 && pCity->GetUnhappinessFromReligiousUnrest() > 0)
 		iScore -= -1 * pkMissionInfo->getEventDuration() * iEmpireUnhappy * pkMissionInfo->getReligiousUnrestModifier() / 75;
 	
-	int iEra = m_pPlayer->GetCurrentEra();
```
azum4roll: Casting is not necessary from C enum to int.

Hokath:
Ok after a big effort today I think things are behaving pretty well.
Still more improvements to make and testing (especially late game), which I am continuing, but maybe worth merging what's completed so the next version can have some testing + the new directive behavior which is quite a big change (not everyone spams embassy)

CvGameCoreDLL_Expansion2/CvGlobals.cpp
```diff
@@ -1578,7 +1578,7 @@ CvGlobals::CvGlobals() :
 	GD_INT_INIT(BASE_GOLDEN_AGE_UNITS, 1),
 	GD_INT_INIT(GOLDEN_AGE_UNITS_MULTIPLIER, 1),
 	GD_INT_INIT(GOLDEN_AGE_LENGTH, 10),
-	GD_INT_INIT(GOLDEN_AGE_GREAT_PEOPLE_MODIFIER, 100),
+	GD_INT_INIT(GOLDEN_AGE_GREAT_PEOPLE_MODIFIER, 0),
```
azum4roll, line 1581: You have to set it in the database too (CoreDefineChanges.sql, put under Golden age section).

Hokath:
[TestGameEspionage.zip](https://github.com/user-attachments/files/29057359/TestGameEspionage.zip)
some logs showing use of all modes of spy for (almost?) all missions
Small map (amusingly Stockholm becomes the capital of Babylon at some point!)

Hokath:
Need to fix that fact that creating embassy is currently also giving spy points

azum4roll:
Can't you add the spy points to the mission function instead of the expended code?

Hokath:
yes, and tbh it doesn't even have to be a great person. ill have a look at refactoring that commit

RecursiveVision:
@Hokath Is this tested?

Hokath:
Yes, have been testing on-and-off since starting a few weeks ago.
You can see in some logs up there ^ that the AI valuation seems to work pretty well (but ofc it's hard to test it exhaustively)

RecursiveVision:
@Hokath Don't use GetJonRandNum() ever, use the various randRange/limit functions.

Hokath:
@RecursiveVision  ok good to know
I would have copied it from somewhere else (probably same file), so maybe that needs fixing, or?

azum4roll:
{331B7626-68A6-4673-B119-7E43B11C5BA6} ???

PR #13077, opened 2026-06-01
Crash reporting improvements

schnetziomi5:
- Generate mindumps in a separate folder `\crashlogs` instead of in the game's root directory
-  Collect additional information about the process like memory markers and whether the system is native windows or a wine derivative(e.g. Proton)
- Show this information in the crash popup and write(append) it to the plaintext file `\crashlogs\crashes.log`
- Soften the wording of the popup message when an exception occurs outside of the gamecore dll: There are a myriad of reasons the engine may crash other than being out of memory...


schnetziomi5:
image
image

axatin:
Thanks! Please also update the minidump path in the bug report template

PR #13091, opened 2026-06-03
Fix #13082

### LINKED ISSUE #13082: Cannot see yield icons in city screen
### 1. Mod version

5.3.0

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

_No response_

### 4. Describe the Issue

The city screen does not show yield icons on tiles.
Image
Can be bypassed by enabling yield icons through map options (or the Y key), which shows all yield icons on the map, as usual.

_No response_

_No response_

### 7. Crash dump and crashes.log (ALWAYS ATTACH IF REPORTING A CRASH)

_No response_

vaderkos:
Should fix https://github.com/LoneGazebo/Community-Patch-DLL/issues/13082

### DIFF
--- a/(1) Community Patch/Core Files/Overrides/YieldIconManager.lua
+++ b/(1) Community Patch/Core Files/Overrides/YieldIconManager.lua
@@ -15,6 +15,7 @@ local civ_map_get_plot_by_index = Map.GetPlotByIndex
 
 local Stack = CPK.Util.Stack
 local Always = CPK.FP.Always
+local IsNumber = CPK.Type.IsNumber
 
 local MakeInstance = CPK.UI.Control.Instance.Make
 local MoveInstance = CPK.UI.Control.Instance.Move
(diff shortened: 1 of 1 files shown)


(1) Community Patch/Core Files/Overrides/YieldIconManager.lua
```diff
@@ -287,35 +313,30 @@ end
...
-local RequestViewPlots = LuaEvents.RequestViewPlots
-local RespondViewPlots = LuaEvents.RespondViewPlots
-
-local RequestRender = function()
-	if updates_enabled then RequestViewPlots(CALLER) end
-end
 
 RespondViewPlots.Add(function(caller, plots)
-	if caller == CALLER then
+	if is_listening and caller == CALLER then
```
vaderkos: This fixes race condition when hovering fast over city-banner that @schnetziomi5 mentioned

PR #13096, opened 2026-06-05
Fix non-EUI load screen error for unique improvements

axatin:


Hokath:
what happened here?

(btw title says projects but code seems to be improvements, is it?)

axatin:
some github error, triggering a re-run of the workflows fixed it.

yeah, true, it was an improvements query that was wrong, so the processing of the unique projects later in the file was never reached

Issue #13085, opened 2026-06-01
Game Crashes


Tarzan737:
### 1. Mod version

5.3.0

### 2. Installed Components

Vox Populi Non-EUI

### 3. Additional mods used

[mods.zip](https://github.com/user-attachments/files/28477374/mods.zip)

### 4. Describe the Issue

Game crashes

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[AutoSave_Post_0104 AD-0880.zip](https://github.com/user-attachments/files/28477408/AutoSave_Post_0104.AD-0880.zip)

### 6. Logs (ALWAYS ATTACH THESE IF POSSIBLE)

[Logs.zip](https://github.com/user-attachments/files/28477438/Logs.zip)

### 7. Crash dump and crashes.log (ALWAYS ATTACH IF REPORTING A CRASH)

<img width="1680" height="1050" alt="Image" src="https://github.com/user-attachments/assets/78710d04-51b5-424d-8b62-80e4e68b514c" />

schnetziomi5:
Can you please upload the minidump?

schnetziomi5:
<img width="1340" height="716" alt="Image" src="https://github.com/user-attachments/assets/fe01ae46-40be-405e-ae67-1a875bc17871" />

@axatin 
it throws here
`kTradeConnection.m_aPlotList[kTradeConnection.m_iTradeUnitLocationIndex]` returns `NULL`
presumably linked to #13072 ...

PR #13102, opened 2026-06-05
Fix trade route crash (#13085)

axatin:


azum4roll:
That doesn't seem to work?

schnetziomi5:
(Let's continue this discussion here, where it belongs, rather than on discord)

Curious about two things:

image

... but how would a trade route with an empty path not have crashed earlier in that case?


... Secondly according to your fix (and also by some debugging of the issue I did after I encountered it in a game myself) the root cause of the crash that `GetPath(pOriginCity->plot(), pDestCity->plot(), data)` can (sometimes) not find a path found by `GetMultiplePaths(pOriginCity->plot(), vDestPlot, data)`. Is that in accordance with the supposed behavior of these functions or is this in itself a bug?

axatin:
I've seen trade routes with zero plots while debugging the other save, they definitely could happen. Accessing an array out of bounds is undefined behavior, it's unpredictable what happens.

axatin:
> ... Secondly according to your fix (and also by some debugging of the issue I did after I encountered it in a game myself) the root cause of the crash that `GetPath(pOriginCity->plot(), pDestCity->plot(), data)` can (sometimes) not find a path found by `GetMultiplePaths(pOriginCity->plot(), vDestPlot, data)`. Is that in accordance with the supposed behavior of these functions or is this in itself a bug?

The difference is that `GetMultiplePaths` can't use a heuristic cost because there is more than one target. So `GetMultiplePaths` uses Dijkstra instead of A* pathfinding. The heuristic cost for the step finder is `distance * PATH_BASE_COST` , but trade route steps can have a much lower cost than `PATH_BASE_COST` (see TradePathLandCost), so the heuristic is inadmissible and paths can be missed.

Thinking about it, it's better to just change the heuristics...

axatin:
Updated. Now the heuristic for Trade Route steps on land is `PATH_BASE_COST / 4` , which is exactly the lowest possible value  that can be reached

axatin:
Just ran a test game, everything worked fine. I'll make a hotfix

PR #13104, opened 2026-06-05
Finish Animism implementation

Hokath:
was missing yields

note in VP natural wonders cannot be improved, but in modmods they can. I think it's best to let modmods update this one since they can use a PseudoNaturalWonder column but I'm not 100% how


azum4roll:
Helper tables? :)

Hokath:
I've decided helpers aren't necessary in situations like this

azum4roll:
But what if you typo YIELD_FOOD or YIELD_CULTURE?

PR #13109, opened 2026-06-08
ASP rewrite part 1 + 5.3.2 fixes

azum4roll:


(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
```diff
@@ -0,0 +1,1041 @@
...
+
+include("VPUI_core");
+
+local VP = MapModData and MapModData.VP or VP;
+
+--- Shuffle an array table randomly in place.
+--- @generic T
+--- @param t T[]
+--- @see GetShuffledCopyOfTable
+function Shuffle(t)
```
vaderkos, line 79: Is this deterministic random? I ask cause resource placement sometimes causes desync in mp.

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
```diff
@@ -0,0 +1,1041 @@
...
+	for row in GameInfo.Civilization_Start_Place_First_Along_Ocean{CivilizationType = civType} do
+		if row.PlaceFirst then
+			return true;
+		end
+	end
+	return false;
+end
+
+--- Detect if a civ has Start Region Priority needs and return the count
+function GetNumStartRegionPriorityForCiv(civType)
```
vaderkos, line 464: `count(*)` always returns so it can be collapsed to just, but the matter of preference :D

```lua
return DB.Query("select count(*) as count from Civilization_Start_Region_Priority where CivilizationType = ?", civType)().count
```

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
```diff
@@ -0,0 +1,1041 @@
...
+	return false;
+end
+
+--- Designed to operate on array tables. Does not affect original table.
+--- @generic T
+--- @param incoming_table T[]
+--- @return T[]
+--- @see Shuffle
+function GetShuffledCopyOfTable(incoming_table)
+	local copy = {};
```
vaderkos: What about using `CPK.Table.Copy` here?
Simplifies to
```lua
return Shuffle(TableCopy(incoming_table))
```

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
```diff
@@ -0,0 +1,1041 @@
...
+			table.insert(table_of_indices, loop);
+		end
+	end
+	return bFoundValue, iNumTimesFoundValue, table_of_indices;
+end
+
+--- For debugging purposes. LOTS of table data being handled here.
+function PrintContentsOfTable(incoming_table)
+	print("--------------------------------------------------");
+	print("Table printout for table ID:", table);
```
vaderkos: maybe incoming_table should be passed here?

(1) Community Patch/Core Files/Overrides/Includes/FLuaVector.lua
```diff
@@ -31,16 +31,6 @@
 		return
 	end
 
-	--- @class Vector2
```
vaderkos, line 34: Why removed?

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
```diff
@@ -0,0 +1,1041 @@
+------------------------------------------------------------------------------
```
vaderkos, line 1: As I understand that's a copy of base file + your optimizations or just comments and types @azum4roll ?

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
```diff
@@ -0,0 +1,1041 @@
...
+		iNorthY = iH - 1;
+		iHeight = iH;
+	else -- 1 gap in the middle; the section wraps
+		bWrapY = true;
+		iSouthY = iBottomGapN + 1;
+		iNorthY = iBottomGapS - 1;
+		iHeight = iNorthY + iH - iBottomGapN;
+	end
+
+	---[[ Log dump for debug purposes only, disable otherwise.
```
vaderkos, line 308: Is this uncommented intentionally?

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
```diff
@@ -0,0 +1,1041 @@
...
+	end
+end
+
+-- Memoize table of plots
+local _plots = {};
+
+--- Get an iterator of all plots, optionally shuffled with the passed `Sort` function
+--- @param Sort fun(table)? # Function that modifies a table in place, used on plot indices
+--- @return fun(): PlotId, Plot
+function Plots(Sort)
```
vaderkos, line 94: Unnecessary GC pressure here.

`_indices` is rebuilt from scratch on every call. For `N` plots that's roughly `16B * N` in vanilla Lua 5.1.4 or `8B * N` in LuaJIT. For a 5000-plot map that's ~78 KB / ~39 KB allocated, filled, and discarded per call and since the contents `[0..NumPlots-1]` never change between calls, all of this work after the first call is pure waste.

I propose such changes, so we allocate indices only if sort function is passed.

```lua
local _plots = nil ---@type nil | table<integer, Plot>

function Plots(Sort)
	local max = Map.GetNumPlots()
	local cur = 0

	if not _plots then
		_plots = table.fill(nil, max)

		for i = 0, max - 1 do
			_plots[i + 1] = Map.GetPlotByIndex(i)
		end
	end

	if not Sort then
		return function()
			cur = cur + 1

			if cur > max then return nil end

			return cur - 1, _plots[cur]
		end
	end

	local indices = table.fill(nil, max)

	for i = 1, max do
		indices[i] = i - 1
	end

	Sort(indices)

	return function()
		cur = cur + 1
		
		local idx = indices[cur]

		if not idx then return nil end

		return idx, _plots[idx + 1]
	end
end
```

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
```diff
@@ -0,0 +1,1041 @@
...
+local tAdjustmentForEvenY = {{0, 1}, {1, 0}, {0, -1}, {-1, -1}, {-1, 0}, {-1, 1}};
+local tAdjustmentForOddY = {{1, 1}, {1, 0}, {1, -1}, {0, -1}, {-1, 0}, {0, 1}};
+
+--- @param pPlot Plot
+--- @param iMinR integer
+--- @param iMaxR integer?
+--- @param bReturnPlots true
+--- @return fun(): Plot, integer
+--- @overload fun(pPlot: Plot, iMinR: integer, iMaxR: integer?, bReturnPlots: false): fun(): integer, integer, integer
+local function Plot_GetPlotsOrCoordinatesInCircle(pPlot, iMinR, iMaxR, bReturnPlots)
```
vaderkos, line 636: Roughly ~1-1.5 KB of garbage per call. That's a lot of overhead for something that, for a fixed radius range, is a fully deterministic sequence of positions - no suspension/resumption is (actually?) needed.

**BUT**, because it's rather called once per player on the map nothing to do here

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
```diff
@@ -0,0 +1,1041 @@
...
+		end
+	end
+	return avoidRegionTypes;
+end
+
+--- Test whether `value` exists in `table`
+--- @param table table
+--- @param value any
+--- @return boolean
+function TestMembership(table, value)
```
vaderkos, line 535: This one says: `I'll eat your CPU, nom-nom` XD. Not especially because it itself is bad, but rather how it's used in `AssignStartingPlots`.

As I see `TestMembership` is called to answer "have I seen this ID before," while also maintaining a parallel table keyed by that exact ID (`landmass_fert[id]`, `area_fert[id]`, `land_area_plot_count[id]`, `self.landmass_areas[id]` ...). That keyed table is nil until the ID is first seen and non-nil afterward - i.e. it already is an O(1) membership test. `TestMembership` is redundant work layered on top of data we're building anyway...

(2) Vox Populi/Core Files/Overrides/AssignStartingPlots.lua
```diff
@@ -605,654 +398,525 @@ function AssignStartingPlots:__Init()
...
-				if not TestMembership(landmass_IDs, iLandmass) then -- This plot is the first detected in its landmass.
-					table.insert(landmass_IDs, iLandmass);
-					self.landmass_areas[iLandmass] = {};
-				end
-				if not TestMembership(area_IDs, iArea) then -- This plot is the first detected in its area.
-					table.insert(area_IDs, iArea);
-					table.insert(self.landmass_areas[iLandmass], iArea);
-					-- print("Area#", iArea, "is a new area in landmass#", iLandmass);
-				end
+	for _, plot in Plots() do
```
vaderkos, line 417: https://github.com/LoneGazebo/Community-Patch-DLL/pull/13109/changes#r3375917683

Something like this should be much better in terms of unnecessary work.

```lua

local area_seen = {};
for _, plot in Plots() do
    if not plot:IsWater() then
        local iLandmass = plot:GetLandmass();
        local iArea = plot:GetArea();

        local areas = self.landmass_areas[iLandmass];
        if not areas then
            areas = {};
            self.landmass_areas[iLandmass] = areas;
        end

        if not area_seen[iArea] then
            area_seen[iArea] = true;
            table.insert(areas, iArea);
        end
    end
end

```

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
azum4roll: Map.Rand should be deterministic random (comes from pregame config "MapRandSeed", or otherwise timeGetTime() if not set). Current Communitu_79a still uses math.random so it causes desyncs (not fixed in this PR).

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
azum4roll: I couldn't use CPK because it used to break in pregame. Maybe it's worth a try now?

(1) Community Patch/Core Files/Overrides/Includes/FLuaVector.lua
azum4roll: Repeated definition in Global.d.lua

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
azum4roll: There are additional functions moved from Communitu_79a, and more freshly-made utilities. You may move some of the non-map-specific functions (e.g. GetRandomFromWeightedTable) to CPK.

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
azum4roll: Yes. No harm in having more logs. It's <= #players / 2 calls per map.

azum4roll:
To be fair, memory usage isn't really a concern during game initiation. It's all released when the game starts, before the containers in CvPlayer get filled.

(1) Community Patch/Core Files/Overrides/Includes/MapmakerUtilities.lua
vaderkos: @azum4roll I'd happily fix issue with CPK if something doesn't work in some contexts, if you ever test it and encounter issues keep me on track :D

vaderkos:
> To be fair, memory usage isn't really a concern during game initiation. It's all released when the game starts, before the containers in CvPlayer get filled.

Agreed, my comments were rather what could be done to make the game load litle bit faster :D

PR #13116, opened 2026-06-10
MPMPM DLL integration

azum4roll:
Build success, not tested. Savegame compatible.

The Lua entry point should be in another mod (Modpack Maker for VP), whether integrated into VP or not.


vaderkos:
Instead of adding more things to Game, can we have it in separate module like `MPMP` and put all the functions here?

azum4roll:
Too advanced for me. I'd like an accessible object before starting a game optimally. Nothing about the code requires a game.

PR #13120, opened 2026-06-11
display CultureRateModifier in tooltips

Hokath:
The global version was there but not the local version. It is used on Royal Collection and Opera House


Hokath:
In game it works already, column must be translated in the DLL (but not the global one)

azum4roll:
GlobalCultureRateModifier is translated but InfoTooltipInclude queries the Building_GlobalYieldModifiers table instead of calling the DLL.

azum4roll:
Hmm CultureRateModifier won't appear in pregame Civilopedia in this case. Will need a fix.

PR #13121, opened 2026-06-12
Stream database in chunks during serialization

axatin:
Reduces memory consumption when loading/unloading mods


schnetziomi5:
What's the purpose of `Civ5SavedGameDatabase.db` actually? For me it has a size of 3kb anyway and contains a database with only a single, empty table...

Edit:
It's the database for data attached to the savegame via the lua api, right? And we need to write it to disk so that the engine can retrieve its data? That's insanity.

axatin:
> It's the database for data attached to the savegame via the lua api, right?

Yes

PR #13139, opened 2026-06-16
Feature Request for TheRealHumanoid -- Automaton change from Projects table

Hokath:
This effect exists in Buildings, Traits, Policies, and even Techs. 
This will add it to Projects also.


Hokath:
[hokath alt text: there is a city screen with only one population and it has just completed a public works. there are 2 citizens working tiles in addition to the city center tile]

azum4roll:
Your image is forever uploading

PR #13141, opened 2026-06-16
script for moving non dll changes to game folders

Alpakinator:
Mostly vibe coded script i use for testing changes. Works correctly for me, but i haven't reviewed the code yet.


azum4roll:
Isn't it better to call the build modinfo script, then consult the modinfo to decide which files to move? And then delete the LUA folders for EUI.

Alpakinator:
True, i'll close this to not waste dev time and i'll after i know ithe script is ok i'll make a new one

PR #13148, opened 2026-06-19
nuke pop loss at least 1 and 2

Alpakinator:
Currently nuclear weapons remove x% of remaining population in a city, rounded down. With strategic defense system this means that population loss from subsequent strikes is sharply dropping, even with high rng damage rolls. To the point the population loss is often below 1 when attacking a 7-5 population city, rounded down to 0, making it literally impossible to destroy a city with 1000 nuclear missiles fired in a single turn. Cause city needs to drop below 5 pop to be destroyed.

To address this problem, this PR makes it so that lvl 2 nuke (nuke missile) always lowers city pop by at least 2, and lvl 1 nuke (atomic bomb, or partially intercepted missile) by at least 1.

An alternative is to always round up pop loss but it'll strengthen earlier strikes by +1.

While it seems a balance change that should go through congress, this in practice is a flavor change for Armageddon LARP, cause noone nukes cities to destruction, it's much more practical to nuke a city 2 times and conquer it normally cause it has 0 hp.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvUnitCombat.cpp
+++ b/CvGameCoreDLL_Expansion2/CvUnitCombat.cpp
@@ -2631,7 +2631,7 @@ uint CvUnitCombat::ApplyNuclearExplosionDamage(const CvCombatMemberEntry* pkDama
 					iNukedPopulation *= std::max(0, (pkCity->getNukeModifier() + 100));
 					iNukedPopulation /= 100;
 
-					pkCity->changePopulation(-(std::min((pkCity->getPopulation() - 1), iNukedPopulation)));
+					pkCity->changePopulation(-(std::min(std::max(iNukedPopulation, iDamageLevel), (pkCity->getPopulation() - 1))));
 
 					// Add damage to the city
 					pkCity->setDamage(kEntry.GetFinalDamage());


Alpakinator:
It's still a problem that you need around 15 nuclear missiles to destroy a 30 population city with strategic defense system, how better to address this?

Hokath:
Why's it a problem that Strategic Defense System makes it hard to nuke cities to death?

Alpakinator:
> Why's it a problem that Strategic Defense System makes it hard to nuke cities to death?

Because it's impossible as i mentioned above, completely flavorless. And with this pr, still requiring 10-20 nuclear missiles to do so is also flavorless. Completely unrealistic and makes these weapons feel weak despite massive building cost.

CvGameCoreDLL_Expansion2/CvUnitCombat.cpp
```diff
@@ -2641,8 +2628,10 @@ uint CvUnitCombat::ApplyNuclearExplosionDamage(const CvCombatMemberEntry* pkDama
 
 					int iNukedPopulation = pkCity->getPopulation() * (iBaseDamage + iRandDamage1 + iRandDamage2) / 100;
 
-					iNukedPopulation *= std::max(0, (pkCity->getNukeModifier() + 100));
-					iNukedPopulation /= 100;
+					const int iNukePopModifier = std::max(0, pkCity->getNukeModifier() + 100);
+					iNukedPopulation = iNukedPopulation * iNukePopModifier / 100;
+					if (iNukePopModifier)
+						iNukedPopulation = std::max(iNukedPopulation, 1 + (iDamageLevel > 1));
```
RecursiveVision: I would do `min(max(iNukedPopulation, iDamageLevel), pkCity->getPopulation() - 1)`

RecursiveVision: Generally you should avoid using integers as booleans. Bad practice.

Alpakinator: True, i force pushed and removed the iNukePopModifier check i added, it was useless.

Hokath:
But isn't the point of Strategic Defense System that it would let you survive a nuclear war?

RecursiveVision:
I don't recall that rationale being stated at the time G included it. It's intended to make nukes much more survivable, but not to grant you the ability to withstand 1000 nukes.

Alpakinator:
Afaik this is vanilla nuke behaviour. Bomb shelter has the same -75% modifier and nuke dmg. formula is the same.

Alpakinator:
> But isn't the point of Strategic Defense System that it would let you survive a nuclear war?

Yes, but surviving a nuclear war and the point I made is almost unrelated. Surviving a nuclear war and having indestructible cities is not related. I said i want to fix an edge case. nukes still drOp city hp to almost 0 so they're easy to conquer and destroy buildings. I only want to fix the effect of citizens being immune to nukes at some point. The building after this PR still makes population loss very low to the point razing cities with nukes is a total waste of production.

FWIW. IRL strategic defense system is not a thing, it's science fiction. it's impossible even now to intercept a serious amount of ICBM warheads, and no city has nuke bunkers enough to protect a noticeable amount of population from more than a few explosions.

azum4roll:
It's just unfun for the receiving end if there's no counter like in real life, especially since we don't have MAD implemented.

It doesn't matter how many nukes your enemy has if you can declare war, drop nukes, and take the city all on the same turn.

Alpakinator:
> It's just unfun for the receiving end if there's no counter like in real life, especially since we don't have MAD implemented.
> 
> It doesn't matter how many nukes your enemy has if you can declare war, drop nukes, and take the city all on the same turn.

I agree, but it's also unrelated to this PR. Its vanilla and current VP behaviour you're describing, and this PR won't change that in any way.

schnetziomi5:
> It doesn't matter how many nukes your enemy has if you can declare war, drop nukes, and take the city all on the same turn.

Guess that should read as take/destroy. Because the destroy part is what's changed with this PR

Then again land based nuclear missiles are not necessarily that important in the MAD equation. Submarines carrying nuclear missiles are really the final deterrent. Meanwhile I have heard people calling nuclear missile silos _nuclear sinkholes_, as in you don't want to rely on them for retaliation - instead you hope your enemy targets those sites over populated land with their nukes.

The nuclear weapons game is poor in Civ5/VP. This PR won't fix that but not being able to destroy cities with repeated nuclear strikes is silly, so I am in support of the proposed change.

Hokath:
Probably want to increase % nuke kills and then also lower % bunker prevents.
We can discuss at congress in sept

This PR is totally fine by me

PR #13149, opened 2026-06-19
integrate modpack maker

azum4roll:


Cyclopia31:
Hi azum, thank you a lot for your integration of MPMP to VP !

I have one question, 

Would it be possible to bypass FireTuner by replacing the "CreateMP()" command with, for example, an in-game keyboard shortcut like it was done for the observer mod (ctrl+maj+o) ?

Installing and configuring FireTuner is, I think, the step that discourages people from creating their own modpacks. Another benefit of bypassing FireTuner (as in axatin's version of MPMP) is that it also allows for (theoretically) Linux (Wine/Proton) compatibility.

Thank you again

azum4roll:
I don't know about Linux civ5 enough, but I imagine we'll need different code to grab the MODS folder path (it finds the Documents folder using registry value, which I guess Linux doesn't have).

All other paths are relative and should work on Linux as long as the folder structure of the civ5 app folder is the same.

As for hotkey, that shouldn't be hard but you won't get the satisfaction of seeing it done in real time via logs.

azum4roll:
Hotkey implemented, but you aren't going to know whether it's successful or not.

schnetziomi5:
I am on record saying this is a bad idea, but I haven't had time make an actual argument:

My issue with this is that the conversion of an arbitrary set of enabled/active mods into a DLC will inherently produce a flawed result.
- It's practically impossible to get the virtual file system right. While mods have a load order(if poorly defined), within a given DLC package all scanned files are added to the virtual file system in an undefined order [likely based on their path(directory+filename)?]. When the same file exists within multiple DLCs the priority seems to differ based on the file type: At least the gamecore.dll is determined based on path, while lua and xml files seem to respect the priority field in the civ5.pkg file.

- There is no direct way to query what DLC mods are currently active. This is because the engine considers them to be invalid/locked content(=a DLC you don't own. That it still loads the files into the virtual file system might be a bug). This is a debugging horror. And confusing for the end user too.

Admittedly MPMPM seems to work more or less, but that doesn't make integration into VP a good idea:
- By integrating, especially if directly provided as (5) it's implied that we provide support the creation of modpacks. But how can we support functionality when we don't fully understand the workings behind it, and when we know it's expected to produce undefined results? At most we can add code-tape around specific interactions(like the UI Addins), but at the moment of writing there are already 263 open issues so I would really rather not.

As I see it, this is bogus code added to the main project's dll source when it's known that it is bogus. That can never be a good idea. A officially separate mod maintained by either you or Axatin would be more reasonable. Also, as I announced on Discord, I have found a way to make the engine allow modded multiplayer from the mod menu(pending testing, again haven't really had time) - so that would be an option too. Theoretically it would be even possible to auto-load mods on launch using the same technique - drastically cutting down boot time, which is actually nice about modpacks.

schnetziomi5:
If you want to go ahead with integrating anyway, I would suggest the following:

Each time a modpack is packed:
-  generate random unique GUID for civ5.pkg so that we tell them apart
- Create a lua file that (in some way) exports information about the modpack - e.g. [most importantly] the list of mods
- Give that lua file a name that's reasonable unique for the generated GUID and pack it in the same directory the mods get copied to. ( e.g. `__VPDLCMOD__<32bit hash of guid>.lua` - point is we can find all such files by doing a regular expression `include("__VPDLCMOD__*",true)`
- Also pack a modified main menu that does the above include and: if multiple/none are found shows an error message or if one is found shows the information somewhere in the GUI. Maybe also just disable all submenus except singleplayer,multiplayer and settings...

Edit: 
- and possibly pack the Lua functions bindings in another API table `__vp` that's available without needing a game to be launched.

Cyclopia31:
> Hotkey implemented, but you aren't going to know whether it's successful or not.

Thank you !

azum4roll:
> * Also pack a modified main menu that does the above include and: if multiple/none are found shows an error message or if one is found shows the information somewhere in the GUI. Maybe also just disable all submenus except singleplayer,multiplayer and settings...

This would be in VPUI, yet another DLC mod...

> * and possibly pack the Lua functions bindings in another API table `__vp` that's available without needing a game to be launched.

I have no idea how to do this. I'd love to be able to access the DLL without starting a game.

azum4roll:
Either way it's currently the only released way to play multiplayer, and the integration to DLL makes it much less janky.

azum4roll:
> * within a given DLC package all scanned files are added to the virtual file system in an undefined order [likely based on their path(directory+filename)?]

The UI files obey the UISkin order in .Civ5Pkg

The XML database files seem to be loaded based on directory order?

schnetziomi5:
> This would be in VPUI, yet another DLC mod...

I was suggesting to pack it with the newly created Modpack. But actually deploying it by default would be even better. Can't be VPUI though, because that one isn't installed with CP only. 

Let me stress here that I am not against DLC-mods in general. The point is that the conversion of an arbitrary set of loaded into a [single] dlc-mod will not give stable results. The ways mods and dlcs are loaded are too different for that to work correctly in all(or even most) cases.

> I have no idea how to do this. I'd love to be able to access the DLL without starting a game.

I am planning to eventually introduce a `__vp` lua API available without a game active, we can move it there later I guess


> The XML database files seem to be loaded based on directory order?

The database files obey the order they are listed in the various civ5.pkg files of real DLCs. MPMPM overrides the individual 
files in virtual file system. 

> The UI files obey the UISkin order in .Civ5Pkg

That something is off here, is actually known, that's why deletion of some files in the mods folder(All files with InGameUI addins) is a required step. Since you already made edits to all our files that have Addins why not change them so that the query the civ5 API and additionally some database table, and at modpack creation time you copy the InGame addin info from the API into the table. Then you wouldn't have to create and pack an edited version of the files.

Also note this comment on Discord:
image

azum4roll:
Directory names matter for database files. The modpack folder needs to be loaded after Expansion2. But otherwise it should work.

axatin:
Is it intended that it's not included in the autoinstaller?

azum4roll:
No.

PR #13151, opened 2026-06-20
Un-hardcode UI placement site eval

Hokath:
this fixes some incorrect site eval in VP. For example Kasbah was assumed to be placed on Desert, but it no longer is, yet Morocco was still trying to settle Desert. 

It's not easy to completely un-hardcode the inca UI without making the check even more expensive, but this fixes it for current VP

This loop of improvements does make the check a bit more expensive unfortunately. Intentionally avoided putting in VP Polder's GetRequiresXAdjacentLand condition for this reason.


CvGameCoreDLL_Expansion2/CvSiteEvaluationClasses.cpp
```diff
@@ -611,95 +620,64 @@ int CvSiteEvaluatorForSettler::PlotFoundValue(CvPlot* pPlot, const CvPlayer* pPl
...
-		static ImprovementTypes eBrazilImprovement = (ImprovementTypes)GC.getInfoTypeForString("IMPROVEMENT_BRAZILWOOD_CAMP", true);  
-		if(eBrazilImprovement != NO_IMPROVEMENT)
+		// Custom code for UIs
+		for (int iImprovementLoop = 0; iImprovementLoop < GC.getNumImprovementInfos(); iImprovementLoop++)
 		{
-			CvImprovementEntry* pkEntry = GC.getImprovementInfo(eBrazilImprovement);
-			if(pkEntry != NULL && pkEntry->IsSpecificCivRequired())
+			ImprovementTypes eImprovement = (ImprovementTypes)iImprovementLoop;
+			CvImprovementEntry* pkEntry = GC.getImprovementInfo(eImprovement);
+			if (pkEntry != NULL && pkEntry->IsSpecificCivRequired())
```
azum4roll: It can never be NULL

PR #13167, opened 2026-06-26
NewCustomModOptions clean up

nobubblegums:
Modmod custom options for LOCAL_GENERALS_NEAREST_CITY and RELIGIOUS_SETTLERS_WEIGHTED were added as default on. 
They both depend on other modmods (GLOBAL_LOCAL_GENERALS and GLOBAL_RELIGIOUS_SETTLERS), so they are inactive. Better to have them default off to avoid confusions.


RecursiveVision:
Do not agree with this, I left them set to 1 intentionally because they're good defaults. Like other 'setting' values, they are under the relevant mod. If someone is going to modify the game files, I think it's fair to assume they're not just a casual player.

nobubblegums:
Oh my bad then, one of the mas my modmod and I assumed I left it on by mistake.

PR #13180, opened 2026-07-02
Problem with EventLinks for Spy Mission Blocking

Hokath:
Received this report
https://forums.civfanatics.com/threads/new-beta-version-%E2%80%93-5-3-3-june-15-2026.703525/post-16966679

Looking at the code, the structure is like
```
Loop over players
  check if player != this player for CheckKnownPlayers == false
    continue
  Loop over cities
     check if city != this city for CheckOnlyEventCity == true
       continue
     look for events/choices and return false if you find them
```

so it seems that contrary to the documentation, the combination actually works just fine?


Hokath:
So now I'm testing the status quo (because I'm sure I already tested that this works so I'm a bit confused)

My spy in a City with a Counterspy on Establish Network shows
image

Hokath:
Ah, I found the reason, it's not in base VP

PR #13185, opened 2026-07-04
Some religion fixes

Hokath:
These are for modmods, I don't think they are used in base VP (Shrine gives happiness from Goddess of Home but since it's always unlocked you will never hit that code)

I was also trying to find why AI sometimes picks God of the Sun even when it doesn't have farm resources, but couldn't reproduce it.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlayer.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlayer.cpp
@@ -26699,7 +26699,7 @@ void CvPlayer::doInstantYield(InstantYieldType iType, bool bCityFaith, GreatPers
 					{
     					PlayerTypes ePlotOwner = pUnit->plot()->getOwner();
 						bool bOwnedTerritory = (ePlotOwner != NO_PLAYER && ePlotOwner == pUnit->getOwner());
-						iYieldPer100HP += pReligion->m_Beliefs.GetYieldPerHeal(eYield, GetID(), pLoopCity, true, bOwnedTerritory);
+						iYieldPer100HP += pReligion->m_Beliefs.GetYieldPerHeal(eYield, GetID(), pLoopCity, false, bOwnedTerritory);
 
 						if (iYieldPer100HP > 0)
 						{
--- a/CvGameCoreDLL_Expansion2/CvReligionClasses.cpp
+++ b/CvGameCoreDLL_Expansion2/CvReligionClasses.cpp
@@ -8357,7 +8357,12 @@ int CvReligionAI::ScorePantheonBeliefAtCity(CvBeliefEntry* pEntry, CvCity* pCity
 
 						int iEraNeeded = pkTechInfo->GetEra();
 						int iCurrentEra = m_pPlayer->GetCurrentEra();
-						iAvailabilityModifier = max(0, 3 - (iEraNeeded - iCurrentEra));  // lose remaining value if we have to wait
+						iAvailabilityModifier = 3 - (iEraNeeded - iCurrentEra);  // lose remaining value if we have to wait
+						if (!pCity)
+						{
+							iAvailabilityModifier--;
+						}
+						iAvailabilityModifier = max(0, iAvailabilityModifier);
 					}
 				}
 			}
@@ -8667,11 +8672,12 @@ int CvReligionAI::ScorePantheonBeliefAtCity(CvBeliefEntry* pEntry, CvCity* pCity
 
 							int iEraNeeded = pkTechInfo->GetEra();
 							int iCurrentEra = m_pPlayer->GetCurrentEra();
-							iAvailabilityModifier = max(0, 3 - (iEraNeeded - iCurrentEra));  // lose remaining value if we have to wait
-						}
-						if (!pCity)
-						{
-							iAvailabilityModifier--;
+							iAvailabilityModifier = 3 - (iEraNeeded - iCurrentEra);  // lose remaining value if we have to wait
+							if (!pCity)
+							{
+								iAvailabilityModifier--;
+							}
+							iAvailabilityModifier = max(0, iAvailabilityModifier);
 						}
 					}
 				}
--- a/CvGameCoreDLL_Expansion2/CvReligionClasses.cpp
+++ b/CvGameCoreDLL_Expansion2/CvReligionClasses.cpp
@@ -8416,11 +8416,16 @@ int CvReligionAI::ScorePantheonBeliefAtCity(CvBeliefEntry* pEntry, CvCity* pCity
 			iTempValue += 2 * min(iCap, iExpectedGrowth / pEntry->GetFollowerRequiredPerYield(iI));
 		}
 
-		// caps at half number of followers. assume we are at cap.
+		// caps at half number of followers.
 		if (pEntry->GetYieldPerGPT(iI) > 0)
 		{
-			iTempValue += 10 * (iCurrentCityPop / 2) ;
-			iTempValue += 5 * (iExpectedGrowth / 2);
+			if (pCity)
+			{
+				iTempValue += 10 * min((iCurrentCityPop / 2), (pCity->getYieldRateTimes100((YieldTypes)iI, false) / (pEntry->GetYieldPerGPT(iI) * 100)));
+				int iNewPop = iCurrentCityPop + iExpectedGrowth;
+				// assume yields will increase proportionally to pop
+				iTempValue += 5 * min(iNewPop / 2, (iNewPop * pCity->getYieldRateTimes100((YieldTypes)iI, false) / (pEntry->GetYieldPerGPT(iI) * 100 * iCurrentCityPop)));
+			}
 		}
 
 		// yield per birth


azum4roll:
Thoughts on refactoring the secondary pantheon and permanent pantheon code (which is supposed to be called for every belief, currently missing in a lot of places) into a common function?

Hokath:
Yes it is extremely annoying and ugly code.

Hmm there are slight differences between calls, which is a bit annoying. I will look into it.

Hokath:
the cpp version is too old to pass functions like lambda, so the documentation suggests to use this thing "template" instead? I have no idea about this before so what do you think?
```
template<typename ReligionFn, typename BeliefFn>
int CvCityReligions::SumBeliefSources(ReligionFn religionFn, BeliefFn beliefFn) const
{
    int iTotal = 0;

    ReligionTypes eMajority = GetReligiousMajority();
    BeliefTypes eSecondaryPantheon = GetSecondaryReligionPantheonBelief();

    const CvReligion* pReligion = GetMajorityReligion();

    // Standard source of the Majority religion
    if (pReligion)
    {
        iTotal += religionFn(pReligion->m_Beliefs);
    }

    // Secondary pantheon?
    if (eSecondaryPantheon != NO_BELIEF)
    {
        const CvBeliefEntry* pBelief = GC.GetGameBeliefs()->GetEntry(eSecondaryPantheon);

        if (pBelief)
        {
            iTotal += beliefFn(*pBelief);
        }
    }

    //  Permanent pantheon effect (with no-double-counting)
    if (MOD_BALANCE_PERMANENT_PANTHEONS)
    {
        if (GC.getGame().GetGameReligions()->HasCreatedPantheon(GetOwner()))
        {
            BeliefTypes ePantheonBelief = GC.getGame().GetGameReligions()->GetBeliefInPantheon(GetOwner());

            if (ePantheonBelief != NO_BELIEF &&
                ePantheonBelief != eSecondaryPantheon)
            {
                const CvReligion* pMajorityReligion = GetMajorityReligion();

                bool bAlreadyInReligion = (pMajorityReligion && pMajorityReligion->m_Beliefs.IsPantheonBeliefInReligion(ePantheonBelief, eMajority, GetOwner()));

                if (!bAlreadyInReligion)
                {
                    const CvBeliefEntry* pPantheon =GC.GetGameBeliefs()->GetEntry(ePantheonBelief);

                    if (pPantheon)
                    {
                        iTotal += beliefFn(*pPantheon);
                    }
                }
            }
        }
    }

    return iTotal;
}
```

then you have to refactor the code like functions? As in
```
iHappinessFromReligion += GetCityReligions()->SumBeliefSources(
    [&](const CvReligionBeliefs& b)
    {
        int iVal = 0;

        iVal += b.GetHappinessPerCity(getPopulation(), getOwner(), this);

        if (plot()->isRiver())
            iVal += b.GetRiverHappiness(getOwner(), this);

        iVal += b.GetFullyConvertedHappiness(getOwner(), this, false);

        return iVal;
    },
    [&](const CvBeliefEntry& b)
    {
        int iVal = 0;

        iVal += b.GetHappinessPerCity();

        if (plot()->isRiver())
            iVal += b.GetRiverHappiness();

        return iVal;
    },
    getOwner(),
    this
```

azum4roll:
Better to define actual functions for SumBeliefSources parameters just for easier reads.

azum4roll:
I gave Gemini the code of CvCity::getExtraSpecialistYield() and CvCity::GetWonderProductionModifierFromReligion() and asked it to group all the secondary/permanent pantheon code into one single function. Here's the response:

----

Yes, we absolutely can. Since *every single belief effect check* follows this identical three-tiered structure:

1. **Majority Religion** (if any)
2. **Secondary Pantheon** (if any)
3. **Permanent Pantheon** (if any, with a double-counting safety check)

We can group them entirely into a single unified execution pipeline.

Because different belief checks query different methods (e.g., one calls `GetSpecialistYieldChange`, another calls `GetWonderProductionModifier`), the cleanest way to do this in C++98 without writing 30 different wrapper functions is to pass a **Helper Functor or Aggregator Class**.

Here is how you can abstract the entire loop away so that your yield/modifier functions become clean, 2-line calls.

---

### Step 1: Define a generic `BeliefProcessor` structure

Add this helper mechanism to `CvCity.h` or a shared header. It collects the valid beliefs a city possesses at this exact microsecond:

```cpp
// Inside CvCity.h or a shared religion header
struct SActiveCityBeliefs
{
    BeliefTypes eMajorityPantheon; // If majority religion is a pantheon
    BeliefTypes eSecondaryPantheon;
    BeliefTypes ePantheonPermanent;
    
    // We also store pointers to the actual belief structures for quick iteration
    int iCount;
    BeliefTypes eBeliefs[4]; // Max possible active beliefs impacting a single stat slot simultaneously

    SActiveCityBeliefs() : eMajorityPantheon(NO_BELIEF), eSecondaryPantheon(NO_BELIEF), ePantheonPermanent(NO_BELIEF), iCount(0) {}
};

```

---

### Step 2: Write the central collector in `CvCity`

This method runs the heavy logic **once**, filtering out double-counting and handling the `MOD_BALANCE_PERMANENT_PANTHEONS` rule. It fills our tiny stack-allocated array with only the valid beliefs that should be evaluated.

```cpp
// Inside CvCity.cpp
SActiveCityBeliefs CvCity::GetActiveBeliefs() const
{
    SActiveCityBeliefs active;
    ReligionTypes eMajority = GetCityReligions()->GetReligiousMajority();
    const CvReligion* pMajorityReligion = (eMajority != NO_RELIGION) ? GetCityReligions()->GetMajorityReligion() : NULL;

    // 1. Accumulate Majority Religion Beliefs (Handles founder/follower/etc.)
    // Note: The majority religion's internal m_Beliefs list is processed natively, 
    // but we track if it uses our pantheon to prevent double dipping later.
    
    // 2. Extract Secondary Pantheon
    if (pMajorityReligion)
    {
        active.eSecondaryPantheon = GetCityReligions()->GetSecondaryReligionPantheonBelief();
        if (active.eSecondaryPantheon != NO_BELIEF)
        {
            active.eBeliefs[active.iCount++] = active.eSecondaryPantheon;
        }
    }

    // 3. Extract Permanent Pantheon (with full validation)
    if (MOD_BALANCE_PERMANENT_PANTHEONS && GC.getGame().GetGameReligions()->HasCreatedPantheon(getOwner()))
    {
        BeliefTypes ePantheonBelief = GC.getGame().GetGameReligions()->GetBeliefInPantheon(getOwner());
        if (ePantheonBelief != NO_BELIEF && ePantheonBelief != active.eSecondaryPantheon)
        {
            // Verify our current majority religion doesn't already give us this exact pantheon bonus
            if (pMajorityReligion == NULL || !pMajorityReligion->m_Beliefs.IsPantheonBeliefInReligion(ePantheonBelief, eMajority, getOwner()))
            {
                active.eBeliefs[active.iCount++] = ePantheonBelief;
            }
        }
    }

    return active;
}

```

---

### Step 3: See how incredibly clean the gamecore functions become

Now, any check function in the game engine drops down to a completely standardized, simplified structure:

#### Clean Version of Block 1 (Specialist Yield)

```cpp
    // Process Majority Religion natively
    ReligionTypes eMajority = GetCityReligions()->GetReligiousMajority();
    const CvReligion* pReligion = (eMajority >= RELIGION_PANTHEON) ? GetCityReligions()->GetMajorityReligion() : NULL;
    if (pReligion)
    {
        iYieldMultiplier += pReligion->m_Beliefs.GetSpecialistYieldChange(eSpecialist, eIndex, getOwner(), GET_PLAYER(getOwner()).getCity(GetID()));
    }

    // Dynamically evaluate all extra sub-pantheons in one single loop execution
    SActiveCityBeliefs activeBeliefs = GetActiveBeliefs();
    for (int i = 0; i < activeBeliefs.iCount; ++i)
    {
        iYieldMultiplier += GC.GetGameBeliefs()->GetEntry(activeBeliefs.eBeliefs[i])->GetSpecialistYieldChange(eSpecialist, eIndex);
    }

```

#### Clean Version of Block 2 (Wonder Production)

```cpp
    // Process Majority Religion natively
    ReligionTypes eMajority = GetCityReligions()->GetReligiousMajority();
    const CvReligion* pReligion = (eMajority != NO_RELIGION) ? GetCityReligions()->GetMajorityReligion() : NULL;
    if (pReligion)
    {
        iModifier = pReligion->m_Beliefs.GetWonderProductionModifier(eEra, getOwner(), this);
    }

    // Process all extra sub-pantheons in the exact same way
    SActiveCityBeliefs activeBeliefs = GetActiveBeliefs();
    for (int i = 0; i < activeBeliefs.iCount; ++i)
    {
        CvBeliefEntry* pBeliefInfo = GC.getBeliefInfo(activeBeliefs.eBeliefs[i]);
        if (pBeliefInfo && eEra < pBeliefInfo->GetObsoleteEra())
        {
            iModifier += pBeliefInfo->GetWonderProductionModifier();
        }
    }

```

### Why this approach rules:

Instead of copying and pasting complex condition chains 50 times across `CvCity.cpp`, you now just pull `GetActiveBeliefs()` and write a simple `for` loop over the results. If a city has no secondary or permanent pantheon, `activeBeliefs.iCount` is `0`, and the loop instantly skips with almost zero performance cost

azum4roll:
There are some other problems though, like how CvReligionBeliefs (with the IsBeliefValid check) and CvBeliefEntry (without the check) are being used interchangeably. We'll need to standardize all of them.

I think for secondary/permanent pantheons, we don't need to check if the beliefs are valid since the same checks are already done in the existing code?

vaderkos:
@azum4roll If you want I have claude pro sub, so I can also ask it if you need?

Hokath:
> we don't need to check if the beliefs are valid since the same checks are already done in the existing code?

Hmmm. Not sure. For example in the wonder modifier code it "manually" checks for an obsolete era
`if ((int)eEra < GC.GetGameBeliefs()->GetEntry(eSecondaryPantheon)->GetObsoleteEra())`

you can imagine for a pantheon that requires the holy city (here: capital) would need to check for that too. e.g. Goddess of Wisdom. So maybe we need to do a IsBeliefValid check in CvCity::GetActiveBeliefs() (and pass in NO_RELIGION so it behaves like a pantheon before religions are founded?)?

azum4roll:
{DBADC29B-5B45-4181-B022-B37949EC7682}

It'll just return false here.

Issue #13145, opened 2026-06-18
3rd Place AI Becoming Voluntary Vassal of 1st Place AI


ron-murhammer:
### 1. Mod version

5.2.7

### 2. Installed Components

Vox Populi with EUI

### 4. Describe the Issue

The 3rd place civ shouldn't ever agree to become a voluntary vassal of the 1st place civ. Seems like either a bug or a gap in the AI logic. This is pretty much game ending.

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[AutoSave_Post_0183 AD-1230.zip](https://github.com/user-attachments/files/29073263/AutoSave_Post_0183.AD-1230.zip)

Hokath:
Did you have the logs?

Specific the AI logs. It would be interesting to know what they grand strategy intention was at the time.

ron-murhammer:
I had other logging on but no AI logging.

RecursiveVision:
I added a No Voluntary Vassalage game option, which doesn't fix this yet but at least provides relief from stupid AI decisions.

axatin:
The current logic is basically: We aren't close to winning, they are our neighbors and long-term friends and they're significantly stronger than us, so we accept vassalage. What I'm missing there is a basic check if protection is necessary at all: Are there any hostile civs that are much stronger than us? If the answer is no, we should not consider voluntary vassalage

PR #13188, opened 2026-07-05
Fix AI offering voluntary vassalage while not under threat (#13145)

axatin:


CvGameCoreDLL_Expansion2/CvDiplomacyAI.cpp
```diff
@@ -56406,6 +56406,28 @@ bool CvDiplomacyAI::IsVoluntaryVassalageAcceptable(PlayerTypes ePlayer)
...
+		if (GET_PLAYER(eLoopPlayer).getTeam() == GetTeam() || GET_PLAYER(eLoopPlayer).getTeam() == GET_PLAYER(ePlayer).getTeam())
+			continue;
+
+		if (GET_PLAYER(eLoopPlayer).GetDiplomacyAI()->GetSurfaceApproach(GetID()) > CIV_APPROACH_GUARDED)
+			continue;
+
+		if (GetMilitaryStrengthComparedToUs(eLoopPlayer) < STRENGTH_POWERFUL)
+			continue;
+
+		bNeedProtection = true;
```
RecursiveVision, line 56427: Add a break; here?

axatin: done

RecursiveVision:
If the objective here is to fix a fairness issue from other civs' perspective, consider using GetRawMilitaryStrengthComparedToUs() instead, as it isn't reduced by the AI's Boldness flavor.

axatin:
> If the objective here is to fix a fairness issue from other civs' perspective, consider using GetRawMilitaryStrengthComparedToUs() instead, as it isn't reduced by the AI's Boldness flavor.

Changed it.

PR #13189, opened 2026-07-05
More Fixes

Hokath:
Fix the Observatory still having conquest probability of a policy building.
Do an sql query instead of hardcoding so that future changes update automatically.
(I think it's better to overwrite the wonders than do the check for wonders twice?)

Add flavors to the Yassa project
if it works like the building flavors then AI mongolia was not building this before... :S

Fix the text on Sri Pada pedia entry since the promo was changed


azum4roll:
It shouldn't be hard to adjust the UI so we can have unlimited number of lines on UA text right? In that case we can work on standardizing all the wordings instead of using short forms to save space.

Hokath:
I'm not brave enough to try change that, but yes. You could even have a scroll bar inside the box, i think

Hokath:
fix #13202

Hokath:
<details>
<summary>looks like this</summary>
image
image
</details>

Hokath:
image
love that. LOVE

(2) Vox Populi/Database Changes/Units/NewUnits.xml
```diff
@@ -1226,6 +1226,7 @@
 			<CivilianAttackPriority>CIVILIAN_ATTACK_PRIORITY_LOW</CivilianAttackPriority>
 			<NumInfPerEra>15</NumInfPerEra>
 			<RestingPointChange>30</RestingPointChange>
```
azum4roll, line 1228: This one belongs in Iroquois.sql, not here.

PR #13198, opened 2026-07-09
Generalized belief sources code for pantheon effects

Hokath:
Starting the implementation of cleaner pantheon effect code blocks as discussed in PR #13185

For player level effects we'd need another struc, but I think that only appears once in pantheons iirc?


Hokath:
I should remove these obsolete era checks because they aren't used and the way they're used for the wonder modifier is on the building era, not the player era.

Edit: done, will rebase later

azum4roll:
Maybe the struct and the new function fit more in CvCityReligions.

Hokath:
yes good point

I realize we will also need this in DoInstantYield player level function

PR #13215, opened 2026-07-14
Fix permanent pantheon projected improvement yields being double-counted

TheKineticz:
With permanent pantheons on, worker tooltips and builder AI projected yields were calling `getYieldWithBuild` twice and adding the results, so the whole projected yield got counted again. Actual tile yields were already fine because they pass the pantheon in through the proper `pPlayerPantheon` argument.

This makes the projection path do the same thing: resolve the pantheon once, pass it into a single `getYieldWithBuild` call, and only layer on the pantheon-specific religion yields. Also lined up pantheon nature yields in that helper with how real yields are calculated.

Bug isolated by @Hokath at [https://forums.civfanatics.com/threads/religion-permanent-pantheons.670446/#post-16852120](https://forums.civfanatics.com/threads/religion-permanent-pantheons.670446/#post-16852120)


RecursiveVision:
This will conflict with https://github.com/LoneGazebo/Community-Patch-DLL/pull/13198

Hokath:
merge this first and ill update my PR accordingly -- it's far from done atm

p.s. nice job!

PR #13221, opened 2026-07-16
Congress 10 mergables

azum4roll:
Untested but buildable


CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -12921,6 +12927,14 @@ void CvPlayer::receiveGoody(CvPlot* pPlot, GoodyTypes eGoody, CvUnit* pUnit)
 		}
 	}
 
+	// Free promotion
+	if (kGoodyInfo.getPromotionType() != NO_PROMOTION)
+	{
+		ASSERT(pUnit);
+		PromotionTypes ePromotion = kGoodyInfo.getPromotionType();
```
axatin: This doesn't compile, getPromotionType() returns int, not PromotionTypes

azum4roll: Maybe the fix got snuck into the other commit...

PR #13241, opened 2026-07-23
Congress 10

azum4roll:
Old PR seems broken, so here's a new one.


CIVITAS-John:
**Straight from Opus 5.5. Not myself.**

I found six problems in PR #13241 (the Recon rework). Two are serious: a probable crash and a reversed Sapper bonus. Nothing was compiled or run; all findings come from reading the diff against the 5.4.6 base (dcb33a65). I checked #1 and #2 against the code myself; #3–#6 came from a review subagent and I haven't checked them yet.

High

1. Stale aura entries can crash the game (CvUnit.cpp:34349, UpdateAreaEffect)
When a unit stops having any aura, the code sets SetAreaEffectCacheIndex(bHasEffect ? AddOrReplace(...) : -1). The unit's index becomes -1, but its old entry stays in the player's list of aura units.
- Example: a land anti-air unit embarks, so it can no longer intercept. Its entry stays in the list. If it then dies, kill() and startDelayedDeath() see index -1 and skip removing the entry.
- Crash: the next nearby air attack runs GetBestInterceptor() or HasAirCover(), gets NULL for the dead unit, and dereferences it. RemoveAreaEffectUnit can crash the same way when it swaps a stale entry into place.
- Jammers: a Jammer that loses its aura promotion leaves plots marked as jammed.
- Duration: the list is only rebuilt when that player's turn starts, so the bad entry lasts through every other player's turn.
- Fix: remove the existing entry when bHasEffect is false and the index is ≥ 0.

2. The Sapper bonus is backwards (CvPlayer.cpp:47252)
- Before: full SAPPED_CITY_ATTACK_MODIFIER when closer than max range, half at exactly max range.
- Now: full at max range, half when closer.

So a Sapper next to a city gives half the bonus, and one 2 tiles away gives the full bonus. Swapping the == and < branches fixes it.

Medium

3. Jammers do nothing after loading a save (CvPlot.cpp:12093)
The per-plot jamming flag m_bfJammingActive is never saved or loaded, and CvPlot::reset doesn't clear it. The load path calls UpdateAreaEffectUnits, which never rebuilds jamming. After a load, every Jammer is inactive until it moves or changes promotions. That changes interception results and could desync multiplayer games.

4. Heal rate is reported for units that can't heal (CvUnit.cpp:7876-7878, ActualHealRate)
Flat heal, terrain heal, the city's always-heal and passive area heal are now added outside the canHeal() check. That check used to block units at full HP, hovering units over water, units on mountains and units that can't end their turn on the plot. As a result:
- Free yields: doHeal sees a rate above zero for full-HP units, such as any land unit in a city with always-heal. The "yields from healing" bonus then fires every turn with no HP restored.
- AI decisions: AI checks of the form ActualHealRate(...) == 0 or > 0 now give wrong answers.

Low

5. Interceptor ranking uses outdated HP (CvUnit.cpp:34338)
iInterceptValue includes current HP for air units, but it is now cached and only refreshed on moves, promotions, embarking or the intercept mission. A fighter damaged by an air sweep keeps its full-health ranking in GetBestInterceptor. The old code worked this out fresh on every call.

6. Possible divide-by-zero in AI promotion scoring (CvUnit.cpp:33372)
dTerrainRatio = iTerrainCount / iPlotCount can divide by zero. Two cases:
- a barbarian unit, or a player with no cities, outside any territory;
- an air unit of a landlocked player scoring water terrain.

The result is NaN, and converting NaN to int is undefined, which garbles that unit's promotion scores.

I didn't review the balance and SQL/XML/text changes in depth. The checkout is in F:\vox-deorum\vox-populi-review\repo on branch pr-13241.

Hokath:
Copying from discord for the record.

High starting recon XP allows Zulu (and anyone else with yields from level up) to spam recon units and generate extreme yields.
image

Easiest fix: extend the recent no-heal on train check to a no-yields on train check as well (?)

PR #13257, opened 2026-07-29
ubsan and other fixes

JohnsterID:


CvGameCoreDLL_Expansion2/CvDealAI.cpp
```diff
@@ -2227,7 +2227,8 @@ int CvDealAI::GetCityValueForDeal(CvCity* pCity, PlayerTypes eAssumedOwner, bool
 	else if (pCity->getOwner() != eAssumedOwner)
 	{
 		// reduce value of a city we're buying if we won't have an internal border with it based on distance to our closest city, up to half of the value
-		iItemValue -= min(iItemValue / 2, goldPerPlot * 5 * iMinDistance);
+		// iMinDistance is INT_MAX if the buyer has no other city - use 64 bit math to avoid overflow (the min() takes care of the rest)
+		iItemValue -= (int)min((long long)iItemValue / 2, (long long)goldPerPlot * 5 * iMinDistance);
```
azum4roll: Isn't it better to have a separate condition for INT_MAX? We should never perform arithmetic on it.

CvGameCoreDLL_Expansion2/CvTechAI.cpp
```diff
@@ -347,7 +347,12 @@ void CvTechAI::ReweightByCost(CvPlayer *pPlayer, bool bWantsExpensive)
 		int iNewWeight = 0;
 		if (bNeedExpensiveTechs)
 		{
-			iNewWeight = int(double(m_ResearchableTechs.GetWeight(iI)) * fWeightDivisor);
+			// the product can exceed INT_MAX (or be infinity when iTurnsLeft is the
+			// INT_MAX "cannot research" sentinel): saturate in the double domain,
+			// since an out-of-range double-to-int cast is undefined behavior and
+			// would invert the intent by turning a huge weight into INT_MIN
+			double fNewWeight = double(m_ResearchableTechs.GetWeight(iI)) * fWeightDivisor;
+			iNewWeight = fNewWeight >= double(INT_MAX) ? INT_MAX : int(fNewWeight);
```
azum4roll: This could need a further rewrite. bNeedExpensiveTechs is basically when AI is considering which free tech to take, so it wants the most expensive one that also provides the best benefit. The function as is would tie all expensive techs at 10000 weight, regardless of their original score.

JohnsterID: Yes, flattening at 10000, the cap already tied every tech whose weight × pow(turnsLeft, …) product landed anywhere in [10000, INT_MAX), which is virtually all expensive techs (pow(60, 2.3) ≈ 12000, so even base weight 1 saturates). The commit only changed the overflow cases, previously the UB cast produced INT_MIN and the < 1 clamp sent exactly the most expensive techs to weight 1. Now they tie at the cap like the rest.
Rewrite so the original score still differentiates among expensive techs? e.g. two passes: compute the products in double, then rescale by the max (10000 * product / maxProduct, min 1) instead of clamping. That keeps cost dominant, preserves relative ordering, and stays within the existing [1,10000] range. Another PR? Or add to this one?

azum4roll: There's no need to cap the weight at 10000. It can safely have values up to INT_MAX / 100.

`bool bNeedExpensiveTechs = pPlayer->getOverflowResearchTimes100() > (pPlayer->GetScienceTimes100() * 2);`

This line is not right either. AI shouldn't want the most expensive tech (that can theoretically take forever) just because it has more than 2 turns of science overflow.

I think the function needs a full rewrite.

CvGameCoreDLL_Expansion2/CvDealAI.cpp
JohnsterID: Fixed in eb31d849b

CvGameCoreDLL_Expansion2/CvTechAI.cpp
JohnsterID: Ok, pushed a potential rewrite to techai-reweight-rewrite (2d1b626) for discussion before it goes anywhere.

Should it be overflow-completes-cheapest (current draft), overflow-completes-that specific tech per-candidate, free-tech-only (drop the overflow path entirely), or some other threshold? Then we can adjust the formula weighting as well if required.

azum4roll: Overflow research doesn't seem to be lost if you choose to research a cheap tech, unlike production. So there's no reason to have special handling for it.

JohnsterID: Fixed in ec56b0aee805fcfa8127cad98b2d659e066242ad. Also, GetMaxResearchOverflow (declaration + definition) could be deleted as dead code.

JohnsterID: Added to PR as 565d487c7dfb961ec8844436cb3e2910b3d2d988 and 081de59dddcd91ed9a6d5c3900ce109dea4247b5.

JohnsterID:
There is more todo but I can do it on another PR. Next crash is EXE side so looking at per turn memory tracking to make debugging easier.

CvGameCoreDLL_Expansion2/CvGameCoreUtils.cpp
schnetziomi5, line 432: I genuinely don't understand why we need this code segment a third time. I am also working on debug related stuff and was thinking of combining this into one?

schnetziomi5 (commented):
Also what are all those `#ifdef`s really good for?
`release` and `debug` should be enough. Has anyone _ever_ built the .dll for a target other than `win32`? 
As for `MOD_DEBUG_MINIDUMP` I added a opt-out via a file existence check(`nodumps.please`) just an hour ago on my active branch... 
`VPRELEASE_ERRORMSG` could also just be a runtime check with an opt out for those who run autoplays...?

JohnsterID:
> Also what are all those `#ifdef`s really good for? `release` and `debug` should be enough. Has anyone _ever_ built the .dll for a target other than `win32`? As for `MOD_DEBUG_MINIDUMP` I added a opt-out via a file existence check(`nodumps.please`) just an hour ago on my active branch... `VPRELEASE_ERRORMSG` could also just be a runtime check with an opt out for those who run autoplays...?

https://github.com/Hello71/Community-Patch-DLL/tree/linux
Long term goal would be to enable the Mod for Civ5XP so might as well gate pure win32 as we go. Makes it easier to search later.

I would prefer a build config option similar to DISABLE_CVASSERT. NO_MINIDUMP? If we could add it a runtime option sure, I would keep the default on.

CvGameCoreDLL_Expansion2/CvGameCoreUtils.cpp
JohnsterID: PRECONDITION, ASSERT and crash. I see that as three dialogs.

```
Use ASSERT (always active in debug/release, can disable with DISABLE_CVASSERT):
- To enforce invariants and catch logic bugs in engine or game logic
- When a pointer/value should never be null if the code is correct
- In all places where a failure means a developer mistake, not user action
- Shows a dialog and allows "ignore always" option, then optionally breaks to debugger
- Does NOT crash the game - allows continuation after user acknowledgment
Use PRECONDITION (always active, crashes immediately):
- For input validation and parameter checking at function entry points
- When the condition failure indicates corrupted state that cannot be recovered
- For bounds checking on arrays/indices where out-of-bounds access would be dangerous
- Always crashes with BUILTIN_TRAP() - no user dialog, no continuation possible OR dialog without IGNORE option.
- Use when the program cannot safely continue if the condition fails
```

I already think the crash dialog may too much dev info, we could increase non-devs readability, leave most to the crash log dump?

schnetziomi5: Maybe I wasn't precise, here I am bothered specifically by the three way `#idef` within the function. It's clutter.

```

#if defined(VPRELEASE_ERRORMSG)
	bool bMsg = msg && msg[0] != '\0';
	char szBuffer[4096];
#if defined(MOD_DEBUG_MINIDUMP)
	const TCHAR* dumpPath = GetLastMiniDumpPath();
	_snprintf_s(szBuffer, _countof(szBuffer), _TRUNCATE,
		"An error in the code has occurred. Please report the issue at https://github.com/LoneGazebo/Community-Patch-DLL/issues so it can be fixed.\n\n"
		"Please provide the VP version number, the list of other mods in use, and a screenshot of this message. If possible, attach a savegame from immediately before the bug occurs.\n\n"
		"To prevent undefined or gamebreaking behavior, the game will now crash.\n\n"
		"==================\n"
		"Detailed information:\n"
		"Expression: %s\n"
		"File: %s\n"
		"Line: %u\n"
		"%s%s%s"
		"%s%s\n",
		expr, ShortenFilePath(szFile), uiLine,
		bMsg ? "Message: " : "", bMsg ? msg : "", bMsg ? "\n" : "",
		dumpPath ? "Minidump: " : "", dumpPath ? dumpPath : ""
	);
#else
	_snprintf_s(szBuffer, _countof(szBuffer), _TRUNCATE,
		"An error in the code has occurred. Please report the issue at https://github.com/LoneGazebo/Community-Patch-DLL/issues so it can be fixed.\n\n"
		"Please provide the VP version number, the list of other mods in use, and a screenshot of this message. If possible, attach a savegame from immediately before the bug occurs.\n\n"
		"To prevent undefined or gamebreaking behavior, the game will now crash.\n\n"
		"==================\n"
		"Detailed information:\n"
		"Expression: %s\n"
		"File: %s\n"
		"Line: %u\n"
		"%s%s%s\n",
		expr, ShortenFilePath(szFile), uiLine,
		bMsg ? "Message: " : "", bMsg ? msg : "", bMsg ? "\n" : ""
	);
#endif

	SetPreconditionFired();

	// Show dialog
	MessageBoxA(NULL, szBuffer, "Error",
		MB_OK | MB_ICONERROR | MB_SYSTEMMODAL | MB_SETFOREGROUND);

#else // VPRELEASE_ERRORMSG
	bool bIgnoreAlways = false;
	CvAssertDlg(expr, ShortenFilePath(szFile), uiLine, bIgnoreAlways, msg);
	// Do NOT reuse CvAssertDlg here: its "No - Continue" / "Cancel - Ignore"
	// options are lies for a precondition - PRECONDITION executes
	// BUILTIN_TRAP() unconditionally after this function returns.
	bool bMsg = msg && msg[0] != '\0';
	char szBuffer[4096];
#if defined(MOD_DEBUG_MINIDUMP)
	const TCHAR* dumpPath = GetLastMiniDumpPath();
#else
	const TCHAR* dumpPath = NULL;
#endif
	_snprintf_s(szBuffer, _countof(szBuffer), _TRUNCATE,
		"Precondition Failed!\n"
		"==================\n"
		"Expression: %s\n"
		"File: %s\n"
		"Line: %u\n"
		"%s%s%s"
		"%s%s\n"
		"\nPreconditions are fatal: the game will crash when you close this dialog.\n"
		"A minidump will be written to the crashlogs folder.\n"
		"\nAttach a debugger now if you want to inspect the live process.",
		expr, ShortenFilePath(szFile), uiLine,
		bMsg ? "Message: " : "", bMsg ? msg : "", bMsg ? "\n" : "",
		dumpPath ? "Minidump: " : "", dumpPath ? dumpPath : ""
	);

	SetPreconditionFired();

	MessageBoxA(NULL, szBuffer, "Precondition Failed",
		MB_OK | MB_ICONERROR | MB_TASKMODAL | MB_SETFOREGROUND);
#endif
```

schnetziomi5: > I already think the crash dialog may too much dev info, we could increase non-devs readability, leave most to the crash log dump?

Hm. I am working on adding stack traces at the moment. The way I see it is that the popup is the first thing we look at when someone posts on a report Github, or asks for help about a crash on Discord or in the forum. Sure we can ask them to provide the logs, and even the minidump, and check those, but the more information the popup provides the less likely we have to. Posting - and checking out - a screenshot takes very little time for all people involved, so we better get the most out of it.

Same for us developers and modmodders - the more information I immediately get, the less time I spend on digging through logs or trial-and-error-testing to find out what the problem is.

image
WIP....

schnetziomi5:
That person sure is brave. Do we know whether they succeeded? If support for CivXP is indeed(?) a long term goal(?) then so be it. Are we using windows API for anything else other than the popups, the unhandled exception filter, and the minidumps?

azum4roll:
The modpack maker?

PR #13271, opened 2026-08-07
Ideology picking AI refactor

Hokath:
Work in progress.

Idea is to unhardcode the ideology -> victory -> grand strategy pipeline in the DoChooseIdeology function so that:

a) an existing ideology could be assigned a different victory pursuit (i.e. remove diplo focus from Autocracy)

b) new, modded ideologies can have policy picking logic applied to them (also without needing a new grand strategy) rather than needing lua hacks


CvGameCoreDLL_Expansion2/CvPolicyAI.cpp
```diff
@@ -413,7 +413,7 @@ void CvPolicyAI::DoChooseIdeology(CvPlayer *pPlayer)
 	int iPriorityToDivide = /*1000*/ GD_INT_GET(IDEOLOGY_SCORE_GRAND_STRATS);
 	int iHappinessModifier = /*10*/ GD_INT_GET(IDEOLOGY_SCORE_HAPPINESS);
 	
-	for (int i = 0; i < vIdeologyPriorities.size(); i++)
+	for (int i = 0; i < (int)vIdeologyPriorities.size(); i++)
```
azum4roll: use `for (size_t i = 0; i < vIdeologyPriorities.size(); i++)` instead

Hokath:
Tested with custom ideologies and, indeed, the AI can pick them!
image
No lua involved

CvGameCoreDLL_Expansion2/CvPolicyAI.cpp
```diff
@@ -384,333 +404,198 @@ void CvPolicyAI::DoChooseIdeology(CvPlayer *pPlayer)
...
+				if (pBranch->IsVictorySupported(eVictory))
 				{
-					iAutocracyPriority += /*500*/ GD_INT_GET(IDEOLOGY_SCORE_HOSTILE);
-					iOrderPriority += /*500*/ GD_INT_GET(IDEOLOGY_SCORE_HOSTILE);
-				}
-				else if (eOtherPlayerIdeology == eAutocracyBranch)
-				{
-					iFreedomPriority += /*500*/ GD_INT_GET(IDEOLOGY_SCORE_HOSTILE);
-					iOrderPriority += /*500*/ GD_INT_GET(IDEOLOGY_SCORE_HOSTILE);
+					vIdeologyPriorities[i] += vGrandStrategyPriorities[iPrefGrandStrategy];
```
axatin: guard against iPrefGrandStrategy being -1 here?

Hokath: good point, I guess that is possible sometimes

PR #13289, opened 2026-08-16
Policy and Tech choice fixes

KungCheops:
Fix policy choice not being triggered during the player's turn.

Fix AI not choosing policies and techs if it becomes available during their turn.


axatin:
nice, thanks!

Issue #13274, opened 2026-08-10
Trade Routes being affected by Puppet Yield Penalty


Hokath:
### 1. Mod version

5.4.2

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

.

### 4. Describe the Issue

International Trade routes sent by Puppets are affected by the multiplicative penalty

<img width="514" height="406" alt="Image" src="https://github.com/user-attachments/assets/39b9f3b1-eda2-4763-a87c-7e67de997dd1" />

This was not true before the change. I don't believe it should apply to them.

axatin:
According to the [proposal](https://forums.civfanatics.com/threads/9-62-puppet-city-rebalance.701681/):
> The puppet yield modifier is now a multiplier, applied at the very end of yield calculation (after post-modifier yields) so it doesn't double penalize "yield conversions".

So this is intentional

Hokath:
I read the proposal but it doesn't specifically discuss TRs it seems.

I know @azum4roll has written messages in discord implying you should use the Venice UA by sending TRs from multiple cities to the same foreign city -- but if Venice other Cities are still on a 50% reduction that wouldn't make sense. Since this was his proposal I assume he would know that. Perhaps he can comment?

Alpakinator:
If it's intentional, then it's a bad intention. Otherwise 'target restrictions removed' can be removed from Venice UA, since its purpose is to allow TR to the same cities from multliple cities, and rn. TR from not Venice are useless.

azum4roll:
The only intention was "not double penalize yield conversions (i.e. process)", as that's based on the amount of yields after modifiers. Any other interaction was not intended.

PR #13291, opened 2026-08-16
Suggested change to the puppet modifier -- trade route interaction

Hokath:
I think this is would exclude trade route yields from the puppet multiplier and not have any other effects?

#13274


azum4roll:
How does this affect the tooltip?

Hokath:
It should put the trade route gold after the +/- x% line if it exists,
so or just at the end of the standard post-additive-modifier list if it doesn't.

I need to test it in game to make sure I didn't miss anything.

Hokath:
image

looks like this.
seems to work correctly

azum4roll:
Do you have a screenshot of a tooltip with a process running?

PR #13318, opened 2026-08-24
Fixes

axatin:


azum4roll:
This still won't be accurate, since producing the unit reduces production which in turn increases local distress and further reduces production. We should at least exclude the penalty from being counted in needs unhappiness.

Vordikk:
> This still won't be accurate, since producing the unit reduces production which in turn increases local distress and further reduces production

If so, i see the solution in separating general modifiers with unhappiness + type-specific modifiers. Account only general modifiers before unhappiness reduction toward distress, then add the rest modifiers.

PR #13110, opened 2026-06-08
10-035b_linear_scaling_increase_per_city

Alpakinator:
As per:
https://forums.civfanatics.com/threads/10-035b-wide-nerf-counterproposal-linear-scaling-increase-per-city.702802/

plus:
- Replaced rounding down of policy costs down to nearest 5, with down to integer. Both in CPP and VP. Tech costs were already like that. More consistent, accurate and fixes cases where settling extra city causes no policy cost increase cause the increase was more than 1 but less than 5 - less confusing.
- 1st city counts to policy cost increase just like it already did for tech, both in CPP and VP. More consistent, less confusing for players as this behaviour was not explained anywhere. It's a balance change, but the exclusion for policies only is so arbitrary and hidden i realized it must be changed.
- Policy and tech penalties can have hundrets of a percent resolution. eg. vanilla and CPP have 7.5% policy penalty and 3.75% tech penalty on Large maps but it gets pruned to 7 and 3% both for actual calc. and UI.

- Improved top panel tech and policy tooltips to give breakdown and prediction of penalties of settling.
   CPP (2 normal cities and 1 puppet):

image
image

  VP-EUI and non-EUI (2 normal cities and 1 puppet):
image
image
(fixed culutrecost typo)
- Cleaned dead code, no legacy methods left afaik.  

Open to feedback and improvements.

PR #13320, opened 2026-08-24
Show game speed and difficulty factors in the policy cost tooltip

Alpakinator:
In response to this bug where epic speed 50% modifier is not applied of base policy cost and instead silently multiplies the standard speed base cost (22938):
image

Now the base cost in the tooltip is 50% higher on epic and by 200% on marathon.

Also made phrasing less confusing and added display of low difficulty factors.

The bug affected only tooltip display.
Also removed the rounding of policy costs to lower 5 in CPP, and made it 1, just like i changed it in VP when implementing these tooltips for the 1st time (https://github.com/LoneGazebo/Community-Patch-DLL/pull/13110), for consistency and removal of confusing mechanics, since tech costs are not rounded to 5 in CPP.


azum4roll:
That's going to heavily affect the balance of Community Patch though. People expect a near-vanilla experience when playing it.

Alpakinator:
> That's going to heavily affect the balance of Community Patch though. People expect a near-vanilla experience when playing it.

I figured since i was allowed to remove the lack of policy penalty for the 1st city in Community Patch, that such tiny change as rounding would be fine. I don't see how it affect balance meaningfully. Also i said in (https://github.com/LoneGazebo/Community-Patch-DLL/pull/13110)  that I did exactly that and it was still accepted.

axatin:
> I figured since i was allowed to remove the lack of policy penalty for the 1st city in Community Patch, that such tiny change as rounding would be fine.

I reconsidered later and reverted that change. Community Patch doesn't make any changes to the policy cost calculations

Alpakinator:
>>
 I reconsidered later and reverted that change. Community Patch doesn't make any changes to the policy cost calculations
> >

I'll revert to 5 rounding then.

Alpakinator:
> > I figured since i was allowed to remove the lack of policy penalty for the 1st city in Community Patch, that such tiny change as rounding would be fine.
> 
> I reconsidered later and reverted that change. Community Patch doesn't make any changes to the policy cost calculations

unfortunately you changed rounding to from 1 to 3, while in vanilla it's 5

PR #13321, opened 2026-08-24
Rework military supply tooltip; use silver fist icon for supply

Alpakinator:
Screenshot_20260825_001041

12 from cities cause they all have Ikandas.  It also lists modifier from war weariness, not seen on the screenshot.

image

THx to @Hokath suggestion War icon is now never used for military supply, it's heavily overused across topics. Different icons for military supply and military supply from population is nonsense.


(2) Vox Populi/Database Changes/Text/en_US/UI/NewUIText.xml
```diff
@@ -718,14 +718,66 @@
...
+		<Row Tag="TXT_KEY_DIFFICULTY_MODIFIER_SUPPLY">
+			<Text>Difficulty Modifier</Text>
+		</Row>
 		<Row Tag="TXT_KEY_CITY_COUNT_SUPPLY_TT">
 			<Text>Reduced from number of Cities</Text>
 		</Row>
 		<Row Tag="TXT_KEY_CITY_COUNT_SUPPLY">
 			<Text>City Count Reduction</Text>
 		</Row>
+		<!-- The old TXT_KEY_UNIT_SUPPLY_REMAINING_TOOLTIP* rows stay: stock EUI calls them -->
```
azum4roll: We don't care about stock EUI. Just remove it.

Alpakinator: True, fixed

Cyclopia31:
I find the ".+10 from 6 cities" a little strange. I think there is too much space between the "6" and "cities" for my taste. 

Or maybe i am just crazy ?! :D

Alpakinator:
> I find the ".+10 from 6 cities" a little strange. I think there is too much space between the "6" and "cities" for my taste.
> 
> Or maybe i am just crazy ?! :D

Maybe i could drop the 6 or the city icon. not sure.

azum4roll:
Drop the number of cities. The player already knows that.

Alpakinator:
I dropped the city number in the 1st line but left in the penalty section so players have numbers right there to understand the penalty

Hokath:
Very nice!

Is the tech progress line at the end necessary?

Alpakinator:
> Very nice!
> 
> Is the tech progress line at the end necessary?

No, but i thought eg.114% tech progress would be confusing so needed an explanation. I bet most people don't know how the whole supply system and this in particular works.

PR #13323, opened 2026-08-26
Fix GetPercentGlobalMonopolies method

Hokath:
I tested this in a stupid way. It was actually giving 100% if you owned any number of monopolies...

If anyone is free, would you double check I didn't do something stupid. Will test later today.

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlayer.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlayer.cpp
@@ -38915,12 +38915,24 @@ void CvPlayer::SetHasGlobalMonopoly(ResourceTypes eResource, bool bNewValue)
 
 int CvPlayer::GetPercentGlobalMonopolies() const
 {
-	int iGlobalsOnMap = GetNumGlobalMonopolies();
+	int iGlobalsOnMap = 0;
+
+	for (int iResourceLoop = 0; iResourceLoop < GC.getNumResourceInfos(); iResourceLoop++)
+	{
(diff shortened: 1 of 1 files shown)


CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -38915,12 +38915,24 @@ void CvPlayer::SetHasGlobalMonopoly(ResourceTypes eResource, bool bNewValue)
...
 int CvPlayer::GetPercentGlobalMonopolies() const
 {
-	int iGlobalsOnMap = GetNumGlobalMonopolies();
+	int iGlobalsOnMap = 0;
+
+	for (int iResourceLoop = 0; iResourceLoop < GC.getNumResourceInfos(); iResourceLoop++)
+	{
+		ResourceTypes eResource = static_cast<ResourceTypes>(iResourceLoop);
+		CvResourceInfo* pkResourceInfo = GC.getResourceInfo(eResource);
+		if (pkResourceInfo && pkResourceInfo->isMonopoly())
```
azum4roll: Redundant null check

Hokath: why's it redundant? I usually see it for resource, or something like

```
const CvResourceInfo* pkResourceInfo = GC.getResourceInfo(eResource);
ASSERT(pkResourceInfo != NULL);
```

It's because of the way the loop is constructed with getNumResourceInfos()?

Hokath:
Tested and code is now working as intended

PR #13328, opened 2026-08-28
Potential stability improvements

rikonen:
I asked Claude Code / Opus 5 to analyze the repo and look for possible bugs that could cause stability issues, especially towards late game. It did report several findings that could have some impact.

I'm not familiar with the codebase myself nor did I have build tools available so validation for these has been minimal. If you think this is AI slop feel free to close the pull request, or if there's some additional steps you wish me to take to make this more useful, please let me know.

It likely doesn't make sense to merge the full branch as-is, but if there are commits that look like they're actually valid they could be cherry-picked to some other branch.


axatin:
Thanks for this! I merged some commits that are clearly correct and don't break save games, I'll include them in the next hotfix. I'll review the other ones later when making the next save-game breaking release.

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -10416,10 +10416,23 @@ void CvPlayer::doTurnPostDiplomacy()
 		// AI spaceship production and building Citizen Earth Protocol is planned on player level, overriding the normal AI city production selection
 		AI_doSpaceshipAndUtopiaProduction();
 
+		// a city that finishes razing disbands itself inside doTurn(), which removes it from the city
+		// list and shifts the cities after it down one index. nextCity() only increments the index, so
+		// iterating live would skip the city that moved into the freed slot and it would lose its turn.
+		vector<int> viCityIDs;
 		int iLoop = 0;
 		for(CvCity* pLoopCity = firstCity(&iLoop); pLoopCity != NULL; pLoopCity = nextCity(&iLoop))
```
azum4roll, line 10424: Just iterate backwards then. Call `firstCity(&iLoop, true)` and `nextCity(&iLoop, true)` instead.

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -10836,9 +10836,18 @@ void CvPlayer::DoUnitAttrition()
 
 void CvPlayer::RepositionInvalidUnits()
 {
+	//two step approach because deleting a unit invalidates the iterator
+	std::vector<int> vUnitIDs;
 	int iLoop = 0;
 	for (CvUnit* pLoopUnit = firstUnit(&iLoop); pLoopUnit != NULL; pLoopUnit = nextUnit(&iLoop))
```
azum4roll, line 10842: Same here, just iterate backwards.

CvGameCoreDLL_Expansion2/Lua/CvLuaPlayer.cpp
```diff
@@ -8328,9 +8328,13 @@ int CvLuaPlayer::lGetNextGreatPersonCost(lua_State* L)
...
 	{
+		// The index comes straight from Lua, and getGreatPersonInfo traps on an invalid one
 		const GreatPersonTypes eGreatPerson = static_cast<GreatPersonTypes>(lua_tointeger(L, 2));
-		const UnitClassTypes eUnitClassType = static_cast<UnitClassTypes>(GC.getGreatPersonInfo(eGreatPerson)->GetUnitClassType());
-		iResult = pCity->GetCityCitizens()->GetSpecialistUpgradeThreshold(eUnitClassType);
+		if (eGreatPerson >= 0 && eGreatPerson < GC.getNumGreatPersonInfos())
+		{
+			const UnitClassTypes eUnitClassType = static_cast<UnitClassTypes>(GC.getGreatPersonInfo(eGreatPerson)->GetUnitClassType());
+			iResult = pCity->GetCityCitizens()->GetSpecialistUpgradeThreshold(eUnitClassType);
+		}
```
azum4roll, line 8337: These ones are expected to crash instead of silently returning -1 (which is a nonsensical number anyway). Instead of waiting for the PRECONDITION to fire crashing the whole game, we could use luaL_error crashing the Lua context, warning the player that something has gone wrong.

CvGameCoreDLL_Expansion2/CvCity.cpp
```diff
@@ -5311,7 +5311,9 @@ CvString CvCity::GetScaledHelpText(CityEventChoiceTypes eEventChoice, bool bYiel
 					localizedSpecialistText = Localization::Lookup("TXT_KEY_EVENT_YIELD_SCALED_SPECIALIST");
 				}
 
-				CvUnitClassInfo* pkUnitClassInfo = GC.getUnitClassInfo((UnitClassTypes)pkSpecialistInfo->getGreatPeopleUnitClass());
+				// A specialist need not have a great person unit class, and the accessor traps on -1
```
azum4roll, line 5314: This one is from modders incorrectly adding an event that modifies the GPP of a specialist, which doesn't make sense if the specialist doesn't have a corresponding great person. I'd leave an ASSERT here (or where the event starts to take effect) to warn the modders.

CvGameCoreDLL_Expansion2/CvMinorCivAI.cpp
```diff
@@ -5178,7 +5178,10 @@ void CvMinorCivAI::DoFirstContactWithMajor(PlayerTypes eMeetingPlayer, bool bSup
 				szTxtKeySuffix = "UNIT";
 				break;
 			default:
-				UNREACHABLE();
+				// A City-State whose MinorCivTrait is unset or unrecognised gets the friendship
+				// boost but no trait-specific gift. This is the same state as bNoGifts above, which
+				// also leaves every gift at zero and the suffix at "UNKNOWN".
+				break;
 			}
```
azum4roll, line 5185: I don't think a CS should be allowed to not have a valid trait?

CvGameCoreDLL_Expansion2/CvMinorCivAI.cpp
```diff
@@ -17335,6 +17340,12 @@ bool CvMinorCivInfo::CacheResults(Database::Results& kResults, CvDatabaseUtility
 	szTextVal = kResults.GetText("MinorCivTrait");
 	CvString strMinorCivTraitType = szTextVal;
 	m_iMinorCivTrait = GC.getInfoTypeForString(szTextVal, true);
```
azum4roll, line 17342: Maybe remove the `true` here to crash at the source.

CvGameCoreDLL_Expansion2/CvMinorCivAI.cpp
```diff
@@ -17335,6 +17340,12 @@ bool CvMinorCivInfo::CacheResults(Database::Results& kResults, CvDatabaseUtility
...
+	// getInfoTypeForString looks the string up in a single map shared by every info table, so an
+	// unset or misspelled MinorCivTrait yields -1 and a string belonging to some other table yields
+	// that table's row id. Normalise both to NO_MINOR_CIV_TRAIT_TYPE so the trait switches see a
+	// value they recognise as "no trait" rather than a foreign id that reads as a valid trait.
+	if (m_iMinorCivTrait < 0 || m_iMinorCivTrait >= NUM_MINOR_CIV_TRAIT_TYPES)
```
azum4roll, line 17347: And then an ASSERT/PRECONDITION here to catch most other table types.

CvGameCoreDLL_Expansion2/CvProjectClasses.cpp
```diff
@@ -152,6 +152,12 @@ bool CvProjectEntry::CacheResults(Database::Results& kResults, CvDatabaseUtility
 				const char* szVictoryType = kDBResults.GetText("VictoryType");
 				const int idx = GC.getInfoTypeForString(szVictoryType, true);
 
+				// getInfoTypeForString returns -1 for a string it does not know, and the row id of
+				// some other info table for a string that is not a victory type at all. Both index
+				// outside these arrays, which are only iNumVictories long.
+				if (idx < 0 || idx >= iNumVictories)
```
azum4roll, line 158: Add an ASSERT/PRECONDITION on idx < iNumVictories.

CvGameCoreDLL_Expansion2/CvTeam.cpp
```diff
@@ -8157,7 +8157,10 @@ void CvTeam::processTech(TechTypes eTech, int iChange, bool bNoBonus)
 				{
 					UnitClassTypes ePikemanClass = (UnitClassTypes)GC.getInfoTypeForString("UNITCLASS_PIKEMAN");
 					UnitTypes eZuluImpi = (UnitTypes)GC.getInfoTypeForString("UNIT_ZULU_IMPI");
-					if(pLoopUnit != NULL && pLoopUnit->getUnitClassType() == ePikemanClass && kPlayer.canTrainUnit(eZuluImpi, false, false, true))
```
azum4roll, line 8160: This one is vanilla-esque jank. I'd leave the crash here instead of silently not doing anything. Modders deserve this if they remove the Pikeman class or the Impi unit without also disabling the trait.

CvGameCoreDLL_Expansion2/CvVotingClasses.cpp
```diff
@@ -4827,7 +4827,8 @@ int CvLeague::GetProjectCostPerPlayer(LeagueProjectTypes eLeagueProject) const
 
 		if (MOD_BALANCE_VP)
 		{
-			const CvProcessInfo* pkProcessInfo = GC.getProcessInfo(pProjectInfo->GetProcess());
+			// The process is optional in the database, and the accessor traps on -1
+			const CvProcessInfo* pkProcessInfo = pProjectInfo->GetProcess() != NO_PROCESS ? GC.getProcessInfo(pProjectInfo->GetProcess()) : NULL;
```
azum4roll, line 4831: It's implied that the project has an associated process if it has a cost, otherwise it makes no sense. We could add an ASSERT to the top of this function.

azum4roll:
In general, all Lua validations should fail through luaL_error instead of returning nonsensical values like -1 (unless it already works like so in the existing implementation AND doesn't trigger an ASSERT/PRECONDITION downstream).

All setters of arrays/vectors should have PRECONDITIONs instead of silent returns, and getters should avoid returning nonsensical values unless intended.

CvGameCoreDLL_Expansion2/Lua/CvLuaPlayer.cpp
```diff
@@ -9376,8 +9464,21 @@ int CvLuaPlayer::lGetQuestSpyMissionString(lua_State* L)
...
+	{
+		lua_pushstring(L, "");
+		return 1;
+	}
 	const MinorCivQuestTypes eType = (MinorCivQuestTypes) lua_tointeger(L, 3);
 	const CityEventChoiceTypes eSpyMission = (CityEventChoiceTypes)pkPlayer->GetMinorCivAI()->GetQuestData1(ePlayer, eType);
+	// GetQuestData1 returns NO_QUEST_DATA when this player has no quest of this type, and
+	// getCityEventChoiceInfo traps on it
+	if (eSpyMission == (CityEventChoiceTypes)CvMinorCivQuest::NO_QUEST_DATA)
+	{
```
azum4roll, line 9478: Similar case here: "" is a nonsensical return value. Better throw an error instead.

CvGameCoreDLL_Expansion2/CvGame.cpp
azum4roll: Definitely don't change this, especially when it leads to network calls.

azum4roll:
Instead of removing PRECONDITIONs in 1e2b047, consider removing the unreachable code after them. Or leave it as is, if we'll ever consider turning off PRECONDITIONs again.

CvGameCoreDLL_Expansion2/CvPlayer.cpp
```diff
@@ -30666,10 +30666,17 @@ void CvPlayer::SetLegislatureName(const char* strKey)
 
 int CvPlayer::GetPoliticPercent(int iID) const
 {
+	// Both accessors are exposed to Lua with no validation on the index
```
azum4roll, line 30669: PRECONDITION here.

CvGameCoreDLL_Expansion2/CvCity.cpp
```diff
@@ -29491,7 +29491,17 @@ void CvCity::popOrder(int iNum, bool bFinish, bool bChoose)
...
 	if (bFinish && pOrderNode->bSave)
 	{
+		int iOrderIndex = iCount;
 		pushOrder(pOrderNode->eOrderType, pOrderNode->iData1, pOrderNode->iData2, true, false, true);
+
+		// m_orderQueue is a vector, not a linked list - the append above can reallocate its
+		// buffer and free the old one, which leaves pOrderNode dangling for the reads and the
+		// deleteNode() below. Re-derive the node from the position we found it at.
+		pOrderNode = getOrderFromQueue(iOrderIndex);
+		if (pOrderNode == NULL)
```
azum4roll, line 29501: Nice catch, but could this really be NULL?

CvGameCoreDLL_Expansion2/CvGame.cpp
```diff
@@ -11801,8 +11810,8 @@ void CvGame::DoMinorPledgeProtection(PlayerTypes eMajor, PlayerTypes eMinor, boo
 /// Amount of Gold being gifted to the Minor by the active player
 void CvGame::DoMinorGiftGold(PlayerTypes eMinor, int iNumGold)
 {
```
azum4roll, line 11812: Definitely don't change this (and the ones below), especially when it leads to network calls.

CvGameCoreDLL_Expansion2/CvPlayer.cpp
axatin: kill() can call kill() for other units (cargo) so that wouldn't be safe

CvGameCoreDLL_Expansion2/CvPlayer.cpp
axatin: can we be sure that reversing the order wouldn't have any gameplay effects?

azum4roll: It may matter slightly with production cost scaling with number of cities or number of world wonders. But neither order is ideal in this case.

CvGameCoreDLL_Expansion2/CvPlayer.cpp
azum4roll: True

JohnsterID:
LLM review: https://gist.github.com/JohnsterID/685c20ee4b66e8f84c628f30d0dcea5a

rikonen:
Thanks for the comments. I've incorporated them to LLM instructions so that it won't make the same mistakes again if I'll later ask it to try to find other problems. I'm closing this PR now, I presume you've taken the useful things from here or can find them from the closed PR if not.

PR #13343, opened 2026-09-02
Fix setPlotType crash on some mapscripts

TheKineticz:
A change that went live in 5.3.3 ([35ca55c89](https://github.com/LoneGazebo/Community-Patch-DLL/commit/35ca55c89116cc12297cc43d77e5d02c6c3e9f70)) causes `CvPlot::setPlotType` to crash when a plot's area or landmass ID doesn't map to a valid object.

image

Lua exposes `SetArea`, and calling `SetArea(-1)` before `SetPlotType` with `bRecalculate = true` hits this path.

**Repro**
Any map script calling `plot:SetArea(-1)` then `plot:SetPlotType(PlotTypes.PLOT_OCEAN, true, true)`. Planet Simulator is the one script I have that tripped it in its `AddLakes` function.

[TL;DR](https://www.youtube.com/watch?v=bLHL75H_VEM)

### DIFF
--- a/CvGameCoreDLL_Expansion2/CvPlot.cpp
+++ b/CvGameCoreDLL_Expansion2/CvPlot.cpp
@@ -7417,12 +7417,12 @@ void CvPlot::setPlotType(PlotTypes eNewValue, bool bRecalculate, bool bRebuildGr
 				setArea(-1);
 				setLandmass(-1);
 
-				if (pCurrArea->getNumTiles() == 0)
+				if (pCurrArea && pCurrArea->getNumTiles() == 0)
 				{
 					GC.getMap().deleteArea(pCurrArea->GetID());
 				}
 
-				if (pCurrLandmass->getNumTiles() == 0)
+				if (pCurrLandmass && pCurrLandmass->getNumTiles() == 0)
 				{
 					GC.getMap().deleteLandmass(pCurrLandmass->GetID());
 				}


azum4roll:
Harmless enough fix. But why does a mapscript call SetArea(-1) before recalculating areas? That's redundant.

Recalculation is very slow and should only be called once at the end of map generation, unless you need the areas midway through (Continent Drift does but somehow I don't find any recalc there. Could be bugged?)

TheKineticz:
Why I have no idea. It's a similarly simple fix in the mapscript itself, but if the script works in vanilla it should nominally work in VP as well

jayeye2011:
> Why I have no idea. It's a similarly simple fix in the mapscript itself, but if the script works in vanilla it should nominally work in VP as well

This is invalid logic, a modmod mapscript should not force changes to Vox Populi code to allow it to work, it is the other way around. In the case of Planet Simulator, it's never going to work anymore given the massive changes to code and data tables recently made related to AssignStartingPlots.lua.  Even if you allow SetArea(-1) or SetLandMass(-1) to be valid calls, you will still need to either rewrite all the Planet Simulator AssignStartingPlots override functions in the map script to conform to the Vox Populi changes or more simply delete the AssignStartingPlots overrride functions from the Planet Simulator mapscript code otherwise you get no natural wonders spawned.

jayeye2011:
@azum4roll It seems to me that Planet Simulator is used enough by people who use the Vox  Populi mod that it would best to add a Vox Populi compatible version to the (2) Vox Populi maps folder given how simple the changes to the lua file are:
1) remove the SetArea(-1) call and 
2) remove the three natural wonder related AssignStartingPlots override functions:
function AssignStartingPlots:CanBeThisNaturalWonderType(x, y, wn, rn)
function AssignStartingPlots:AttemptToPlaceNaturalWonderMOD(wonder_number, row_number)
function AssignStartingPlots:PlaceNaturalWondersMOD()
And this would reduce the reporting of issues relating to the mapscript not working in the future. Also, even when it used to work, the natural wonders could spawn next to a city state whereas the Vox Populi natural wonder assignment is superior in that it  prevents this from happening.

TheKineticz:
> > Why I have no idea. It's a similarly simple fix in the mapscript itself, but if the script works in vanilla it should nominally work in VP as well
> 
> This is invalid logic, a modmod mapscript should not force changes to Vox Populi code to allow it to work, it is the other way around. In the case of Planet Simulator, it's never going to work anymore given the massive changes to code and data tables recently made related to AssignStartingPlots.lua. Even if you allow SetArea(-1) or SetLandMass(-1) to be valid calls, you will still need to either rewrite all the Planet Simulator AssignStartingPlots override functions in the map script to conform to the Vox Populi changes or more simply delete the AssignStartingPlots overrride functions from the Planet Simulator mapscript code otherwise you get no natural wonders spawned.

It's perfectly valid logic when the CTD is a regression introduced in a recent update, and was working fine.

axatin:
>  (Continent Drift does but somehow I don't find any recalc there. Could be bugged?)

`Map.RecalculateAreas()` called in line 1532 and 1949


If the error in the map script is just a redundant function call and not something that actually broke things in vanilla, then VP shouldn't crash with it.

PR #13348, opened 2026-09-05
Crash reporting improvement

JohnsterID:
@schnetziomi5 https://github.com/LoneGazebo/Community-Patch-DLL/commit/6e0d3680d4ec95a63fde61e624f807fae33578d0 based on your comments

### DIFF
--- a/CvGameCoreDLL_Expansion2/CustomMods.h
+++ b/CvGameCoreDLL_Expansion2/CustomMods.h
@@ -60,9 +60,15 @@
 // Enables writing gameplay statistics to a local SQLite database (stats.db) via SqliteLogger
 #define MOD_SQLITE_LOGGING											gCustomMods.isSQLITE_LOGGING()
 
-// Comment out this line to remove minidumps - see http://forums.civfanatics.com/showthread.php?t=498919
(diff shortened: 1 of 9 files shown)


schnetziomi5:
Nice, thank you!

Arguably we could remove the memory usage reduction tips with detection now in place? Even with the softened wording people still like to report any crash they see them as an out-of-memory one even when the actual information in the popup doesn't support it at all - and who can blame then when one third of the space is always filled with those tips...
@axatin 


(Also the stacktraces I would like to add need some lines too, and I haven't found a reasonable way to control the size of the popup or its text)

JohnsterID:
I can look at a custom DLGTEMPLATE dialog later as a possibility.

vaderkos:
I'm wondering if we can have a dump what consumes the memory? It doesn't have to be detailed, but simple counts...
Or it's available and I don't know about it :D, CC @axatin, @schnetziomi5, @JohnsterID.

JohnsterID:
We have reserve space now and also my last commit will have hopefully a useful stack even without a full minidump. A out of process dumper by having procdump placed next to the binary is something I will get back to testing as well.

PR #13353, opened 2026-09-07
Require an enemy to cut a city's usable exits to blockade it

TheKineticz:
As it's currently written, `IsBlockaded` will consider a city that's surrounded on all sides by mountains to be blockaded, even with no hostile presence. This is an edge case, but not necessarily rare - any civ can do it after dynamite, and the Inca can do it all game, and all they need to do is settle in the middle of a thick mountain range.
Screenshot From 2026-09-07 18-34-57

This change makes sure that `IsBlockaded` only returns `true` when there is an actual enemy threatening the city, not just when mother nature is embracing it; and it requires that enemy to be blocking a valid exit, if there is one.


azum4roll:
Is this a human-placed city or AI?

TheKineticz:
> Is this a human-placed city or AI?

Human-placed

RecursiveVision:
This is intentional. Historically it was done to nerf Incan cities surrounded entirely by mountains specifically because they can be made unassailable.

Requiring an actual enemy is probably reasonable but it's still substantial since it allows the city to heal.

Issue #13349, opened 2026-09-05
VP-5.4.4_Multiplayer Desync, might be related to the pathfinding


Cyclopia31:
### 1. Mod version

5.4.4

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

Modpack: https://drive.google.com/file/d/1BX7sE_Ya5SfAut30g3ihxtClzdUjxBvs/view?usp=drive_link

https://forums.civfanatics.com/attachments/1781803634709-png.772166/ (Mods inside the modpack)

### 4. Describe the Issue

Hello everyone,

I am testing a little the multiplayer from the patch 5.4.4 and I am facing three desyncs:

- **The first one** 
This is the classical m_aiWarDamageValue desync. It was on turn 109. I couldn't repeat it but I know there are news logs to try to understand this issue so I am still reporting it.

- **The second one** 
It happened during turn 111. I have a unit desync between host and non-host. 

So I took a screenshot of the Host situation (INCA) and then I have reload the same turn using the Non-Host save (with INCA again) and you can see that some units are not at the same position:

Host position: 

<img width="2396" height="967" alt="Image" src="https://github.com/user-attachments/assets/c98e747b-f4ca-4c33-9c2b-bfe31b585530" />

Non_Host position:

<img width="2477" height="1007" alt="Image" src="https://github.com/user-attachments/assets/578e8c5e-43fa-4317-b4cb-2c39d4a7922e" />

You can find the logs/saves in the "Turn 111_1" folder

- **The third one** 
I tried to reproduce the second issue, but I couldn't remember how I had played the first time and failed to reproduce it... However, a new desynchronization issue appeared! See "Turn 111-2" and "Turn 111_3" folders. "Turn 111_3" is the "exact" repeat of "Turn 111-2".

So the issue is repeatable ! 

I made a video of what I did to trigger the desync : "Civilization V_2026.09.05-18.54.mp4"

Furthermore, while attempting to reproduce the synchronization issue that occurred on turn 111, I encountered 2 crashes (CTD) on the host side immediately after clicking "End Turn."--> See "Turn111-112 Crash" Folder, there are the logs/saves/cache and crashlogs.

<img width="542" height="1072" alt="Image" src="https://github.com/user-attachments/assets/613116b2-e9e5-4352-90b1-a9790b5e4139" />

I hope it helps!

### 5. Save Game From 1 Turn Before (ALWAYS ATTACH THIS IF POSSIBLE)

[...]

PR #13361, opened 2026-09-13
 Make equal-cost pathfinding deterministic

CIVITAS-John:
Units can choose different routes when equally good paths are explored in a different heap order. Give tied nodes a stable order so the same pathfinding inputs produce reproducible choices across heap implementations.

This is mostly needed for the offline tactical simulation project. It may make seeded games more reproducible as a result too. 

`PrNodeIsBetter` in `CvAStarNode.h` preserves total-cost and heuristic priorities, then orders ties by Y, X, and normal nodes before stop nodes.


schnetziomi5:
Nice one!
#13349

azum4roll:
Is this a common occurrence? If so, will AI behaviour be too predictable (they prefer using the northeastern path)?

CIVITAS-John:
@azum4roll It is rare, but could contribute to desync. If we want to make AI pathfinding less deterministic, we should do that directly rather than getting the heap instability doing the job :)

azum4roll:
Any identifier you can use that aren't coordinates?

vaderkos:
I think we had a desync each time non-host player spawned a unit near the city (religion/gold buy)

CIVITAS-John:
@azum4roll I don't think so, but maybe we can make the tie rule ties to, eg, the oddity of X/Y? so it is not always resolved to one direction (but still with patterns)?

vaderkos:
@CIVITAS-John what about using the map seed or time to keep it both random and deterministic?

azum4roll:
You could pass it into the deterministic RNG. Not sure about performance though.

CIVITAS-John:
I guess we can use some cheap operators like xor?

PR #13370, opened 2026-09-19
Fix notification text of additional delegates gained. Issue #13368

### LINKED ISSUE #13368: 5.4.6 Wrong text in the notification log on additional delegates after Global Hegemony vote
### 1. Mod version

5.4.6

### 2. Installed Components

Vox Populi with EUI

### 3. Additional mods used

_No response_

### 4. Describe the Issue

In the notification log it says, that after Global Hegemony the leader gains "an additional Delegate", while the leader actually gets 5 delegates.

Image

_No response_

_No response_

### 7. Crash dump and crashes.log (ALWAYS ATTACH IF REPORTING A CRASH)

_No response_

OlegPyatakov:
Fixes #13368
Changed the text to explicitly say 5 delegates.


Hokath:
For my curiosity I checked what determines this. It is set in the Resolution as LeadersVoteBonusOnFail
Indeed, it is set to 5. There doesn't appear to be any kind of possible scaling.

